PR #3, opened 2025-03-28
OpenAPI spec for official registry API

tadasant:
This is a rough first pass at what the API shape should look like. Not meant to be final, we may not even merge this particular PR; just meant to give readers a sense of what we're going for to spark feedback conversations.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -19,9 +19,8 @@ GET /v0/servers?limit=5000&offset=0
       "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
       "repository": {
         "url": "https://github.com/modelcontextprotocol/servers",
-        "subfolder": "src/filesystem",
-        "branch": "main",
-        "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+        "source": "github",
+        "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
       },
       "version": "0.0.3"
     }
@@ -49,9 +48,8 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
   "version": "0.0.3",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
-    "subfolder": "src/filesystem",
-    "branch": "main",
-    "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+    "source": "github",
+    "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
   },
   "registry_canonical": "npm",
   "registries": [
@@ -201,9 +199,8 @@ API Response:
   "version": "1.0.0",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
-    "subfolder": "src/brave-search",
-    "branch": "main",
-    "commit": "abc123"
+    "source": "github",
+    "id": "abc123de-f456-7890-ghij-klmnopqrstuv"
   },
   "registry_canonical": "npm",
   "registries": [
@@ -264,9 +261,8 @@ API Response:
   "version": "0.0.3",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
-    "subfolder": "src/filesystem",
-    "branch": "main",
-    "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+    "source": "github",
+    "id": "d94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
   },
   "registry_canonical": "docker",
   "registries": [
@@ -354,8 +350,8 @@ API Response:
   "version": "1.0.0",
   "repository": {
     "url": "https://github.com/example/remote-fs",
-    "branch": "main",
-    "commit": "xyz789"
+    "source": "github",
+    "id": "xyz789ab-cdef-0123-4567-890ghijklmno"
   },
   "remotes": [
     {
--- a/api/README.md
+++ b/api/README.md
@@ -22,7 +22,11 @@ GET /v0/servers?limit=5000&offset=0
         "source": "github",
         "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
       },
-      "version": "0.0.3"
+      "version_detail": {
+        "version": "1.0.2",
+        "release_date": "2023-06-15T10:30:00Z",
+        "is_latest": true
+      }
     }
   ],
   "next": "https://registry.modelcontextprotocol.io/servers?offset=50",
@@ -45,11 +49,15 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
   "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
   "name": "@modelcontextprotocol/servers/src/filesystem",
   "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
-  "version": "0.0.3",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
     "source": "github",
     "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
+  },
+  "version_detail": {
+    "version": "1.0.2",
+    "release_date": "2023-06-15T10:30:00Z",
+    "is_latest": true
   },
   "package_canonical": "npm",
   "packages": [
@@ -194,11 +202,15 @@ API Response:
   "id": "brave-search-12345",
   "name": "@modelcontextprotocol/server-brave-search",
   "description": "MCP server for Brave Search API integration",
-  "version": "1.0.0",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
     "source": "github",
     "id": "abc123de-f456-7890-ghij-klmnopqrstuv"
+  },
+  "version_detail": {
+    "version": "1.0.2",
+    "release_date": "2023-06-15T10:30:00Z",
+    "is_latest": true
   },
   "package_canonical": "npm",
   "packages": [
@@ -255,11 +267,15 @@ API Response:
   "id": "filesystem-67890",
   "name": "@modelcontextprotocol/servers/src/filesystem",
   "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations",
-  "version": "0.0.3",
   "repository": {
     "url": "https://github.com/modelcontextprotocol/servers",
     "source": "github",
     "id": "d94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
+  },
+  "version_detail": {
+    "version": "1.0.2",
+    "release_date": "2023-06-15T10:30:00Z",
+    "is_latest": true
   },
   "package_canonical": "docker",
   "packages": [
@@ -343,11 +359,15 @@ API Response:
   "id": "remote-fs-54321",
   "name": "Remote Brave Search Server",
   "description": "Cloud-hosted MCP Brave Search server",
-  "version": "1.0.0",
   "repository": {
     "url": "https://github.com/example/remote-fs",
     "source": "github",
     "id": "xyz789ab-cdef-0123-4567-890ghijklmno"
+  },
+  "version_detail": {
+    "version": "1.0.2",
+    "release_date": "2023-06-15T10:30:00Z",
+    "is_latest": true
   },
   "remotes": [
     {
--- a/api/README.md
+++ b/api/README.md
@@ -98,7 +98,8 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
         {
           "name": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
-          "required": false
+          "required": false,
+          "default_value": "info"
         }
       ]
     },
@@ -170,7 +171,8 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
         {
           "name": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
-          "required": false
+          "required": false,
+          "default_value": "info"
         }
       ]
     }
@@ -223,7 +225,8 @@ API Response:
         {
           "name": "BRAVE_API_KEY",
           "description": "Brave Search API Key",
-          "required": true
+          "required": true,
+          "default_value": ""
         }
       ]
     }
--- a/api/README.md
+++ b/api/README.md
@@ -64,6 +64,7 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
     {
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-filesystem",
+      "version": "1.0.2",
       "command": {
         "name": "npx",
         "subcommands": [],
@@ -112,6 +113,7 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
     {
       "name": "docker",
       "package_name": "mcp/filesystem",
+      "version": "1.0.2",
       "command": {
         "name": "docker",
         "subcommands": [
@@ -217,6 +219,7 @@ API Response:
     {
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-brave-search",
+      "version": "1.0.2",
       "command": {
         "name": "npx",
         "subcommands": [],
@@ -282,6 +285,7 @@ API Response:
     {
       "registry_name": "docker",
       "name": "mcp/filesystem",
+      "version": "1.0.2",
       "command": {
         "name": "docker",
         "subcommands": [
--- a/api/README.md
+++ b/api/README.md
@@ -59,7 +59,8 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "name": "npm",
       "package_name": "@modelcontextprotocol/server-filesystem",
       "license": "MIT",
-      "command_arguments": {
+      "command": {
+        "name": "npx",
         "subcommands": [],
         "positional_arguments": [
           {
@@ -107,7 +108,8 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "name": "docker",
       "package_name": "mcp/filesystem",
       "license": "MIT",
-      "command_arguments": {
+      "command": {
+        "name": "docker",
         "subcommands": [
           {
             "name": "run",
@@ -209,7 +211,8 @@ API Response:
       "name": "npm",
       "package_name": "@modelcontextprotocol/server-brave-search",
       "license": "MIT",
-      "command_arguments": {
+      "command": {
+        "name": "npx",
         "subcommands": [],
         "positional_arguments": [],
         "named_arguments": [
@@ -271,7 +274,8 @@ API Response:
       "name": "docker",
       "package_name": "mcp/filesystem",
       "license": "MIT",
-      "command_arguments": {
+      "command": {
+        "name": "docker",
         "subcommands": [
           {
             "name": "run",
--- a/api/README.md
+++ b/api/README.md
@@ -56,7 +56,6 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
     {
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-filesystem",
-      "license": "MIT",
       "command": {
         "name": "npx",
         "subcommands": [],
@@ -105,7 +104,6 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
     {
       "name": "docker",
       "package_name": "mcp/filesystem",
-      "license": "MIT",
       "command": {
         "name": "docker",
         "subcommands": [
@@ -207,7 +205,6 @@ API Response:
     {
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-brave-search",
-      "license": "MIT",
       "command": {
         "name": "npx",
         "subcommands": [],
@@ -269,7 +266,6 @@ API Response:
     {
       "registry_name": "docker",
       "name": "mcp/filesystem",
-      "license": "MIT",
       "command": {
         "name": "docker",
         "subcommands": [
--- a/api/README.md
+++ b/api/README.md
@@ -1,0 +1,360 @@
+# Examples
+
+## /v0/servers
+
+### Request
+
+```http
+GET /v0/servers?limit=5000&offset=0
+```
+
+### Response
+
+```json
+{
+  "servers": [
+    {
+      "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
+      "name": "@modelcontextprotocol/servers/src/filesystem",
+      "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
+      "repository": {
+        "url": "https://github.com/modelcontextprotocol/servers",
+        "subfolder": "src/filesystem",
+        "branch": "main",
+        "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+      },
+      "version": "0.0.3"
+    }
+  ],
+  "next": "https://registry.modelcontextprotocol.io/servers?offset=50",
+  "total_count": 1
+}
+```
+
+## /v0/servers/:id
+
+### Request
+
+```http
+GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
+```
+
+### Response
+
+```json
+{
+  "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
+  "name": "@modelcontextprotocol/servers/src/filesystem",
+  "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
+  "version": "0.0.3",
+  "repository": {
+    "url": "https://github.com/modelcontextprotocol/servers",
+    "subfolder": "src/filesystem",
+    "branch": "main",
+    "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+  },
+  "registry_canonical": "npm",
+  "registries": [
+    {
+      "name": "npm",
+      "package_name": "@modelcontextprotocol/server-filesystem",
+      "license": "MIT",
+      "command_arguments": {
+        "subcommands": [],
+        "positional_arguments": [
+          {
+            "position": 0,
+            "name": "package",
+            "description": "NPM package name",
+            "default_value": "@modelcontextprotocol/server-filesystem",
+            "is_required": true,
+            "is_editable": false,
+            "is_repeatable": false,
+            "choices": []
+          },
+          {
+            "position": 1,
+            "name": "path",
+            "description": "Path to access",
+            "default_value": "/Users/username/Desktop",
+            "is_required": true,
+            "is_editable": true,
+            "is_repeatable": true,
+            "choices": []
+          }
+        ],
+        "named_arguments": [
+          {
+            "short_flag": "-y",
+            "requires_value": false,
+            "is_required": false,
+            "is_editable": false,
+            "description": "Skip prompts and automatically answer yes",
+            "choices": []
+          }
+        ]
+      },
+      "environment_variables": [
+        {
+          "name": "LOG_LEVEL",
+          "description": "Logging level (debug, info, warn, error)",
+          "required": false
+        }
+      ]
+    },
+    {
+      "name": "docker",
+      "package_name": "mcp/filesystem",
+      "license": "MIT",
+      "command_arguments": {
+        "subcommands": [
+          {
+            "name": "run",
+            "description": "Run the Docker container",
+            "is_required": true,
+            "subcommands": [],
+            "positional_arguments": [],
+            "named_arguments": [
+              {
+                "short_flag": "-i",
+                "requires_value": false,
+                "is_required": true,
+                "is_editable": false,
+                "description": "Run in interactive mode"
+              },
+              {
+                "long_flag": "--rm",
+                "requires_value": false,
+                "is_required": true,
+                "is_editable": false,
+                "description": "Remove container when it exits"
+              },
+              {
+                "long_flag": "--mount",
+                "requires_value": true,
+                "is_required": true,
+                "is_repeatable": true,
+                "is_editable": true,
+                "description": "Mount a volume into the container",
+                "default_value": "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
+                "choices": []
+              }
+            ]
+          }
+        ],
+        "positional_arguments": [
+          {
+            "position": 0,
+            "name": "image",
+            "description": "Docker image name",
+            "default_value": "mcp/filesystem",
+            "is_required": true,
+            "is_editable": false,
+            "is_repeatable": false,
+            "choices": []
+          },
+          {
+            "position": 1,
+            "name": "root_path",
+            "description": "Root path for filesystem access",
+            "default_value": "/projects",
+            "is_required": true,
+            "is_editable": false,
+            "is_repeatable": false,
+            "choices": []
+          }
+        ],
+        "named_arguments": []
+      },
+      "environment_variables": [
+        {
+          "name": "LOG_LEVEL",
+          "description": "Logging level (debug, info, warn, error)",
+          "required": false
+        }
+      ]
+    }
+  ],
+  "remotes": [
+    {
+      "transport_type": "sse",
+      "url": "https://mcp-fs.example.com/sse"
+    }
+  ]
+}
+```
+
+### Server Configuration Examples
+
+#### Local Server with npx
+
+API Response:
+```json
+{
+  "id": "brave-search-12345",
+  "name": "@modelcontextprotocol/server-brave-search",
+  "description": "MCP server for Brave Search API integration",
+  "version": "1.0.0",
+  "repository": {
+    "url": "https://github.com/modelcontextprotocol/servers",
+    "subfolder": "src/brave-search",
+    "branch": "main",
+    "commit": "abc123"
+  },
+  "registry_canonical": "npm",
+  "registries": [
+    {
+      "name": "npm",
+      "package_name": "@modelcontextprotocol/server-brave-search",
+      "license": "MIT",
+      "command_arguments": {
+        "subcommands": [],
+        "positional_arguments": [],
+        "named_arguments": [
+          {
+            "short_flag": "-y",
+            "requires_value": false,
+            "is_required": false,
+            "description": "Skip prompts"
+          }
+        ]
+      },
+      "environment_variables": [
+        {
+          "name": "BRAVE_API_KEY",
+          "description": "Brave Search API Key",
+          "required": true
+        }
+      ]
+    }
+  ]
+}
+```
+
+claude_desktop_config.json:
+```json
+{
+  "brave-search": {
+    "command": "npx",
+    "args": [
+      "-y",
+      "@modelcontextprotocol/server-brave-search"
+    ],
+    "env": {
+      "BRAVE_API_KEY": "YOUR_API_KEY_HERE"
+    }
+  }
+}
+```
+
+#### Local Server with Docker
+
+API Response:
+```json
+{
+  "id": "filesystem-67890",
+  "name": "@modelcontextprotocol/servers/src/filesystem",
+  "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations",
+  "version": "0.0.3",
+  "repository": {
+    "url": "https://github.com/modelcontextprotocol/servers",
+    "subfolder": "src/filesystem",
+    "branch": "main",
+    "commit": "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
+  },
+  "registry_canonical": "docker",
+  "registries": [
+    {
+      "name": "docker",
+      "package_name": "mcp/filesystem",
+      "license": "MIT",
+      "command_arguments": {
+        "subcommands": [
+          {
+            "name": "run",
+            "description": "Run the Docker container",
+            "is_required": true,
+            "named_arguments": [
+              {
+                "short_flag": "-i",
+                "requires_value": false,
+                "is_required": true,
+                "description": "Run in interactive mode"
+              },
+              {
+                "long_flag": "--rm",
+                "requires_value": false,
+                "is_required": true,
+                "description": "Remove container when it exits"
+              },
+              {
+                "long_flag": "--mount",
+                "requires_value": true,
+                "is_required": true,
+                "is_repeatable": true,
+                "description": "Mount a volume into the container"
+              }
+            ]
+          }
+        ],
+        "positional_arguments": [
+          {
+            "position": 0,
+            "name": "image",
+            "description": "Docker image name",
+            "default_value": "mcp/filesystem"
+          },
+          {
+            "position": 1,
+            "name": "root_path",
+            "description": "Root path for filesystem access",
+            "default_value": "/projects"
+          }
+        ]
+      }
+    }
+  ]
+}
+```
+
+claude_desktop_config.json:
+```json
+{
+  "filesystem": {
+    "command": "docker",
+    "args": [
+      "run",
+      "-i",
+      "--rm",
+      "--mount", "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
+      "--mount", "type=bind,src=/path/to/other/allowed/dir,dst=/projects/other/allowed/dir,ro",
+      "--mount", "type=bind,src=/path/to/file.txt,dst=/projects/path/to/file.txt",
+      "mcp/filesystem",
+      "/projects"
+    ]
+  }
+}
+```
+
+#### Remote Server
+
+API Response:
+```json
+{
+  "id": "remote-fs-54321",
+  "name": "Remote Brave Search Server",
+  "description": "Cloud-hosted MCP Brave Search server",
+  "version": "1.0.0",
+  "repository": {
+    "url": "https://github.com/example/remote-fs",
+    "branch": "main",
+    "commit": "xyz789"
+  },
+  "remotes": [
+    {
+      "transport_type": "sse",
+      "url": "https://mcp-fs.example.com/sse"
+    }
+  ]
+}
+```
--- a/api/README.md
+++ b/api/README.md
@@ -51,11 +51,11 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
     "source": "github",
     "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
   },
-  "registry_canonical": "npm",
-  "registries": [
+  "package_canonical": "npm",
+  "packages": [
     {
-      "name": "npm",
-      "package_name": "@modelcontextprotocol/server-filesystem",
+      "registry_name": "npm",
+      "name": "@modelcontextprotocol/server-filesystem",
       "license": "MIT",
       "command": {
         "name": "npx",
@@ -202,11 +202,11 @@ API Response:
     "source": "github",
     "id": "abc123de-f456-7890-ghij-klmnopqrstuv"
   },
-  "registry_canonical": "npm",
-  "registries": [
+  "package_canonical": "npm",
+  "packages": [
     {
-      "name": "npm",
-      "package_name": "@modelcontextprotocol/server-brave-search",
+      "registry_name": "npm",
+      "name": "@modelcontextprotocol/server-brave-search",
       "license": "MIT",
       "command": {
         "name": "npx",
@@ -264,11 +264,11 @@ API Response:
     "source": "github",
     "id": "d94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
   },
-  "registry_canonical": "docker",
-  "registries": [
+  "package_canonical": "docker",
+  "packages": [
     {
-      "name": "docker",
-      "package_name": "mcp/filesystem",
+      "registry_name": "docker",
+      "name": "mcp/filesystem",
       "license": "MIT",
       "command": {
         "name": "docker",
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -87,20 +87,20 @@ components:
       type: object
       required:
         - url
+        - source
+        - id
       properties:
         url:
           type: string
           format: uri
           example: "https://github.com/modelcontextprotocol/servers"
-        subfolder:
+        source:
           type: string
-          example: "src/filesystem"
-        branch:
+          enum: [github, gitlab] # TODO: Add all supported sources as a whitelist
+          example: "github"
+        id:
           type: string
-          example: "main"
-        commit:
-          type: string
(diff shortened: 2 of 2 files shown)


api/README.md
```diff
@@ -0,0 +1,360 @@
...
+            "description": "Skip prompts and automatically answer yes",
+            "choices": []
+          }
+        ]
+      },
+      "environment_variables": [
+        {
+          "name": "LOG_LEVEL",
+          "description": "Logging level (debug, info, warn, error)",
+          "required": false
```
alexhancock: Thoughts on including a `default_value` field for a server author that would want to indicate one? We've run across this as a request from MCP server authors.

api/README.md
```diff
@@ -0,0 +1,360 @@
...
+    {
+      "name": "npm",
+      "package_name": "@modelcontextprotocol/server-brave-search",
+      "license": "MIT",
+      "command_arguments": {
+        "subcommands": [],
+        "positional_arguments": [],
+        "named_arguments": [
+          {
+            "short_flag": "-y",
```
alexhancock, line 229: Is this supposed to be indicating `-y` as an argument if running it via `npx`? Just noticing `npx` doesn't actually show up in the rest of the response here.

I could see clients assuming it for npm packages, but it might be more scalable to other registries to fully specify everything needed for a client to run a command that would run the server.

tadasant: Good callout - one of the concerns we discussed early was the potential rabbit hole of being prescriptive about the command. e.g. some people might prefer `bun`, some might prefer `npm`, some might prefer `npx`. So the hypothesis was that we could be agnostic to that and just tie it to the registry and let the client be opinionated from there.

But you bring up a good point that something like `-y` might be specific to `npx`, though... So maybe this is too brittle and we should be more explicit about the command. Maybe we make it an enum just to avoid free form input security risks.

api/README.md
tadasant: Good point yes, will add

api/openapi.yaml
```diff
@@ -0,0 +1,320 @@
...
+          example: "https://github.com/modelcontextprotocol/servers"
+        subfolder:
+          type: string
+          example: "src/filesystem"
+        branch:
+          type: string
+          example: "main"
+        commit:
+          type: string
+          example: "d94b5f7ec7c6d7602c78a5e9b8a5b8c94d093eda"
```
tadasant: Tag or release would be better

api/openapi.yaml
```diff
@@ -0,0 +1,320 @@
...
+  jsonSchemaDialect: "https://json-schema.org/draft/2020-12/schema"
+    Repository:
+      type: object
+      required:
+        - url
+      properties:
+        url:
+          type: string
+          format: uri
+          example: "https://github.com/modelcontextprotocol/servers"
```
tadasant, line 96: GitHub repo ID would be better

api/README.md
alexhancock: I think we either go all the way to fully specifying a command per kind of registry the package is coming from, as well as required args (like `-y` for `npx`) or leave it all out and let the client form commands per registry.

We can always add more attributes into the server metadata, but hard to take things away once we expose them in responses.... perhaps this encourages us towards not including commands to begin and only enough metadata in the response to allow a client to form a command?

tadasant: >  fully specifying a command per kind of registry the package is coming from, as well as required args (like -y for npx)

I'm leaning in this direction I think. I'm realizing that the lack of context of a top level command with these arguments is not great -- `-y` doesn't make sense for anything but `npx`, but even the positional `package` in this example would potentially vary. The subcommands/args only really make sense for a specific command, and it probably makes sense to just be explicit about the command instead of implicit.

> leave it all out and let the client form commands per registry

Not sure I follow, how would this be different from the schema in the current PR? We still somehow need to communicate all the CLI args needed to get a given MCP server to run.

alexhancock: - Not sure I follow, how would this be different from the schema in the current PR?

If we didn't include anything about commands we wouldn't surface `-y` as a named arg for example. As you note, that's an argument usually used for `npx`. The path I am outlining here would be mean not putting anything about commands in the responses. We'd just put metadata about the npm package for a client to use to construct their own command.

There are pros and cons to either the MCP registry choosing to specify a command in responses that can be run to run the package, or leaving it all out.

If we go the route of specifying anything about the command (args, etc) we should fully specify it.

tadasant: > We'd just put metadata about the npm package for a client to use to construct their own command.

Got it - your implication here is that clients would have to construct some multi-step (likely AI-powered, agentic flow) that involves parsing a README, maybe analyzing some code, to infer how to run the CLI command, based on the source code they can scrape, right? I believe Cline does something like this.

IMO that flow is too clunky to be practical for most clients and a good end-user UX, so if I'm understanding your proposal correctly I'd suggest we lean into defining the whole command.

alexhancock: Good point. I hadn't thought about the fact without positional and named args (if required) we wouldn't have enough info on the client to construct a command, even assuming a base (`npx` `uvx`). Perhaps fully specifying required args seems best?

tadasant: > Perhaps fully specifying required args seems best?

Are you saying something above and beyond how we have already defined named_arguments, positional_arguments, and subcommands in this PR?

alexhancock: No. I am just seeing the wisdom in your original response design now 😄

api/README.md
```diff
@@ -0,0 +1,360 @@
...
+
+## /v0/servers
+
+### Request
+
+```http
+GET /v0/servers?limit=5000&offset=0
+```
+
+### Response
```
tadasant, line 11: Should be thoughtful about ordering here to keep CDN caching in place (e.g. order by server creation date, rather than alphabetical)

api/README.md
tadasant: https://github.com/modelcontextprotocol/registry/pull/3/commits/df3ef54617ac95453d4c4fdf6bceaca297a66f7e

api/README.md
tadasant: https://github.com/modelcontextprotocol/registry/pull/3/commits/1669e453e8a0debfb2c216a0faa17bad5350d7ae

api/openapi.yaml
tadasant: https://github.com/modelcontextprotocol/registry/pull/3/commits/52cf2c11fb37a4309d2db6f31db730fee8c92e74

api/openapi.yaml
tadasant: https://github.com/modelcontextprotocol/registry/pull/3/commits/52cf2c11fb37a4309d2db6f31db730fee8c92e74

tadasant:
@sridharavinash I made a few changes I think are improvements:

https://github.com/modelcontextprotocol/registry/pull/3/commits/df3ef54617ac95453d4c4fdf6bceaca297a66f7e - add a `default_value` field for EnvironmentVariable

https://github.com/modelcontextprotocol/registry/pull/3/commits/1669e453e8a0debfb2c216a0faa17bad5350d7ae - rename `CommandArguments` to `Command` and allow for a `name` field (like `npx`, `uvx`, etc; so it's explicit what the command arguments are tied to)

https://github.com/modelcontextprotocol/registry/pull/3/commits/52cf2c11fb37a4309d2db6f31db730fee8c92e74 - simplify the notion of Repository. Original PR was from back when we were planning to tightly couple source code to our MCP server concept, but the DNS-based naming approach as superseded that.

https://github.com/modelcontextprotocol/registry/pull/3/commits/50d22a9bf01305fb81f100d172dbdc502bc993a5 - renamed `registries` and `registry_canonical` to `packages` and `package_canonical`; realized that naming makes a lot more sense

https://github.com/modelcontextprotocol/registry/pull/3/commits/e9ad7352dd35f905e6216b94cf760d6bd124d8a0 - removed license field; unnecessary denormalized data.

api/README.md
```diff
@@ -0,0 +1,359 @@
...
+{
+  "servers": [
+    {
+      "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
+      "name": "@modelcontextprotocol/servers/src/filesystem",
+      "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
+      "repository": {
+        "url": "https://github.com/modelcontextprotocol/servers",
+        "source": "github",
+        "id": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
```
sridharavinash, line 23: Is this repoID that's represented within GitHub or respective source control services? E.g the repo id here that's returned by the https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository ?

api/README.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+### Response
+
+```json
+{
+  "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
+  "name": "@modelcontextprotocol/servers/src/filesystem",
+  "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations.",
+  "version": "0.0.3",
```
sridharavinash: Do we want extra fields to denote that this is the latest version ? Also do we care about when it was published to the mcp registry?

api/README.md
```diff
@@ -0,0 +1,359 @@
...
+```http
+GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
+```
+
+### Response
+
+```json
+{
+  "id": "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1",
+  "name": "@modelcontextprotocol/servers/src/filesystem",
```
sridharavinash, line 50: Do we want a separate field to denote that this publish was by a verified org / authors? Alternatively are we implying any publish to makes it here is from authors who are verified?

api/README.md
tadasant: Yes, the idea here is that we want an immutable reference available (such as for the case when a GH repo is transferred or name is changed)

I'm not entirely sure the `url` is fully necessary, but when we do build a registry UI I'm sure we will want a "click here to see source code" type button so it's reasonable to include here.

api/README.md
tadasant: I think we should imply any publish is verified; too many spam vectors open up if we allow a notion of unverified publication.

api/README.md
tadasant: I liked your `version_detail` augmentation here, sorry I didn't update in here yet

For anyone reading: https://github.com/modelcontextprotocol/registry/blob/d32b528104da5915ad776ee3ea0a4f741d9b1307/scripts/test_publish.sh#L64-L67

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+    Package:
+      type: object
+      required:
+        - registry_name
+        - name
+      properties:
+        registry_name:
+          type: string
+          enum: [npm, docker, pypi, homebrew]
+          example: "npm"
```
connor4312: I suggest that there is a "shell" type package that _just_ a shell script without a registry. (Or maybe registry is "shell".)

The type of registry is unbounded, and if people only provide MCP servers with their language's package manager, then client support will always be a patchwork. However, it should be easy for packages to provide a shell command fallback (e.g. manually using npx or uvx) and doing so will ensure any client can run any MCP server.

So I think there should be a shell type and it be recommended or even required for publishers to supply.

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+          type: string
+          example: "@modelcontextprotocol/servers/src/filesystem"
+        description:
+          type: string
+          example: "Node.js server implementing Model Context Protocol (MCP) for filesystem operations."
+        repository:
+          $ref: '#/components/schemas/Repository'
+        version:
+          type: string
+          description: Dynamically pulled from registry_canonical reference
```
connor4312: Imo this should be required at the package level, or clarified. As this reads, it applies only to the canonical registry and the version of the package on other registries is basically unknown.

I think that we must have the version associated with the package because the package defines the args and environment variables specific to that version of the server. And it shouldn't be the case that I have the same "MCP server version" with a config that works on one machine and fails on another (if I happen to have an old version in my `npx` cache and npm isn't the canonical registry)

And of course if the (non-canonical) version is unspecified there's the perennial security risk of automatically pulling in new versions which could be malicious.

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+        - url
+      properties:
+        transport_type:
+          type: string
+          enum: [streamable, sse, ws]
+          example: "sse"
+        url:
+          type: string
+          format: uri
+          example: "https://mcp-fs.example.com/sse"
```
connor4312, line 327: Users will want a way to configure headers here too. I think the same direction as what exists for arguments is probably good

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+    
+    Remote:
+      type: object
+      required:
+        - transport_type
+        - url
+      properties:
+        transport_type:
+          type: string
+          enum: [streamable, sse, ws]
```
connor4312: `ws` isn't an official transport type, should it be part of the registry spec?

```suggestion
          enum: [streamable, sse]
```

api/openapi.yaml
tadasant: Sorry, I agree with your thinking here: the idea of pinning the server's version to the version in `registry_canonical`  (now called `package_canonical`) was an earlier idea I now agree wouldn't work well, and I forgot to update the description here.

It wouldn't work for the reasons you mentioned + there are use cases where you might e.g. make a typo in the `Command`, publish, and then want to update the `Command` without bumping the registry package version. Thanks for flagging.

api/openapi.yaml
tadasant: In the shell case, where do you imagine the shell script code to be stored? Somewhere in this metaregistry?

I worry about storing free-form code like that - then we are exposed to that code potentially being a gateway to performing any kind of programmatic (potentially malicious) actions, and would need to build some additional systems to, at minimum, react to attacks. Versus, for example, relying on `npm` to take down packages that have been reported to be victims of supply chain attacks. 

It would also make it difficult to pass standardized command arguments (like API keys) if we omit `command` for this entry, unless I'm misunderstanding the suggestion.

Instead of storing a static shell script that would typically contain a command like `npx org@myservername`, any reason to not suggest clients just build that script on the fly based on what they see in `packages.command`?

api/openapi.yaml
tadasant: Right, let's skip WS, thanks

api/openapi.yaml
tadasant: What's an example of a header you might expect here? I was thinking maybe we'd want some notion of auth requirements here, e.g. OAuth - Auth Code vs. OAuth - Client Creds vs. No Auth, but that's a different level of abstraction than headers so curious what you are going for.

connor4312: Authorization would definitely be the most common. Even in a world where MCP auth is widely adopted, there'll always be some cases and servers who want to do basic API key types things for any variety of reasons. I've also seen some APIs that do versioning via `X-Api-Version` headers.

tadasant: Got it, fair - it does feel analogous to `Command` for local servers. This makes sense, will plan to add in next iteration. Thanks!

api/openapi.yaml
connor4312: > any reason to not suggest clients just build that script on the fly based on what they see in packages.command?

I think they cannot both do this and properly respect the package-level `verson`, assuming one gets added. In your example you have...

```
      "registry_name": "npm",
      "name": "@modelcontextprotocol/server-filesystem",
      "command": {
        "name": "npx",
        "subcommands": [],
        "positional_arguments": [
          {
            "position": 0,
            "name": "package",
            "description": "NPM package name",
            "default_value": "@modelcontextprotocol/server-filesystem",
            "is_required": true,
            "is_editable": false,
            "is_repeatable": false,
            "choices": []
          },
```

but assuming there is a package-level version, the "NPM package name" should not get blindly added to the command because the version needs to be mixed in there as well. If the arguments are blindly built then the version between what the command asks for and the nominal package version can drift. Which is all kinds of confusion.

In our past thinking for the VS Code registry, arguments only specify the arguments for _the package itself_ and not `npx`/`uvx`/whatever runner the client uses. Instead, it's up to the client to use a tool of its choice to retrieve the package at the specific version from the registry. This is also good because in cases where there is not a canonical way to install and run packages in that language (e.g. in Python, uvx is just one way to do things, and dotnet does not yet have any official such tool) the client is able to feature detect what capabilities are on the machine and do the right thing.

That then presents the need for a shell command that's "just a command" else clients are not guarenteed to be able to run MCP servers that are found in the repository. I think this can be specified easily enough with the existing structure, e.g.

```

  "packages": [
    {
      "registry_name": "shell",
      "command": {
        "positional_arguments": [
          {
            "position": 0,
            "name": "npx",
            "default_value": "@modelcontextprotocol/server-filesystem@1.2.3",
            "is_required": true,
            "is_editable": false,
          },
          {
            "position": 1,
            "name": "path",
            "description": "Path to access",
            "default_value": "/Users/username/Desktop",
            "is_required": true,
            "is_editable": true,
            "is_repeatable": true,
            "choices": []
          }
        ],
```

> It would also make it difficult to pass standardized command arguments (like API keys) if we omit command for this entry, unless I'm misunderstanding the suggestion.

Can you clarify the difficulty here?

api/README.md
olaservo: Is this `version` intended to match the package version when applicable, and should it be verified against the server version?

Bringing this up here related to [this](https://github.com/orgs/modelcontextprotocol/discussions/340).

tadasant: We were actually planning to have a separate notion of version. Thoughts here: https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2074198582

Curious to hear if you think we should reconsider

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+          type: array
+          items:
+            $ref: '#/components/schemas/EnvironmentVariable'
+    
+    Remote:
+      type: object
+      required:
+        - transport_type
+        - url
+      properties:
```
annaji-msft, line 319: An optional, OAuth server configuration could help here as well specifically for those MCP authorizations servers that do not support Dynamic Client Registration. 

Described in the [spec](https://modelcontextprotocol.io/specification/draft/basic/authorization#2-5-dynamic-client-registration ), Clients could hardcode the ClientId for the authorization server. 

The metadata in the registry can help guide consumers a ClientId is expected for the specific MCP server as well as where to acquire it, usually link to service provider developer portals. 

For reference see issue - https://github.com/microsoft/vscode/issues/247759 and also briefly touched on DCR topic and clientId acquisition in the [blog](https://aaronparecki.com/2025/04/03/15/oauth-for-model-context-protocol)

Could be inspired from [OAS](https://swagger.io/docs/specification/v3_0/authentication/oauth2/).

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+          type: string
+          format: uri
+          example: "https://mcp-fs.example.com/sse"
+    
+    ServerDetail:
+      allOf:
+        - $ref: '#/components/schemas/Server'
+        - type: object
+          properties:
+            package_canonical:
```
luoxiner, line 334: What‘s the difference between the package_canonical and packages[]. registry_name?

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+              type: string
+              enum: [npm, docker, pypi, crates]
+              example: "npm"
+            packages:
```
luoxiner, line 338: Is packages required for all mcp servers? There might be MCP servers that only provide an endpoint for SSE/streaming MCP API and do not offer a package.

api/openapi.yaml
tadasant: Ah I see, sorry I misunderstood your suggestion; I thought you were suggesting submission of a shell script separate from this shape. I see per your example that we'd still well-define the script within `command`.

> assuming there is a package-level version, the "NPM package name" should not get blindly added to the command because the version needs to be mixed in there as well. 

That's a good point. It should certainly be at least best practice to encode a version inside `command.positional_arguments.default_value`, like you did in your second example, `@modelcontextprotocol/server-filesystem@1.2.3`. Would be even better to enforce that somehow, so people aren't creating that implicit `@latest` inconsistency across registry version vs. MCP version... Need to think more on that.

> In our past thinking for the VS Code registry, arguments only specify the arguments for the package itself and not npx/uvx/whatever runner the client uses. Instead, it's up to the client to use a tool of its choice to retrieve the package at the specific version from the registry. This is also good because in cases where there is not a canonical way to install and run packages in that language (e.g. in Python, uvx is just one way to do things, and dotnet does not yet have any official such tool) the client is able to feature detect what capabilities are on the machine and do the right thing.

The thinking behind being explicit with `command.name` was that it limits the surface area for malicious commands and spam (there is only so much you can do with an enum type that has to be one of a few dozen options, like `npm`, `uvx`, etc.).

That said, a prior iteration of this shape did entirely omit `command.name`, in an attempt to do as you said:

> arguments only specify the arguments for the package itself and not npx/uvx/whatever runner the client uses

An example use case I found troubling was the case where some MCP servers will fail to run if you just do:

```
npx org@myserver --stdio
```

But will succeed if you do:

```
npx -y org@myserver --stdio
```

`-y` is an `npx` argument here.

But I don't feel strongly on this. Your points here make me think we should go back to omitting `name` and just let client apps figure out the `-y` nuances if they decide to use `npx` to run npm packages.

@alexhancock I think it was you and I that had the discussion that led to the addition of `name`. What do you think?

Found that discussion here: https://github.com/modelcontextprotocol/registry/pull/3/files#r2018897939

I am open to rethinking this. Maybe we could solve both this and the prior version-related concern by doing the following:
- Remove `command.name`
- Instead of putting the package name as the 0th positional argument, omit that
- Introduce a `version` field (required?)
- That means the client can create its preferred cocktail of `npx` or `uvx` or `pip`, whatever, knowing the package name and version should be enough to construct the first part of the command in a repeatable way
- Then the rest of the MCP-specific args can be constructed from this API shape (like Connor said, "arguments only specify the arguments for the package itself")

I think that approach should address your reason for wanting `shell`:

> I think they cannot both do this and properly respect the package-level verson, assuming one gets added.

So a client could see, for example:

```
      "registry_name": "npm",
      "name": "@modelcontextprotocol/server-filesystem",
      "version": "1.2.3",
      "command": {
        "subcommands": [],
        "positional_arguments": [
          {
            "position": 0,
            "name": "root_path",
            "description": "Root path for filesystem access",
            "default_value": "/projects",
            "is_required": true,
            "is_editable": true,
            "is_repeatable": false,
            "choices": []
          }
```

And then construct, based on its preference for `npx`:

```sh
npx -y @modelcontextprotocol/server-filesystem@1.2.3 /projects
```

api/openapi.yaml
tadasant: `packages[]` can have multiple entries, so `package_canonical` tells us which one of them is the canonical (preferred) one to use.

That said, maybe we can ditch `package_canonical`. We originally had it because we were pinning `version` across the MCP servers <> registry package version. We're not currently planning to do that, so I don't think it's necessary. It might still be a nice hint to MCP clients that "you should use this entry by default", but I don't see that as a strong need and we should probably simplify and skip. Thanks for the question!

api/openapi.yaml
tadasant: Yes, it is an optional field in this spec.

api/openapi.yaml
tadasant: I'm thinking it might be helpful to lift the Remote shape into a separate PR/discussion. This Remote shape has not had much group input, and I haven't invested too much time in thinking through it + it was drafted a couple months ago before remotes grew in popularity and support. Agree that thinking through what auth-related concerns are worth including here is a good idea.

annaji-msft: Thanks @tadasant happy to participate in the next iteration.

api/openapi.yaml
luoxiner: Thank you for your explanation. I've noticed that previous conversations have brought up the property relationship. I agree that we can skip it and let the client decide which package is the best fit for the user's environment.

api/openapi.yaml
luoxiner: Is the "version" here related to the [#/definitions/Implementation](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/c255349bd98d9dc73f3579c828d195bd09c0bee8/schema/2025-03-26/schema.json#L586) version field?

Does the "version" field here convey the differences resulting from different implementations of the MCP server?
For instance, adding a tool, removing a tool, or modifying the functions of a tool. When the capabilities of the MCP server itself change, the version here needs to be adjusted accordingly. 

Does the specific version information of the package need to be directly included in the Package?
For the same version of the MCP server, there are different versions of package implementations, but their capabilities are the same.

api/openapi.yaml
tadasant: This [point about licensing](https://github.com/modelcontextprotocol/registry/discussions/11#discussioncomment-13065727) is actually a good example of a reason it is helpful to maintain a structured reference to a registry package (versus shell entry) -- MCP clients are then empowered to ensure that MCP servers they are pulling in have licenses that are appropriate for their users to be engaging with.

connor4312: > So a client could see, for example:

Yep, exactly. Your example lgtm and the summary accurately reflects my thinking and suggestions on this 🙂 

> Introduce a version field (required?)

Yes, I think this should should be required (as per my other comments)

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+      description: Returns detailed information about a specific MCP server
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: Unique ID of the server
+          schema:
+            type: string
+            format: uuid
+        - name: version
```
luoxiner, line 61: I reviewed the current Go implementation and noticed that a UUID is generated every time an MCP server is published, ensuring that each version of an MCP server has a unique ID. 

However, it seems that the id here is associated only with the MCP server itself, rather than with the server + version, which is causing some confusion for me. 

And I think the current specification is more reasonable for a clear semantic understanding.

api/openapi.yaml
sandy081:
```json
      "registry_name": "npm",
      "name": "@modelcontextprotocol/server-filesystem",
      "version": "1.2.3",
      // ... args
```

+1 to the above syntax and this aligns with our past thinking in VS Code as @connor4312 mentioned. 

Since we are delegating the command to run to the client, how about simplifying the command/subcommands/positional_arguments/named_arguments to just `arguments` array as a property under `package` object?  For example an argument object schema can be as follows:

```json
{
			"type": "object",
			"description": "Describes an argument that can be passed to the server",
			"required": [
				"name"
			],
			"properties": {
				"name": {
					"type": "string",
					"description": "Name of the argument"
				},
				"value": {
					"oneOf": [
						{
							"type": "string",
							"description": "Value of the argument"
						},
						{
							"type": "array",
							"description": "List of values for the argument",
							"items": {
								"type": "string"
							}
						}
					]
				},
				"description": {
					"type": "string",
					"description": "Description of the argument"
				},
				"default": {
					"type": ["string", "array"],
					"description": "Default value of the argument"
				},
				"is_required": {
					"type": "boolean",
					"description": "If true, the argument is required"
				},
                                 "is_value_required": {
					"type": "boolean",
					"description": "If true, the value is required for the argument"
				},
				"choices": {
					"type": "array",
					"description": "List of possible values for the argument",
					"items": {
						"type": "string",
						"description": "Possible value for the argument"
					}
				}
			}
		}
```

A sample manifest can be as follows:

```json
      "registry_name": "npm",
      "name": "@modelcontextprotocol/server-filesystem",
      "version": "1.2.3",
     "arguments": [
          {
            "name": "root_path",
            "description": "Root path for filesystem access",
            "default": "/projects",
            "is_required": true,
            "is_value_required": true,
          }
      ]        
```

This will let the client just to pass the arguments passed in the manifest as is to the server (including headers, env variables) by validating some flags like `is_required`. 

What do you think?

api/openapi.yaml
sandy081: Yes, it is similar to local package arguments. This is how we envisioned for example:

```json
"remote": {
			"type": "object",
			"description": "Remote server configuration",
			"required": [
				"url"
			],
			"properties": {
				"url": {
					"type": "string",
					"format": "uri",
					"description": "URL of the remote server"
				},
				"transport_type": {
					"type": "string",
					"description": "Transport type for the remote server",
					"enum": [
						"streamable",
						"sse"
					]
				},
				"headers": {
					"type": "array",
					"description": "Headers to be sent to the remote server",
					"items": { "$ref": "#/definitions/argument" }
				}
			}
		}
```

`#/definitions/argument` is the schema for the argument that is also used for local package arguments.

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+          type: string
+          example: "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
+    
+    Server:
+      type: object
+      required:
+        - id
+        - name
+        - description
+        - repository
```
sandy081, line 111: I assume the `version` property will be made required here as per the below discussion.

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+      properties:
+        id:
+          type: string
+          format: uuid
+          example: "a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1"
+        name:
```
sandy081, line 118: Should we also have a `displayName` that can be readable to the user, say `File System MCP`?

api/openapi.yaml
```diff
@@ -0,0 +1,324 @@
...
+          example: "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
+    
+    Server:
+      type: object
+      required:
+        - id
+        - name
+        - description
+        - repository
+      properties:
```
sandy081, line 113: It should be good to include optional `icon` property in the first iteration so that mcp servers are nicely displayed in the client. Eg schema:

```
"icon": {
				"type": "object",
				"description": "Icon of the server",
				"required": [
					"url"
				],
				"properties": {
					"url": {
						"type": "string",
						"description": "URL of the icon"
					}
				}
			}
```

sandy081:
@tadasant / @sridharavinash 

May I know if the server definition supports a required `publisher` property? I remember @sridharavinash mentioning that this can be derived from the repository owner or org. If so then it makes sense that the server registry populates the publisher information in the server list and server response.

api/openapi.yaml
tadasant: @sandy081 I think it'd be reasonable to rename `command` to just be `arguments` or `package_arguments`, and potentially getting rid of `subcommand` as well. The main use case for subcommand was with docker MCP servers -- we should review some examples there to make sure this change wouldn't break them.

I'm not sure getting rid of the distinction between positional vs. named arguments would work. Your proposal seems like it would eliminate the possibility of positional arguments, unless I am misreading?

api/openapi.yaml
tadasant: > Is the "version" here related to the [#/definitions/Implementation](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/c255349bd98d9dc73f3579c828d195bd09c0bee8/schema/2025-03-26/schema.json#L586) version field?

Good callout - I think yes, it should match, but we haven't deeply considered the relationship.

> Does the "version" field here convey the differences resulting from different implementations of the MCP server?
> For instance, adding a tool, removing a tool, or modifying the functions of a tool. When the capabilities of the MCP server itself change, the version here needs to be adjusted accordingly.

Yes, that's the idea. The published "mcp.json" file should be immutable. So any change in that (including a source code change, which would drive the kinds of changes you listed) should be communicated with a version bump and new mcp.json publication.

> Does the specific version information of the package need to be directly included in the Package?
For the same version of the MCP server, there are different versions of package implementations, but their capabilities are the same.

I'm not sure I follow what you are asking here, mind rephrasing?

api/openapi.yaml
tadasant: The implementation is very much a work in progress, so it's probably just a bug, but cc @sridharavinash to confirm

api/openapi.yaml
tadasant: I am generally very hesitant to add free-form fields due to increased exposure for spam and abuse, but this is probably a reasonable one to add, I agree. We can sanitize it to only allow alphanumeric + spaces + maybe `:` and `-`, nothing else.

api/openapi.yaml
tadasant: This one is less clear to me than `displayName`. I see the need, but it is riskier (stuffing URL's is more prone to spam problems + rendering user input in a `src` attribute has more potential security holes) and may be a slippery slope: should we allow for storing different sized icons for different displays? Should we allow some icon to represent the publisher alongside the icon for the MCP server itself?

But, I hear you that if we don't include it, there isn't a great way for a server maintainer to communicate a visual icon through to consumers. So maybe we just need to deal with the complexity here.

Would it be reasonable for you to consider instead using, for example, the favicon of the reverse-DNS name? At least initially, and maybe we could add `icon` support natively later on when there are more eyeballs on this and we can iteratively improve it?

tadasant:
> @tadasant / @sridharavinash
> 
> May I know if the server definition supports a required `publisher` property? I remember @sridharavinash mentioning that this can be derived from the repository owner or org. If so then it makes sense that the server registry populates the publisher information in the server list and server response.

There was no current plan to include a `publisher` property -- it is just implied through the reverse-DNS `name` field. If you think it would be helpful to include, I would suggest opening a Discussion as that would be some net new requirements work that I think it'd be helpful to get more input on.

sandy081:
>  There was no current plan to include a publisher property -- it is just implied through the reverse-DNS `name` field.

Is the `name` field meant for the name of the MCP server, eg: `@modelcontextprotocol/server-filesystem` in which case the owner/publisher would be `modelcontextprotocol` right? If so, it would be helpful if the server registry resolves this information into a separate property with necessary details like `publisher` and `publisherDisplayName`, for eg

```json
"publisher": {
    "name": "modelcontextprotocol",
    "displayName": "Model Context Protocol",
    "verified": true
}
```

I think the registry is the right one to resolve this information and leaving this to clients could cause falsy information. It is important for a user to know the publisher/owner of an MCP server. Hence it would be great if this can be included in the initial version.

api/openapi.yaml
sandy081: Agreed that passing URL's is risky. To prevent this the MCP server registry should scan these URLs and filter them for security like not allowing SVGs and including icons while registering an MCP server. 

> should we allow for storing different sized icons for different displays?

Yeah we should going forward, but for the first cut an icon url to display in the gallery should be sufficient.

> Would it be reasonable for you to consider instead using, for example, the favicon of the reverse-DNS name? At least initially, and maybe we could add icon support natively later on when there are more eyeballs on this and we can iteratively improve it?

If this is the initial direction, then I would like the server to do this reverse engineering and provide the icon url in the server list response. Again, i am trying to prevent inconsistencies of representing an mcp server by letting server providing this information

api/openapi.yaml
luoxiner: I'm sorry for the confusion in my previous question. What I meant is that the version here should not depend on any specific package version, and the version of each package should be defined solely in the Package.version field.

luoxiner: I'm sorry for the confusion in my previous question. What I meant is that the version here should not depend on any specific package version, and the version of each package should be defined solely in the Package.version field.

But I believe you already addressed the question in your previous reply.

api/openapi.yaml
alexhancock: @connor4312 @tadasant Late to the party on this thread but the conclusion you got to in https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2077835152 makes sense to me and I agree

On the topic of flattening into one list of args, it could simplify things. I think it could still work for positional as well as named arguments. For example we use a single arg list in our current approach to configuring the client to run MCP servers in goose. Example YAML:

```yaml
figma:
  args:
    - -y
    - 'imaginary-figma-mcp@latest'
  bundled: true
  cmd: npx
  ... (rest omitted)
```

But I may be missing something. What issue do you see @tadasant ?

connor4312: The actual client config of servers is out of scope of the registry spec, but my plan for VS Code is to have something like (on a data level)

```
{
  "server": "imaginary-figma-mcp@1.2.3",
  "type": "npm", // optionally diambiguate
  "config": { "apiKey": "something-here" }
}
```

...and generate the command line behind the scenes such that versions can be updated and largely tolerant to argument changes so long as inputs are stable.

alexhancock: I agree @connor4312. As long as there is enough information in the responses to form commands given some assumptions of popular ways of running packages from various registries, we're meeting the goal of the project IMO.

I just posted this as an example to show that positional and named arguments can be represented in a single flat list as long as the client constructing the command respects/uses the ordering (re: @tadasant's question as to whether @sandy081's simplification of the schema could still represent positional arguments).

api/README.md
tadasant: ea6623e1a8a6efe87fb2e879ef953997b0f2016e

api/openapi.yaml
tadasant: e7f8916d504365e185e02ddb45a0807a58cd38a5

api/openapi.yaml
tadasant: Removed the errant language here ea6623e1a8a6efe87fb2e879ef953997b0f2016e

api/openapi.yaml
tadasant: I'm having a doubt on removing `subcommand`: how would we then make the [filesystem reference server work with Docker](https://github.com/modelcontextprotocol/servers/tree/HEAD/src/filesystem#docker)?

```json
{
  "mcpServers": {
    "filesystem": {
      "command": "docker",
      "args": [
        "run",
        "-i",
        "--rm",
        "--mount", "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
        "--mount", "type=bind,src=/path/to/other/allowed/dir,dst=/projects/other/allowed/dir,ro",
        "--mount", "type=bind,src=/path/to/file.txt,dst=/projects/path/to/file.txt",
        "mcp/filesystem",
        "/projects"
      ]
    }
  }
}
```

Maybe this example is a major outlier and we shouldn't design around it, but I wouldn't feel great about not even supporting our official reference servers.

What if we tried to simplify using a templating solution? e.g. something like:

```
{
  "id": "filesystem-67890",
  "name": "@modelcontextprotocol/servers/src/filesystem",
  "description": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations",
  "repository": { "url": "https://github.com/modelcontextprotocol/servers", "source": "github", "id": "d94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9" },
  "version_detail": { "version": "1.0.2", "release_date": "2023-06-15T10:30:00Z", "is_latest": true },
  "package_canonical": "docker",
  "packages": [
    {
      "registry_name": "docker",
      "name": "mcp/filesystem",
      "command_templates": {
        "docker": "docker run -i --rm --mount {{MOUNT_CONFIG}} {{IMAGE_NAME}} {{ROOT_PATH}}",
        "podman": "podman run -i --rm --mount {{MOUNT_CONFIG}} {{IMAGE_NAME}} {{ROOT_PATH}}"
      },
      "preferred_command": "docker",
      "variables": [
        {
          "name": "MOUNT_CONFIG",
          "description": "Mount a volume into the container",
          "default_value": "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
          "is_required": true,
          "is_editable": true,
          "is_repeatable": true
        },
        {
          "name": "IMAGE_NAME",
          "description": "Docker image name",
          "default_value": "mcp/filesystem",
          "is_required": true,
          "is_editable": false
        },
        {
          "name": "ROOT_PATH",
          "description": "Root path for filesystem access",
          "default_value": "/projects",
          "is_required": true,
          "is_editable": false
        }
      ],
      "environment_variables": [
        {
          "name": "LOG_LEVEL",
          "description": "Logging level (debug, info, warn, error)",
          "required": false,
          "default_value": "info"
        }
      ]
    }
  ]
}
```

api/openapi.yaml
tadasant: f9b0f3121a1021e7053e05bae9bd17f189568193

api/openapi.yaml
tadasant: I just added `version` for now, but acknowledge this is not solved: 8ee389684f6d3e36b72b28ed4d2bd6fd40c62519

tadasant:
Thanks for all the feedback so far @alexhancock @connor4312 @sandy081 @luoxiner @annaji-msft 

I am thinking we have covered a good amount of breadth to start, and at this point there need to be some more detailed discussions about specific points, and it's feeling unwieldy to try to handle in a single monster PR here.

How would you all feel if we merged this OpenAPI spec and then spun up separate PR's/Issues to drill down on the open questions we've identified?

Specifically:
- Incorporating [OAuth details](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2076841333) into the remote servers shape (@annaji-msft). This is probably noncontroversial, just needs to be worked through.
- Incorporating [headers for remotes](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2074202871) (@connor4312). This is probably noncontroversial, just needs to be worked through.
- Consider including "marketing" data, like [displayName](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2079321488) and [icon](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2079330007). I had initially thought displayName might make sense, but I am now leaning towards thinking we should exclude both. Can share my thoughts in an Issue, would like to get more opinions on this one (@sandy081)
- Consider including [publisher data](https://github.com/modelcontextprotocol/registry/pull/3#issuecomment-2863341238). Can discuss more in an issue, but I think there a major issues here with the notion of "trust". We actually have no guarantees that the publisher is who they say they are. We've got the reverse DNS system planned, but all that is is the namespace. (@sandy081)
- Figuring out the [proper way to model commands](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2080826697). I'm starting to think we should simplify into a templating approach. (@connor4312 @alexhancock @sandy081)

Besides these above open points, I've addressed feedback on the other comments.

If you all think it's a good idea to merge this and start to separate out the discussions, I'll start an Issue for each of these, and if someone is eager to finalize they can open a PR without being blocked on me. Does that work?

cc @toby @digitarald

evalstate:
Thanks @tadasant  -- this is brilliant work btw. I was going to ask how to contribute, so I definitely +1 on getting this merged.

api/openapi.yaml
connor4312: > To prevent this the MCP server registry should scan these URLs and filter them for security like not allowing SVGs and including icons while registering an MCP server.

This is not a trivial ask. It's also not really possible to do for external URLs, consider I might have `example.com/icon` which is a png when I submit the package but I change it to a an SVG later. Say that to fix that we spec that the URL must have an extension with a mimetype that matches the URL. I still control the URL so I later could make `example.com/icon.png` redirect to `example.com/evil.svg`, which HTML-based clients in `<img` will follow by default. Say we fix that with some really aggressive server behavior to continuously validate image. I could be smart and see the IP addresses the registry requests from and serve that a png and clients an svg. This is an very deep hole and there may be 3P servers as well which don't do any validation or have their own bugs.

Imo it should be on the client to validate the images it retrieves from the server and do the right thing.

For VS Code, we should probably run our own image proxy server anyway (like github camo) to avoid exposing user IP addresses to third parties and validation can be done there. But we should not require this of registry implementations.

connor4312:
> Figuring out the https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2080826697. I'm starting to think we should simplify into a templating approach.

I considered this in my initial draft of how a registry manifest would work. I thought of templating args, environment variables, and headers ('inputs') with Mustache-like syntax:

```js
"docker": {
  "configuration": {
    "required": ["github-token"],
    "github-token": {
      "type": "string",
      "description": "GitHub Personal Access Token",
      "secret": true
    }
  },
  "name": "ghcr.io/github/github-mcp-server",
  "args": ["-e", "GITHUB_TOKEN"],
  "env": { "GITHUB_TOKEN": "{{github-token}}" }
}
```

In my initial draft I stated that inputs that reference an optional configuration that's missing would be omitted. (Is this is enough? I'm not sure.) Then there are the usual niggles with templating, like how we handle escaping and so on. Configuration being a JSON schema object gives some more power but opens up to nested objects and arrays and I'm not sure how/if those should be handled (or if we'd spec it such that only primitives are allowed.) It also doesn't map cleanly onto repeated arguments unless we get deep into the weeds of templating.

Overall I liked the explicit version more. It kept the world better-bounded and gave clear answers to the above unknowns.  Sandeep [linked](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2079263686) roughly where we ended up on our version of the spec.

I think docker mounts overall may need formal specification. Many clients, like VS Code, will have the notion of the "workspace folder" which is probably what the user wants to mount by default into their container. And additionally thinking of a world where we specify the arguments for the server and not the runtime (i.e. not npx/uvx/docker), perhaps the config for the filesystem server just looks like

```
  "packages": [
    {
      "registry_name": "docker",
      "name": "mcp/filesystem",
      "version": "1.0.2",
      "command": {
        "name": "docker",
        "mounts": [
          {
            "description": "Folder to serve as the root of the filesystem",
            "default_flags": "ro",
            "default_mount_point": "/projects",
            "repeated": true,
          }
        ],
        "positional_arguments": [
          {
            "name": "root_path",
            "description": "Root path for filesystem access",
            "default_value": "/projects"
          }
        ]
      }
    }
  ]
}
```

... where the default_mount_point corresponds to the default of the positional argument which should be good enough hint for users to know what to do (or even for clients to correlate them automatically.)

This also can let us guarentee as a client that we won't have unintended side-effects for the host and can basically run Docker containers without requiring super explicit "are you sure you want to run this untrusted command from the internet on your computer" confirmations. Versus a template, we control the docker command that's executed and the mounts that it's given, so we can ensure it doesn't just try and mounnt ~/.ssh/id_rsa and do bad stuff.

api/openapi.yaml
annaji-msft: for tracking https://github.com/modelcontextprotocol/registry/issues/25

digitarald (approved):
LGTM based on https://github.com/modelcontextprotocol/registry/pull/3#issuecomment-2864970637

api/openapi.yaml
tadasant: https://github.com/modelcontextprotocol/registry/issues/26

api/openapi.yaml
tadasant: To continue discussion: https://github.com/modelcontextprotocol/registry/discussions/27

api/openapi.yaml
tadasant: To continue discussion: https://github.com/modelcontextprotocol/registry/discussions/27

api/openapi.yaml
tadasant: https://github.com/modelcontextprotocol/registry/issues/28

tadasant:
Follow-ups:
- https://github.com/modelcontextprotocol/registry/issues/26 - HTTP Headers. Pretty much shovel-ready if anyone wants to self-assign
- https://github.com/modelcontextprotocol/registry/issues/25 - Thanks @annaji-msft for opening this, I think it's close to shovel-ready
- https://github.com/modelcontextprotocol/registry/issues/28 - `command` shape. I think we should solicit some more opinions on this from folks before locking in a conclusion, but I think @connor4312 and @sandy081 have done a good amount of thinking here that it is probably close to being shovel-ready
- https://github.com/modelcontextprotocol/registry/issues/29 - Publisher metadata. I don't think we should proceed with publisher metadata, but feedback welcome
- https://github.com/modelcontextprotocol/registry/discussions/27 - icon & displayName. I don't think we should proceed with this, but feedback welcome

If anyone wants to self-assign and drive any of this forward, please feel free. Unless I assign myself to an issue, I'm not actively working on it yet.

api/README.md
```diff
@@ -0,0 +1,383 @@
...
+                "description": "Mount a volume into the container",
+                "default_value": "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
+                "choices": []
+              }
+            ]
+          }
+        ],
+        "positional_arguments": [
+          {
+            "position": 0,
```
tiovikram, line 156: Explicitly specifying the position as a key in the JSON seems like an anti-pattern given that the parent object (an array) is inherently ordered by nature.

What are your thoughts about this? Any chance I could open this up as a follow-up issue if you think it is pertinent?

connor4312: You may be interested in looking at this PR: https://github.com/modelcontextprotocol/registry/pull/33

tiovikram: I did take a look and I think #33 may be simplifying the commands a little too much. However, I can move that discussion there instead.

Issue #28, opened 2025-05-09
Simplify `command` in OpenAPI spec


tadasant:
Commentary on this so far:
- https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2080826697
- https://github.com/modelcontextprotocol/registry/pull/3#issuecomment-2866620975

At minimum, I think we need to make sure we can model all the official reference servers and their corresponding registries (pypi, npm, docker). We should also make sure we support the most obvious package registries for any official SDK's language (e.g. Rubygems, etc.).

Options we seem to have:
1) Keep a pretty complex `command` structure as is in the OpenAPI spec at the time of this writing, plus extend it somehow with registry package version so it we can are not hardcoding `package@latest` into the command. This field unwieldly.
2) Remove `command.name` and simplify to just `arguments`, eliminating the differentiation between `named_arguments`, `positional_arguments`, and `subcommand`. It's probably that docker / mounts need some sort of [carve-out for this](https://github.com/modelcontextprotocol/registry/pull/3#issuecomment-2866620975). Are there other cases where this will be a problem?
3) Some sort of [templating solution](https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2080826697)

@connor4312 @sandy081 it seems like you have done a fair bit of exploration here, so I'm inclined to let you drive this to a conclusion. Sounds like you feel pretty strongly on (2)?

connor4312:
Done in #33

PR #33, opened 2025-05-10
feat: refine representation of commands in the API

connor4312:
Refs #28. After some thinking, this tries for a 'best of both worlds' approach. Simple arguments can be present essentially as the original spec recommended. As discussed, we don't specify the args for `npx` and friends directly, so a simple package is merely...

```yaml
- name: my-api-mcp
  # ...
  packages:
    - type: npm
      name: "@c4312/my-api-mcp"
      version: "0.0.1"
      arguments:
        # Named inputs in 'arguments' generate a `--name=<value>`
        - type: named
          name: "api-key"
          is_required: true
          is_secret: true
```

However, I think we need to be able to specify runtime args in some cases, otherwise the registry eventually mirrors all docker configs and options which is not a great situation to be in. So, I suggest a `runtime_args` in the same format. Say I wanted that for npm, I could:

```yaml
- name: my-api-mcp
  # ...
  packages:
    - type: npm
      name: "@c4312/my-api-mcp"
      version: "0.0.1"
      runtime_arguments:
        # Arguments with a pre-specified `value` don't get prompted for input
        - type: positional
          value: "--experimental-eventsource"
```

But Docker mounts are... complex, and ultimately we may deal with cases that need multiple inputs in an argument like that. So for that I suggest the third kind of "templated" input that does basic replacements and has `properties` that follow the same general input schema:

```yaml
- name: server-filesystem
  # ...
  packages:
    - type: docker
      name: "mcp/filesystem"
      version: "0.0.1"
      runtime_arguments:
        - type: template
          description: Mount paths to access.
          name: mount_config
          is_required: true
          is_repeated: true
          # Templates support simple replacement of `{property_names}`
          template: "--mount=type=bind,src={host_path},dst={container_path}"
          properties:
            host_path:
              description: Path to access on the local machine
              is_required: true
              format: filepath
            container_path:
              description: Path to mount in the container. It should be rooted in `/project` directory.
              is_required: true
              default: "/project"

      arguments:
        - type: positional
          value: "/project"
```

And then client config ends up being pretty simple, for the above case for example a client could have this config:

```json
{
  "filesystem": {
    "server": "@modelcontextprotocol/servers/src/filesystem@1.0.2",
    "package": "docker",
    "settings": {
      "mount_config": [
        { "source_path": "/Users/username/Desktop", "target_path": "/project/desktop" },
        { "source_path": "/path/to/other/allowed/dir", "target_path": "/project/other/allowed/dir" },
      ]
    }
  }
}
```

Which might generate a CLI:

```sh
docker run \
  --mount=type=bind,src=/Users/username/Desktop,dst=/project/desktop \
  --mount=type=bind,src=/path/to/other/allowed/dir,dst=/project/other/allowed/dir \
  mcp/filesystem:1.0.2 \
  /project
```

I think this does a good job of accomplishing the goals we care about:

- We get nicely defined inputs that are easy to extend, easy to process, easy for clients to configure, and also easy to author.
- Packages can generally be updated in a backwards compatible way, given there are no changes to required arguments.
- Versioning is enforced as we don't require the package to specify the identifier to npx/uvx/etc.
- We _usually_ can break away from a dependency on a specific runtime. This promise is broken a bit with `runtime_arguments`, but most packages should not need this and those that do are those inherently more tightly coupled to a specific runtime like Docker (e.g. they would be pretty uncommon for Node/npx or Python/uvx.)
- We don't have a suple complex template system with loops and fallbacks and conditionals, like we would if we tried to have a single template string for a command.

cc @tadasant @sandy081 @digitarald @sridharavinash @toby

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -77,8 +77,7 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       ],
       "environment_variables": [
         {
-          "type": "named",
-          "name": "LOG_LEVEL",
+          "key": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
           "default": "info"
         }
@@ -90,9 +89,10 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "version": "1.0.2",
       "runtime_arguments": [
         {
-          "type": "template",
+          "type": "named",
           "description": "Mount a volume into the container",
-          "template": "--mount=type=bind,src={source_path},dst={target_path}",
+          "flag": "--mount",
+          "value": "type=bind,src={source_path},dst={target_path}",
           "is_required": true,
           "is_repeated": true,
           "variables": {
@@ -115,11 +115,10 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
           "name": "target_dir",
           "value": "/project",
         }
-      ]
+      ],
       "environment_variables": [
         {
-          "type": "named",
-          "name": "LOG_LEVEL",
+          "key": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
           "default": "info"
         }
@@ -163,8 +162,7 @@ API Response:
       "version": "1.0.2",
       "environment_variables": [
         {
-          "type": "named",
-          "name": "BRAVE_API_KEY",
+          "key": "BRAVE_API_KEY",
           "description": "Brave Search API Key",
           "is_required": true,
           "is_secret": true
@@ -217,9 +215,10 @@ API Response:
       "version": "1.0.2",
       "runtime_arguments": [
         {
-          "type": "template",
+          "type": "named",
           "description": "Mount a volume into the container",
-          "template": "--mount=type=bind,src={source_path},dst={target_path}",
+          "flag": "--mount",
+          "value": "type=bind,src={source_path},dst={target_path}",
           "is_required": true,
           "is_repeated": true,
           "variables": {
@@ -255,7 +254,7 @@ claude_desktop_config.json:
     "server": "@modelcontextprotocol/servers/src/filesystem@1.0.2",
     "package": "docker",
     "settings": {
-      "mount_config": [
+      "--mount": [
         { "source_path": "/Users/username/Desktop", "target_path": "/project/desktop" },
         { "source_path": "/path/to/other/allowed/dir", "target_path": "/project/other/allowed/dir,ro" },
       ]
--- a/api/README.md
+++ b/api/README.md
@@ -68,7 +68,7 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "package_arguments": [
         {
           "type": "positional",
-          "name_hint": "target_dir",
+          "value_hint": "target_dir",
           "description": "Path to access",
           "default": "/Users/username/Desktop",
           "is_required": true,
@@ -112,7 +112,7 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "package_arguments": [
         {
           "type": "positional",
-          "name_hint": "target_dir",
+          "value_hint": "target_dir",
           "value": "/project",
         }
       ],
@@ -238,7 +238,7 @@ API Response:
       "package_arguments": [
         {
           "type": "positional",
-          "name_hint": "target_dir",
+          "value_hint": "target_dir",
           "value": "/project",
         }
       ]
--- a/api/README.md
+++ b/api/README.md
@@ -65,122 +65,61 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-filesystem",
       "version": "1.0.2",
-      "command": {
-        "name": "npx",
-        "subcommands": [],
-        "positional_arguments": [
-          {
-            "position": 0,
-            "name": "package",
-            "description": "NPM package name",
-            "default_value": "@modelcontextprotocol/server-filesystem",
-            "is_required": true,
-            "is_editable": false,
-            "is_repeatable": false,
-            "choices": []
-          },
-          {
-            "position": 1,
-            "name": "path",
-            "description": "Path to access",
-            "default_value": "/Users/username/Desktop",
-            "is_required": true,
-            "is_editable": true,
-            "is_repeatable": true,
-            "choices": []
-          }
-        ],
-        "named_arguments": [
-          {
-            "short_flag": "-y",
-            "requires_value": false,
-            "is_required": false,
-            "is_editable": false,
-            "description": "Skip prompts and automatically answer yes",
-            "choices": []
-          }
-        ]
-      },
+      "arguments": [
+        {
+          "type": "positional",
+          "description": "Path to access",
+          "default": "/Users/username/Desktop",
+          "is_required": true,
+          "is_repeated": true
+        }
+      ],
       "environment_variables": [
         {
+          "type": "named",
           "name": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
-          "required": false,
-          "default_value": "info"
+          "default": "info"
         }
       ]
     },
     {
-      "name": "docker",
-      "package_name": "mcp/filesystem",
+      "registry_name": "docker",
+      "name": "mcp/filesystem",
       "version": "1.0.2",
-      "command": {
-        "name": "docker",
-        "subcommands": [
-          {
-            "name": "run",
-            "description": "Run the Docker container",
-            "is_required": true,
-            "subcommands": [],
-            "positional_arguments": [],
-            "named_arguments": [
-              {
-                "short_flag": "-i",
-                "requires_value": false,
-                "is_required": true,
-                "is_editable": false,
-                "description": "Run in interactive mode"
-              },
-              {
-                "long_flag": "--rm",
-                "requires_value": false,
-                "is_required": true,
-                "is_editable": false,
-                "description": "Remove container when it exits"
-              },
-              {
-                "long_flag": "--mount",
-                "requires_value": true,
-                "is_required": true,
-                "is_repeatable": true,
-                "is_editable": true,
-                "description": "Mount a volume into the container",
-                "default_value": "type=bind,src=/Users/username/Desktop,dst=/projects/Desktop",
-                "choices": []
-              }
-            ]
+      "runtime_arguments": [
+        {
+          "type": "template",
+          "description": "Mount a volume into the container",
+          "template": "--mount=type=bind,src={source_path},dst={target_path}",
+          "is_required": true,
+          "is_repeated": true,
+          "properties": {
+            "source_path": {
+              "description": "Source path on host",
+              "format": "filepath",
+              "is_required": true
+            },
+            "target_path": {
+              "description": "Path to mount in the container. It should be rooted in `/project` directory.",
+              "is_required": true,
+              "default": "/project",
+            }
           }
-        ],
-        "positional_arguments": [
-          {
-            "position": 0,
-            "name": "image",
-            "description": "Docker image name",
-            "default_value": "mcp/filesystem",
-            "is_required": true,
-            "is_editable": false,
-            "is_repeatable": false,
-            "choices": []
-          },
-          {
-            "position": 1,
-            "name": "root_path",
-            "description": "Root path for filesystem access",
-            "default_value": "/projects",
-            "is_required": true,
-            "is_editable": false,
-            "is_repeatable": false,
-            "choices": []
-          }
-        ],
-        "named_arguments": []
-      },
+        }
+      ],
+      "arguments": [
+        {
+          "type": "positional",
+          "value": "/project",
+        }
+      ]
       "environment_variables": [
         {
+          "type": "named",
           "name": "LOG_LEVEL",
           "description": "Logging level (debug, info, warn, error)",
-          "required": false,
-          "default_value": "info"
+          "default": "info"
         }
       ]
     }
@@ -220,25 +159,13 @@ API Response:
       "registry_name": "npm",
       "name": "@modelcontextprotocol/server-brave-search",
       "version": "1.0.2",
-      "command": {
-        "name": "npx",
-        "subcommands": [],
-        "positional_arguments": [],
-        "named_arguments": [
-          {
-            "short_flag": "-y",
-            "requires_value": false,
-            "is_required": false,
-            "description": "Skip prompts"
-          }
-        ]
-      },
       "environment_variables": [
         {
+          "type": "named",
           "name": "BRAVE_API_KEY",
           "description": "Brave Search API Key",
-          "required": true,
-          "default_value": ""
+          "is_required": true,
+          "is_secret": true
         }
       ]
     }
@@ -286,51 +213,33 @@ API Response:
       "registry_name": "docker",
       "name": "mcp/filesystem",
       "version": "1.0.2",
-      "command": {
-        "name": "docker",
-        "subcommands": [
-          {
-            "name": "run",
-            "description": "Run the Docker container",
-            "is_required": true,
-            "named_arguments": [
-              {
-                "short_flag": "-i",
-                "requires_value": false,
-                "is_required": true,
-                "description": "Run in interactive mode"
-              },
-              {
-                "long_flag": "--rm",
-                "requires_value": false,
-                "is_required": true,
-                "description": "Remove container when it exits"
-              },
-              {
-                "long_flag": "--mount",
-                "requires_value": true,
-                "is_required": true,
-                "is_repeatable": true,
-                "description": "Mount a volume into the container"
-              }
-            ]
+      "runtime_arguments": [
+        {
+          "type": "template",
+          "description": "Mount a volume into the container",
+          "template": "--mount=type=bind,src={source_path},dst={target_path}",
+          "is_required": true,
+          "is_repeated": true,
+          "properties": {
+            "source_path": {
+              "description": "Source path on host",
+              "format": "filepath",
+              "is_required": true
+            },
+            "target_path": {
+              "description": "Path to mount in the container. It should be rooted in `/project` directory.",
+              "is_required": true,
+              "default": "/project",
+            }
           }
-        ],
-        "positional_arguments": [
-          {
-            "position": 0,
-            "name": "image",
-            "description": "Docker image name",
-            "default_value": "mcp/filesystem"
-          },
-          {
-            "position": 1,
-            "name": "root_path",
-            "description": "Root path for filesystem access",
-            "default_value": "/projects"
-          }
-        ]
-      }
+        }
+      ],
+      "arguments": [
+        {
+          "type": "positional",
+          "value": "/project",
+        }
+      ]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


api/README.md
```diff
@@ -65,122 +65,61 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
...
-            "short_flag": "-y",
-            "requires_value": false,
-            "is_required": false,
-            "is_editable": false,
-            "description": "Skip prompts and automatically answer yes",
-            "choices": []
-          }
-        ]
-      },
+      "arguments": [
```
tadasant: Seeing as we have `runtime_arguments`, it may be worth naming this something more explicit, like `server_arguments`, `execution_arguments`, or `package_arguments`?

api/README.md
```diff
@@ -65,122 +65,61 @@ GET /v0/servers/a5e8a7f0-d4e4-4a1d-b12f-2896a23fd4f1?version=0.0.3
...
+          "type": "positional",
+          "description": "Path to access",
+          "default": "/Users/username/Desktop",
+          "is_required": true,
+          "is_repeated": true
+        }
+      ],
       "environment_variables": [
         {
+          "type": "named",
```
tadasant: Seems a little odd to me to require explicitly describing an env var as "named". Do we need the `type` enum here in `NamedInput`? If it's just a single value enum, isn't that implied?

tadasant (commented):
I like this hybrid approach a lot. `named`, `positional`, and `template` as "types" seem to be a good answer here. Keeps it quite minimalist and clean insofar as the config we're asking devs to write.

Just some small comments. @sridharavinash would be interesting to see if our seed data is able to conform to this without issues.

api/README.md
connor4312: I'm ambivalent, I feel like the naming of `arguments` vs `runtime_arguments` helps provide indication that that is the default field vs the 'advanced` runtime_arguments, which should be present only when specific runtime integration is necessary.

api/README.md
connor4312: The benefit is alowing templated input in addition to named inputs. But if we don't care about this, we can model it more simply like

```yaml
        headers:
          type: object
          additionalProperties:
            $ref: '#/components/schemas/UserInput'
```

tadasant: Ah got it, I just misread further down where you have `[named]` in the spec as the enum, but I see that's just a key that could be `template` instead to indicate a template. Allowing for templating in env vars makes sense.

api/README.md
tadasant: Personally I'd feel better to disambiguate. Would say `server_arguments` is most obvious to me.

Don't feel strongly though

connor4312: 👍  will let others weigh in when the work week begins 🙂

api/README.md
connor4312: Actually I think we do need to tweak this because template inputs don't inherently have a "key" which is needed for env variables and heaers

api/README.md
sandy081: I think disambiguation is good and gives clarity. I would go with `package_arguments` since we are talking about packages here.

api/openapi.yaml
```diff
@@ -162,155 +162,143 @@ components:
...
         description:
+          description: A description of the input, which can be used by clients to provide context to the user.
           type: string
-          example: "Path to access"
         is_required:
           type: boolean
-          example: true
-        is_repeatable:
+          default: false
+        is_repeated:
```
sandy081: Should we go with an array type value instead of `is_repeated` property?

api/openapi.yaml
```diff
@@ -162,155 +162,143 @@ components:
...
+          name:
+            type: string
+            example: mount_config
+            description: The name of the input, which can be used by clients as an identifier to reference the input. For environment variables and headers, this is the environment variable or header names. For arguments, this is solely for reference and not part of the command line.
+          description:
+            description: A description of the input, which can be used by clients to provide context to the user.
+            type: string
+          is_required:
+            type: boolean
+          is_repeated:
```
sandy081: Also same as before, should we support array value instead of `is_repeated` property

api/openapi.yaml
sandy081: Also not sure if this makes sense for positional_input

sandy081 (commented):
+1 for this approach.

> However, I think we need to be able to specify runtime args in some cases, otherwise the registry eventually mirrors all docker configs and options which is not a great situation to be in. So, I suggest a runtime_args in the same format. Say I wanted that for npm, I could:

I am wondering how runtime args work for `npm` type package? Since we are leaving the decision about which runtime executable to run to the client, should the runtime_args also specify to what exec these args are meant for? For example, how can a client now the given runtime_args are for `npx` or for someother exec?

connor4312:
> Since we are leaving the decision about which runtime executable to run to the client, should the runtime_args also specify to what exec these args are meant for? For example, how can a client now the given runtime_args are for npx or for someother exec?

I thought about this, perhaps with some `runtime_hint?: string` option. The alternative which would work for now is just for the package to, in its readme or description, describe the runtime it needs to operate. Whether a client is able to serve the runtime is very much a mileage-may-vary situation (for cases where there is that dependency there's no getting around it) and I felt unsure about putting in such an open-ended field into the spec.

For npm packages, I think I it would just be the runtime flags for Node.js itself. `npx` doesn't (currently) have any of its own relevant flags when running from a package, but it does pass through Node.js flags.

api/openapi.yaml
connor4312: I'm not clear on what you mean by this. This spec specifies how package arguments are defined, and is_repeated means that the client can choose to repeat this argument any number of times, such as docker mounts. That would likely result in an array being used to represent that in configs (like in my last example in the PR description.)

Having repeated positional arguments is a valid thing packages can do. `@modelcontextprotocol/server-filesystem` can take repeated positional arguments to define multiple folders to serve data from, e.g. https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem#npx

sandy081: > This spec specifies how package arguments are defined, and is_repeated means that the client can choose to repeat this argument any number of times,

Thats right. But is not it enough to represent this by passing an array value for the argument so that the client can generate the command in `mcp.json` correctly by repeating the arg for each value. I am not sure the client would like to validate the command in `mcp.json` with the server manifest always before running (say, if the user changes the command in mcp.json manually) and implement UI for adding dynamically multiple values for each argument that can be repeatable?

Also what if the server manifest provides array value for a named argument and this flag is not set? I would like to treat server manifest is a simple static transformer of itself into commands in mcp.json by the client. Once it is generated client/user can change it however it wants.

sandy081:
> I thought about this, perhaps with some `runtime_hint?: string` option. 

Yeah something in these lines would be good

api/openapi.yaml
connor4312: > I am not sure the client would like to validate the command in mcp.json with the server manifest always before running (say, if the user changes the command in mcp.json manually) and implement UI for adding dynamically multiple values for each argument that can be repeatable?

I think this is something that we should allow, and `is_repeated` let us do this.

> Also what if the server manifest provides array value for a named argument and this flag is not set?

For VS Code, we should generate a JSON schema for the configuration properties in mcp.json (like we do for debug configurations) and an array in a non-repeated field would be a validation error. But it's up to clients to treat this how they like.

>  I would like to treat server manifest is a simple static transformer of itself into commands in mcp.json by the client. Once it is generated client/user can change it however it wants.

This is a client implementation decision, but I don't think we can do this because then users have no way to change their previous configuration without redoing everything. They can't just tweak a single config value unless we have complex logic to restore the original inputs by parsing the CLI. This is bad UX for iteration, tweaking, and upgrading. Maybe we have an 'expand to args' option but we should not force this on users.

But regardless, I don't think we should drop this from the spec just because we're unsure if we'll use it for VS Code. Other editors have rich UI and will definitely want to build the "multi add" UI.

sandy081: My feedback is from a general client perspective and also defining the server manifest intention. If the server manifest has this support then all clients have to adhere to that. I thought server manifest is just a way of transforming its definition into commands and args. This seems to go a bit more than that. 

Probably we should at least remove array value type if the property is repeatable.

connor4312: > Probably we should at least remove array value type if the property is repeatable.

I'm not super clear on what you mean by this, what would it look like for the changes in this PR?

sandy081: I meant by following schema

```
format:
          type: string
          description: Input format. Includes a variety of standardized formats, including `filepath`, which should be interpreted as a file on the user's filesystem.
          enum: [string, number, boolean, array, filepath]
          default: string
```

I am assuming that the value can be an array?

sridharavinash:
> Just some small comments. @sridharavinash would be interesting to see if our seed data is able to conform to this without issues.


Shouldn't be too much of big lift. Once we're good here and I can finagle the seed data to make the new changes.

api/openapi.yaml
connor4312: Ah, right, yes I agree we should remove that

sandy081: @connor4312 Restarting this conversation because I have an another question about the same property.

- What if a named/templated argument has multiple values (repeatable) and how that can be represented in the server manifest in the registry? Just for an example lets take docker `--mount` argument and how can this be represented if there are more than one value for this arg?

- Since we are reusing the Named/Template Argument schema for environment variables and remote headers, does this `is_repeatable` property makes sense here? Should we extract this property and define it only for named arguments?

api/README.md
```diff
@@ -286,51 +213,33 @@ API Response:
...
-          {
-            "position": 1,
-            "name": "root_path",
-            "description": "Root path for filesystem access",
-            "default_value": "/projects"
+      "runtime_arguments": [
+        {
+          "type": "template",
+          "description": "Mount a volume into the container",
+          "template": "--mount=type=bind,src={source_path},dst={target_path}",
```
sandy081: Should this be changed to following

```json
{
   ...
   "name": "--mount",
   "template": "type=bind,src={source_path},dst={target_path}"
}
```

api/openapi.yaml
```diff
@@ -162,155 +162,143 @@ components:
...
+            type: boolean
+            description: Whether the input can be repeated multiple times in the command line.
+          template:
+            type: string
+            description: The template string that will be used to generate the value. Values wrapped in `{curly_braces}` will be replaced with the corresponding properties. Input which is not `is_required` is replaced with its `default` value, or otherwise an empty string.
+            example: "--mount=type=bind,src={host_path},dst={container_path}"
+          properties:
```
sandy081, line 301: `properties` seems to be too generic. Should this be called `variables`?

api/openapi.yaml
```diff
@@ -162,155 +162,143 @@ components:
...
-        command:
-          $ref: '#/components/schemas/Command'
-        environment_variables:
-          type: array
-          items:
-            $ref: '#/components/schemas/EnvironmentVariable'
-    
+          - type
+          - template
+          - name
```
sandy081: Does the positional argument can't have a template?

api/README.md
connor4312: `template`s aren't expanded into `named` --flags by default. This is so that things that aren't flags can be represented in templates, in fact you could implement `named` things using `template`s, `named` is just shorthand. However, in the argument list, clients do need a way to reference this, so `name` is present. I describe this in the description of `name` for `TemplateInput`:

> The name of the input, which can be used by clients as an identifier to reference the input. For environment variables and headers, this is the environment variable or header names. For arguments, this is solely for reference and not part of the command line.

api/openapi.yaml
connor4312: Positional arguments can be represented using templates (see earlier comment)

api/openapi.yaml
connor4312: > Just for an example lets take docker --mount argument and how can this be represented if there are more than one value for this arg?

I have an example of how that might look in the PR description:

```
  "filesystem": {
    "server": "@modelcontextprotocol/servers/src/filesystem@1.0.2",
    "package": "docker",
    "settings": {
      "mount_config": [
        { "source_path": "/Users/username/Desktop", "target_path": "/project/desktop" },
        { "source_path": "/path/to/other/allowed/dir", "target_path": "/project/other/allowed/dir" },
      ]
    }
  }
```

There is not repetition support within a `property`.

> Since we are reusing the Named/Template Argument schema for environment variables and remote headers, does this is_repeatable property makes sense here?

For headers, it can, HTTP defines canonical representation for repeated headers (concatenation with a `,`) so repetition is fine. For environment variables, there is no such way.

^ I will make a change to refine this. I will move is_repeatable out of the BaseInput so it's not a thing for template variables, and add a doc comment on template and named inputs to note they are not repeatable in environment variables

api/README.md
sandy081: > The name of the input, which can be used by clients as an identifier to reference the input. For environment variables and headers, this is the environment variable or header names. For arguments, this is solely for reference and not part of the command line.

IMO this is confusing to me. `name` is being used differently in different places.

alexhancock:
This looks great to me! I like the design of the three options.

Maybe I am missing something about the intended client logic, but what validations would we imagine for `template` strings? Both for validity but also command injection. For example, what would we do if someone submitted a server with:

```
"template": “;rm -rf ~/Development”
```

alexhancock (approved):

api/README.md
sandy081: @connor4312 How about removing template argument and embed the template/variables in positional/named arguments? This means a positional/named argument can provide variables by defining them. 

Please see this branch - https://github.com/modelcontextprotocol/registry/compare/connor4312/cmd-refinements...sandy081/homely-duck

connor4312: Sure, I think that makes sense. Will bring in your commits. I think we don't want to allow template nesting so I'll make a couple more tweaks.

connor4312: We still will want _a_ name or ID in positional arguments so they can be referenced and stable via config, when user input is needed for them. To avoid some confusion I renamed `name` to `name_hint` on positional args. CLI's will generally already have such a name_hint in their arg parser and available via `--help`, so I think that asking for a `name_hint` is not burdensome.

I did simplify a `KeyValueInput` for headers/env variables which drops the required `type` and `is_repeatable` (headers _can_ repeat, but having them in a single arg is also fine and I would prefer to keep types streamlined) and I inlined templating into `InputWithVariables`.

btw, in API types I tried to avoid combining `anyOf` inside `allOf` because a lot of language can't express this very well in their type system

connor4312:
> Maybe I am missing something about the intended client logic, but what validations would we imagine for template strings? Both for validity but also command injection. For example, what would we do if someone submitted a server with:

In general we have to assume that clients do the right shell escaping for arguments (or don't run the in shell, e.g. `child_process.spawn`). Users can do very valid things like put in quotation/punctuation marks and spaces into arguments, and if the client runs them in a shell it takes on the responsibility of escaping those.

api/README.md
sandy081: I liked the new changes. I will resolve this conversation and provide feedback on latest. Thanks

api/openapi.yaml
```diff
@@ -162,155 +162,141 @@ components:
...
           properties:
-            position:
-              type: integer
-              example: 1
-    
+            type:
+              type: string
+              enum: [positional]
+              example: "positional"
+            name_hint:
```
sandy081: How about calling this just `hint` or `value_hint`? Also please update the description of this property.

api/openapi.yaml
```diff
@@ -162,155 +162,141 @@ components:
...
           properties:
-            short_flag:
+            type:
               type: string
-              example: "-y"
-            long_flag:
+              enum: [named]
+              example: "named"
+            name:
               type: string
```
sandy081, line 287: Do you think should we add some description to this property?

sandy081 (approved):
All looks good except for few minor comments. 

Approving.

tadasant:
> In general we have to assume that clients do the right shell escaping for arguments (or don't run the in shell, e.g. `child_process.spawn`). Users can do very valid things like put in quotation/punctuation marks and spaces into arguments, and if the client runs them in a shell it takes on the responsibility of escaping those.

I think we should probably continue this discussion: https://github.com/modelcontextprotocol/registry/issues/41

connor4312:
🚀

api/README.md
tiovikram, line 1: This greatly simplifies the command structure from the previous draft of the API specification.  
  
There are a few matters I would like to discuss with the current iteration and get your thoughts on the validity of these matters.
   
   
  **1. Making the explicitly detailed parts of the command implicit:**
    
Example: For the command, `docker run --rm -i -e BRAVE_API_KEY mcp/brave-search`, this change makes it such that the parts of the command such as `run`, `--rm`, `-i`, `-e` (possibly) would all have to be implicitly known to be added to the constructed config by the client for the server to work as expected.
  
I am more in favour of the arguments being a single array wherein a consumer of this API can convert the arguments into the MCP config format through a single statement as follows.
```typescript
arguments.map((argument: Argument) => { /* some argument handling logic */})
```

The aforementioned approach would differ to looking up each separate type of argument in the configuration object as follows.
```typescript
arguments.named_arguments((argument: Argument) => { /* some named argument handling logic */ } +  arguments.runtime_arguments((argument: Argument) => { /* some named argument handling logic */ } +  arguments.package_arguments((argument: Argument) => { /* some named argument handling logic */ } + arguments.template_arguments((argument: Argument) => { /* some named argument handling logic */ }
```

Having sets of different arguments in different places also may complicate ordering (although the effects of this may only be minor, if the arguments are mostly constructed well by all package uploaders).
  
  **2. Deleting command causes tight-coupling of `"registry_name"` and "`command`":**
  
Deleting the `"command"` section with `"name": "npx"` or `"name": "docker"` then tightly couples the command to the source of the registry.
    
While this is not a problem if the registries are simply limited to [NPM](https://npmjs.com), [Docker Hub](https://hub.docker.com) or PyPI (https://pypi.org) which imply the runtime `npx`, `docker` and `uvx` as the command, a large benefit of this being open-source software is that enterprises may want to deploy their own versions internally for MCP use which may result in package registries being [JFrog](https://jfrog.com), [GitHub Container Registry (ghcr.io)](https://ghcr.io), the artifact registry of the chosen cloud provider (AWS, Azure, GCP) or any variant that obeys the same API as NPM, PyPI or Docker.
    
This would result in the `"registry_name"` field being changed to "JFrog", "GHCR" or whatever the respective registry is while the command would still remain `npx`, `docker`, or `uvx` and have to be handled by the client.
    
I propose de-coupling the run command from the registry name as a good pattern from the previous iteration of this API draft. Please let me know your thoughts on this.

tadasant: @tiovikram for visibility, I think you should open up an Issue for each of these topics, as this PR is closed and will be hard to get eyeballs on.

I'm not quite following the first suggestion, maybe some more explicit examples of what that means for the API shape would be helpful.

For the second one, it does seem worth a discussion whether we want an analog to `runtime_hint`

connor4312: > Example: For the command, docker run --rm -i -e BRAVE_API_KEY mcp/brave-search, this change makes it such that the parts of the command such as run, --rm, -i, -e (possibly) would all have to be implicitly known to be added to the constructed config by the client for the server to work as expected.

If arguments are needed to make the server work as expected, they should be in the `runtime_arguemtns`. Of those you included, that is only `-e BRAVE_API_KEY`. `--rm` is mostly up to the user/client if they want that behavior, `-i` will be something clients always will add for docker runs that communicate over stdio.

> I am more in favour of the arguments being a single array wherein a consumer of this API can convert the arguments into the MCP config format through a single statement as follows.

Please see previous discussion on package versioning here for context on why this was changed https://github.com/modelcontextprotocol/registry/pull/3#discussion_r2074192996

> Deleting the "command" section with "name": "npx" or "name": "docker" then tightly couples the command to the source of the registry.

For that reason in the past I have suggested a `shell` fallback type, however I didn't add that in this PR to keep things initially scoped

api/README.md
```diff
@@ -340,17 +251,14 @@ claude_desktop_config.json:
 ```json
 {
   "filesystem": {
```
tadasant, line 253: @connor4312 I just realized I overlooked this diff -- it's meant to be a claude_desktop_config.json file, so I don't think this change is accurate as it prescribes what Claude Desktop should be doing. I think we should have left this as-is? Or am I misreading something?

Issue #39, opened 2025-05-13
Add `readmeUrl` property in the repository


sandy081:
Add `readmeUrl` property in the repository to show the details of the mcp server while browsing in the client like below

cc @digitarald 

<img width="748" alt="Image" src="https://github.com/user-attachments/assets/2aac4927-c6f8-4744-aa7a-d2111ffb3bfa" />

tadasant:
Makes sense that we would want something like this. I'd define the requirement as "a reference to a URL that provides additional context about the MCP server".

I think we should make this more generic than "readmeUrl". For example, I would want Zapier to list [this URL](https://zapier.com/mcp) for their server entry.

It also makes me think: can we get rid of the `repository.url` field if we add this? What's the purpose of the repository URL (if we already have the `repository.id`)? Previously, I was thinking you might want to "click to read more" by clicking into the repository, but replacing that with a generic `descriptionUrl` (more generic form of `readmeUrl`) would keep things simple: only one URL to provide and manage for us. Maybe we just call it `url` at the server level. 

Introducing this does open us up to more spam/security problems (we are not otherwise storing or exposing any free-form URL's). But I think it's probably worth the trade-off.

sandy081:
Generally, a cautious user would like to go through the repository before installing an extension which shares similar experience with VS Code extensions. I would retain the repo url unless you do not think so,

tadasant:
Yeah ok fair enough, the notion of source code URL is meaningfully different than `Server.url`. I'd be ok with keeping `Server.repository.url` in place alongside adding a new optional `Server.url`.

digitarald:
I’d frame the requirement as: "a resource URL that clients can fetch a Markdown document from, which can be previewed within the client (or linked to if direct fetching isn’t possible)."

This approach allows users to view detailed information about the MCP server without leaving the editor. While linking to an external homepage is useful for deeper exploration, not having a dedicated README or similar document limits the amount of information the registry can display directly in the client.

I support keeping `repository.url`: Open source repositories add value for security and transparency, allowing users to review the source code and verify the provenance of releases.

As for naming, `readmeUrl` could be generalized to something like `learnMore` or `docs` if needed, but referencing a README isn’t too limiting in what users should put in.

tadasant:
How would we handle versioning alongside guidance to render the markdown like that?

[...]

PR #40, opened 2025-05-13
support readme_url

sandy081:
Ref #39

Add readme_url property to the server object. This is url to the README markdown file of the server. Returns readme content in markdown format.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -110,6 +110,7 @@ components:
         - description
         - repository
         - version_detail
+        - readme_url
       properties:
         id:
           type: string
@@ -123,6 +124,10 @@ components:
           example: "Node.js server implementing Model Context Protocol (MCP) for filesystem operations."
         repository:
           $ref: '#/components/schemas/Repository'
+        readme_url:
+          type: string
+          description: URL to the README markdown file of the server
+          example: "https://raw.githubusercontent.com/modelcontextprotocol/servers/refs/heads/main/src/filesystem/README.md"
         version_detail:
           type: object
           required:
--- a/api/README.md
+++ b/api/README.md
@@ -26,7 +26,8 @@ GET /v0/servers?limit=5000&offset=0
         "version": "1.0.2",
         "release_date": "2023-06-15T10:30:00Z",
         "is_latest": true
-      }
+      },
+      "readme_url": "https://raw.githubusercontent.com/modelcontextprotocol/servers/refs/heads/main/src/filesystem/README.md"
     }
   ],
   "next": "https://registry.modelcontextprotocol.io/servers?offset=50",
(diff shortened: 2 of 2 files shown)


sandy081:
cc @sridharavinash @tadasant

api/openapi.yaml
```diff
@@ -123,6 +124,10 @@ components:
           example: "Node.js server implementing Model Context Protocol (MCP) for filesystem operations."
         repository:
           $ref: '#/components/schemas/Repository'
+        readme_url:
```
tadasant, line 127: I think the notion of `README` is too specific to developers.

The registry should be an MCP server directory that is not specifically designed for developer persona end-users. I'd feel a lot more comfortable with this if the field was simply named `url`.
```suggestion
        url:
```

api/openapi.yaml
```diff
@@ -123,6 +124,10 @@ components:
...
+          type: string
+          description: URL to the README markdown file of the server
```
tadasant, line 129:
```suggestion
          description: URL to additional information about the server, such as a markdown README file or marketing landing page describing the server's functionalities.
```

api/openapi.yaml
connor4312: Even if the end user doesn't care about the readme, they might care about its contents and knowing it in a deterministic way allows:

1. Registry UI's to reliably display the server description, e.g. https://mcp.so/server/github/modelcontextprotocol
2. Tooling to ingest the documentation, e.g. VS Code and I assume Cline(?) pull in the readme to do AI-assisted server setup

The readme/documentation might differ from the homepage of the MCP server. An alternative is embedding the readme in the registry record itself, which npm and pypi do, but that is a bit bulkier for the server to deal with.

tiovikram:
I'm not too particular on `readme_url` vs `url` so will differ to @tadasant and @connor4312 on this matter

api/openapi.yaml
tadasant: I think it's worth continuing this discussion here: https://github.com/modelcontextprotocol/registry/issues/39#issuecomment-2877569440

I think rendering a README is a bad pattern to encourage -- we deliberately chose to not do so with PulseMCP because of the bad UX it creates:
- README's vary wildly in quality, and the poor ones can make your MCP marketplace look bad by association
- They vary wildly in structure, so it can be a disorienting experience flipping from one to another
- If you are interested in a specific MCP client, it may not be present in the README, or your competitors may be the only ones present
- If you actually want the README as an end-user, you may as well click the "View Source" button to get it straight from the source, instead of potentially looking at a stale version of it

The discussion on https://github.com/modelcontextprotocol/registry/issues/39#issuecomment-2877569440 is more about a generic (perhaps purpose-crafted for the official registry) markdown file, which IMO is a little better than shoehorning a technical README - but I still have versioning and perhaps spam-related concerns with adopting it. I think it's fair to hack it into an MVP implementation on the client side, but these are my hesitations with enshrining it as a long-term standard.

nareshrajkumar866 (approved):

tadasant:
Closing in favor of https://github.com/modelcontextprotocol/registry/issues/81

PR #45, opened 2025-05-16
added spectral linting file and removed jsonSchemaDialect b/c it is r…

markweitzel:
Added spectral rule file so we can validate the openapi doc.
In doing so, removed the jsonSchemaDialiect line b/c it was in the wrong place and redundant b/c it specified the default value.

### DIFF
--- a/api/.spectral.yaml
+++ b/api/.spectral.yaml
@@ -1,0 +1,1 @@
+extends: spectral:oas
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -82,7 +82,6 @@ paths:
                     example: "Server not found"
 components:
   schemas:
-    jsonSchemaDialect: "https://json-schema.org/draft/2020-12/schema"
     Repository:
       type: object
       required:


nareshrajkumar866 (approved):

domdomegg:
Thanks for the contribution! Can you provide a bit more context to help review these changes?

I haven't used spectral before. If it's to lint the OpenAPI config, could we maybe add a GitHub Action and/or docs to do this rather than just adding the config file? And are there other alternatives we've considered that might be a better approach here?

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to a git history rewrite mistake. 

Unfortunately, this PR cannot be automatically recreated because it was created from the 'main' branch of your fork. To restore your PR, please:

1. Create a new feature branch in your fork with your changes
2. Push the feature branch to your fork
3. Create a new PR from that feature branch

This approach will also make it easier to manage future updates to your PR. Again, I apologize for the inconvenience this has caused.

PR #47, opened 2025-05-17
feat: use new seed file in docker build

toby:
Updated the `Dockerfile` to use the new seed file `seed_2025_05_16.json` instead of the old `seed.json`.


sridharavinash (approved):
LGTM!!

PR #48, opened 2025-05-17
docs: update readme with new schema, some cleanup

toby:
I updated the `README.md` with the new schema from the API. Also changed the `docker compose` mount volume to `.db` so it doesn't share the same directory with the seed file. I also removed the old `seed.json` to avoid confusion.


README.md
```diff
@@ -146,27 +203,62 @@ Headers:
...
+                }
+            ],
+            "environment_variables": [
+                {
+                    "description": "API Key to access the server",
+                    "name": "API_KEY"
+                }
+            ]
+        },{
+            "registry_name": "docker>",
```
sridharavinash: Small typo here, should be `docker`

sridharavinash (dismissed):

sridharavinash (approved):

PR #61, opened 2025-05-22
fix: Add default seed.json to match config

chenmingyong0423:
When cloning the repository and building the project using:

```bash
go build ./cmd/registry
./registry
```

the application fails to start because the default seed file path (`data/seed.json`) does not exist in the repo:

```
Failed to read seed file: failed to read file: open data/seed.json: The system cannot find the file specified.
```

To address this, this PR adds a `data/seed.json` file to ensure the app can run successfully with default settings.

### Changes:

* Added `data/seed.json` as a default seed file to align with the default configuration
* Contents are copied from the latest available seed file (e.g. `seed_2025_05_16.json`)
* Ensures the application runs out-of-the-box without requiring additional environment configuration
* Developers can still override the seed file path using the `SEED_FILE_PATH` environment variable

Let me know if you'd prefer to solve this differently (e.g., by changing the default path or improving fallback behavior).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN go build -o /build/registry ./cmd/registry
 FROM alpine:latest
 WORKDIR /app
 COPY --from=builder /build/registry .
-COPY --from=builder /app/data/seed_2025_05_16.json /app/data/seed.json
+COPY --from=builder /app/data/seed.json /app/data/seed.json
 COPY --from=builder /app/internal/docs/swagger.yaml /app/internal/docs/swagger.yaml
 EXPOSE 8080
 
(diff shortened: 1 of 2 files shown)


nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

chenmingyong0423:
Hi! Just wondering — would it perhaps be more straightforward to update the envDefault tag in the Config.SeedFilePath field to point directly to data/seed_2025_05_16.json instead?

Adding a seed.json file works, but it might not be the most future-proof solution, especially if the seed file name keeps changing over time. Curious to hear your thoughts — happy to adjust the PR if needed!

Issue #86, opened 2025-05-27
Rename mcp.json to server.json


tadasant:
Pretty simple tweak, but one we should make to avoid overloading the `mcp.json` term.

sridharavinash:
closed by #106

PR #106, opened 2025-05-28
Rename mcp.json to server.json

chenmingyong0423:
Rename mcp.json to server.json
## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Related to #86


punkpeye:
`server.json` feels extremely generic. I would advocate for `mcp-server.json`

tadasant:
> `server.json` feels extremely generic. I would advocate for `mcp-server.json`

I don't feel strongly on this, but I think there is precedent in either direction. e.g. npm's `package.json`, python's `requirements.txt`. Because this is a file that only lives within an MCP server's codebase, I don't think there is significant risk that "server" will mean more than one thing within the context of the MCP spec.

I'm very open to being convinced otherwise, but we had a handful of folks unanimously agree on `server.json` when I first brought it up in the Steering Discord, so I would suggest we get this merged, and if there is compelling reason to revisit we should start an Issue and tag all the folks that initially supported server.json with an argument for why we should reconsider.

sridharavinash:
I don't have any strong opinions as well. The only case that might benefit from explicitness of the name may be in `.well-known` cases, i.e if we go down that path then having it be `.well-known/mcp/server.json` makes it more clearer than having it at a top level `.well-known/server.json` - but one could argue that the top level could be a mcp related domain so in that case having it just be `server.json` wouldn't be too bad.

PR #130, opened 2025-06-05
expose Server URL

maorfr:
## Motivation and Context
the `URL` field is documented but does not exist.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/model/model.go
+++ b/internal/model/model.go
@@ -114,6 +114,7 @@ type Server struct {
 	ID            string        `json:"id" bson:"id"`
 	Name          string        `json:"name" bson:"name"`
 	Description   string        `json:"description" bson:"description"`
+	URL           string        `json:"url" bson:"url"`
 	Repository    Repository    `json:"repository" bson:"repository"`
 	VersionDetail VersionDetail `json:"version_detail" bson:"version_detail"`
 }
--- a/internal/service/fake_service.go
+++ b/internal/service/fake_service.go
@@ -65,6 +65,22 @@ func NewFakeRegistryService() RegistryService {
 				IsLatest:    false,
 			},
 		},
+		{
+			ID:          uuid.New().String(),
+			Name:        "bluewhite/mcp-server",
+			Description: "Another dummy MCP registry for testing with a URL",
+			URL:         "https://example.com/mcp",
+			Repository: model.Repository{
+				URL:    "https://github.com/example/mcp-4",
+				Source: "github",
+				ID:     "example/mcp-4",
+			},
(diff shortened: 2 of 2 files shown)


tadasant:
@maorfr there is no [Server.url](https://github.com/modelcontextprotocol/registry/blob/6b22cf09b376ed94c772c69951a2d125e495b26b/docs/openapi.yaml#L105-L145) field - were you looking at something else?

I think you are looking for Repository.url.

Or maybe you would advocate for addition of some sort of `learn_more_url` or `marketing_url` on `Server`. Which is a reasonable discussion to have, however that is not in the documentation yet, so we'd need to start there. Do you have a use case?

maorfr:
examples of existing `url`:
- https://github.com/modelcontextprotocol/registry/blob/6b22cf09b376ed94c772c69951a2d125e495b26b/internal/docs/swagger.yaml#L222
- https://github.com/modelcontextprotocol/registry/blob/6b22cf09b376ed94c772c69951a2d125e495b26b/README.md#L120

tadasant:
Thanks for flagging - the spec and examples in [docs](https://github.com/modelcontextprotocol/registry/tree/6b22cf09b376ed94c772c69951a2d125e495b26b/docs) are the source of truth, so we should remove those errant examples unless we agree to add the field to the source of truth.

maorfr:
i think a server url makes a lot of sense, to answer the question -
where can you find a running instance of this MCP server? (think of a registry running within an org)

tadasant:
> where can you find a running instance of this MCP server?

This would be the Remote.url field.

The use case that makes sense to me would be a "landing page URL" like https://zapier.com/mcp.

We should discuss what the right naming would be for this, as `url` is clearly too ambiguous (that is evident even from just this conversation). Some options:
- `landing_page_url`
- `marketing_url`
- `learn_more_url`
- `more_information`
- `canonical_url`

And maybe we would require this URL to match the reverse DNS verification in `name`.

jonathanhefner:
> We should discuss what the right naming would be for this, as `url` is clearly too ambiguous (that is evident even from just this conversation). Some options:

Another option is `homepage`.  That's what NPM ([example](https://www.npmjs.com/package/@modelcontextprotocol/sdk)) and RubyGems ([example](https://rubygems.org/gems/mcp)) use, at least when displayed.

> And maybe we would require this URL to match the reverse DNS verification in `name`.

This might be too onerous a requirement.  I can imagine someone wanting to switch to a new domain after the fact.  (Gotta grab that new `.ai` TLD! :fire:)

Issue #131, opened 2025-06-09
Write a better FAQ, living design doc


tadasant:
As @connor4312 has pointed out, we are starting to get a lot of repeat questions that we have answers to scattered throughout docs and issues. The designs in the [initial announcement](https://github.com/modelcontextprotocol/registry/discussions/11) are starting to grow stale, so we should have a better running document somewhere in version control.

This document should include:
- Overarching design goals / design principles to follow
- In scope / out of scope 
- FAQ

connor4312:
I think the overall scope and design goals can be better represented by you and others, but I'm happy to help contribute info about the package structure/openapi spec once there's a draft going

toby:
@tadasant got this done in #134

Issue #103, opened 2025-05-27
Consider mitigating against typosquatting


tadasant:
Seeing as we aren't serving the use case of direct downloading of packages (like some `npm install <packagename>` flow), typosquatting is not a particularly notable risk.

However, we should consider blocking publication of packages within a certain edit distance of existing packages (esp. popular ones).

We may want to use #95 in the definition of "popular".

joelverhagen:
Typo squatting / homograph spoofing / Unicode normalization is a somewhat deep subject. My guess is that the deeper intent is to mitigate both search (e.g. accent insensitive) and browsing (visual similarity) attacks.

Are there rules/recommendations on how punycode MCP server names (reverse DNS names) are rendered? This is a vector for attack if tools handle the punycode.

connor4312:
I would say we should restrict to some reasonable subset of printable ASCII characters for server names, and that the registry should use the punycoded version of hostnames to avoid domain confusion.

rdimitrov:
Closing this out. Most of the concrete risks that came up in-thread are already mitigated structurally:

- **Unicode homograph / punycode confusion**: server names are restricted to ASCII (`^[a-zA-Z0-9.-]+/[a-zA-Z0-9._-]+$`), so this class of attack isn't reachable.
- **Squatting within a namespace**: namespace ownership is verified via GitHub/DNS/HTTP, so you can't register e.g. `io.github.<someone-else>/mcp`.
- **Edit-distance squatting across namespaces**: technically still possible, but as the original description noted, typosquatting isn't a notable risk given the registry isn't the download path.

Easy to reopen if a concrete attack pattern surfaces, but the proposal's own premise (low risk) combined with what's already in place leaves this without much to do.

Issue #100, opened 2025-05-27
Daily re-verification of DNS records


tadasant:
Blocked on https://github.com/modelcontextprotocol/registry/issues/85

E.g. domain expires, someone else buys it; we'd need to make sure the old owners don't still have access.

We'd should not invalidate historical packages that exist, but we should display a warning to anyone who requests information about them.

domdomegg:
May be superceded by #279

domdomegg:
Superceded by #279

Issue #95, opened 2025-05-27
Track and store "download counts" from official registry


tadasant:
Blocked on: https://github.com/modelcontextprotocol/registry/issues/94

Although a crude metric, it may be helpful to track "download counts" of every MCP server in the official registry. Only the official registry has insight into this data, and it can be used as a sorting signal by consumers.

If we do track this, it's worth being mindful of how this may negatively impact CDN caching - and worth designing around that.

rdimitrov:
Hey @tadasant, this one was explicitly blocked on #94, which we just closed as stale. Closing this for the same reason; easy to reopen if download-count tracking resurfaces as a concrete need.

PR #134, opened 2025-06-12
Add docs: design principles, FAQ, roadmap, architecture

tadasant:
Adding more up-to-date documentation to address #131:

> We are starting to get a lot of repeat questions that we have answers to scattered throughout docs and issues. The designs in the https://github.com/modelcontextprotocol/registry/discussions/11 are starting to grow stale, so we should have a better running document somewhere in version control.

This PR adds:

- `design_principles.md` - Core constraints and principles guiding the registry design
- `faq.md` - Frequently asked questions about the MCP Registry
- `roadmap.md` - High-level roadmap for the MCP Registry development
- `architecture.md` - Technical architecture, deployment strategies, and data flows

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -35,11 +35,6 @@ The initial version of the MCP Registry is actively being developed. The initial
 - **Server tagging**: The registry will not support server tagging
 - **Server rankings**: The registry will not rank servers by subjective measures of quality
 
-
-## Non-Goals / Out of Scope
-
-These items are explicitly not planned:
-
 ## Contributing
 
 See [CONTRIBUTING.md](../CONTRIBUTING.md) for details on how to contribute to the MCP Registry project.


docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+
+### What namespaces are available?
+
+- **With DNS verification**: `com.yourcompany/server-name` (reverse DNS notation)
+- **Without DNS verification**: `io.github.yourusername/server-name`
+- DNS verification is done via TXT records and enables authoritative namespacing
+
+### Is open source required?
+
+Locally-run servers are required to be open source. Remote servers are not.
```
connor4312: Is this actually the case? I could easily publish an npm package containing a closed-source binary.

docs/roadmap.md
```diff
@@ -0,0 +1,45 @@
...
+- [ ] CLI/REST API: DNS verification system for namespacing
+
+## Backlog (Future Work)
+
+- [ ] UI implementation
+- [ ] Store and surface other data besides servers (e.g. [clients](https://modelcontextprotocol.io/clients), resources)
+- [ ] Additional IdP support (beyond GitHub)
+- [ ] Download count tracking
+
+## Out of Scope (Never Planned)
```
connor4312:
```suggestion
## Out of Scope (Not Planned)
```

Less absolute as a place to park other ideas

docs/faq.md
tadasant: Good point, you're right, we don't have a reasonable means to enforce that. Will rephrase such that open source is encouraged but not required.

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+
+### 3. DNS Verification Flow
+
+```mermaid
+sequenceDiagram
+    participant User as User
+    participant CLI as CLI Tool
+    participant API as Registry API
+    participant DNS as DNS Provider
+    participant DB as MongoDB
```
jonathanhefner:
```suggestion
    participant DB as Database
```

docs/architecture.md
```diff
@@ -0,0 +1,207 @@
...
+## Data Flow Patterns
+
+### 1. Server Publication Flow
+
+```mermaid
+sequenceDiagram
+    participant Dev as Developer
+    participant CLI as CLI Tool
+    participant API as Registry API
+    participant DB as MongoDB
```
jonathanhefner:
```suggestion
    participant DB as Database
```

docs/design_principles.md
```diff
@@ -0,0 +1,39 @@
+# MCP Registry Design Principles
+
+These are the core constraints that guide the design of the MCP Registry. They are not exhaustive, but they are the most important principles that we will use to evaluate design decisions.
+
+## 1. Single Source of Truth
+
+The registry serves as the authoritative metadata repository for publicly-available MCP servers, both locally-run (open source) and remote (open or closed source). Server creators publish once, and all consumers (MCP clients, aggregators, etc.) reference the same canonical data.
```
jonathanhefner: Pursuant to https://github.com/modelcontextprotocol/registry/pull/134#discussion_r2145015567, you might want to omit "open source" here as well.

docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+
+Yes, versioning is supported:
+- Each version gets its own immutable metadata
+- Version bumps are required for updates
+- Old versions remain accessible for compatibility
+- The registry tracks which version is "latest"
+
+### How do I update my server metadata?
+
+Submit a new `server.json` with an incremented version number. Once published, version metadata is immutable (similar to npm).
```
jonathanhefner, line 72: Is there any place that we specify the version number format?  For example, would we support "1.0.0-alpha" => "1.0.0-beta"?

docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+### How do I know a server is from the claimed organization?
+
+DNS verification ensures namespace ownership. For example:
+- `com.microsoft/server` requires DNS verification of microsoft.com
+- `io.github.username/server` is tied to a GitHub account or GitHub organization
+
+### What about typosquatting?
+
+The registry implements:
+- Automatic blocking of names within a certain edit distance of existing servers
```
jonathanhefner, line 91: Is this planned for the initial release?  (#103 is labeled as "not go-live blocker".)

docs/roadmap.md
```diff
@@ -0,0 +1,45 @@
...
+- **Server hosting**: The registry does not provide hosting for servers
+- **Search engine**: The registry will not provide a commercial grade search engine for servers
+- **Server tagging**: The registry will not support server tagging
+- **Server rankings**: The registry will not rank servers by subjective measures of quality
+
+
+## Non-Goals / Out of Scope
+
+These items are explicitly not planned:
+
```
jonathanhefner: Is this section an editing artifact, or is it WIP?

docs/roadmap.md
```diff
@@ -0,0 +1,45 @@
...
+## Contributing
+
+See [../CONTRIBUTING.md](./CONTRIBUTING.md) for details on how to contribute to the MCP Registry project.
```
jonathanhefner: Not sure about the path to `CONTRIBUTING.md` in the final rendered version of this document, but perhaps:

```suggestion
See [CONTRIBUTING.md](../CONTRIBUTING.md) for details on how to contribute to the MCP Registry project.
```

docs/faq.md
joelverhagen: I had assumed SemVer,  but this is a nuanced problem. I imagine folks will want to use the same version string as the underlying package for simplicity. And there is no guarantee of a unified version model for N package ecosystems. For example NuGet is SemVer-ish with case insensitive and a 4th version digit supported.

Perhaps complaint SemVer 2.0.0 for MCP Registry, and build metadata rejected? (1.0.0-alpha but not 1.0.0-alpha+foo which SemVer 2.0.0 supports but adds complexity).

docs/design_principles.md
```diff
@@ -0,0 +1,39 @@
...
+
+## 2. Minimal Operational Burden
+
+- Design for low maintenance and operational overhead
+- Delegate complexity to existing services where possible (GitHub for auth, npm/PyPI for packages)
+- Avoid features that require constant human intervention or moderation
+- Build for reasonable downtime tolerance (24h acceptable) by having consumers cache data for their end-users
```
toby, line 14: I think it's good to encourage caching of data but I'm hoping we can have a better SLA than 24 hours of downtime since publishers won't be able to publish new servers during that time.

docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+### Is there security scanning?
+
+The MVP delegates security to the underlying package registries. Future iterations may include:
+- Vulnerability scanning
+- Dependency analysis
+
+### How is spam prevented?
+
+- GitHub authentication requirement
+- Rate limiting (e.g., one new server per user per day)
```
toby: We'll _probably_ want a bit higher of a limit. I could imagine a company publishing at least a few servers in day.

docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+- Character limits and regex validation on free-form fields
+- Potential AI-based spam detection
+- Community reporting and admin blacklisting capabilities
+
+## API & Integration
+
+### How often should I poll the registry?
+
+Recommended polling frequency:
+- `/servers` endpoint: once per day
```
toby: We can probably go a bit higher too. Since there won't be millions of registries, I'd say maybe once an hour? We can cache on our end as well.

docs/faq.md
```diff
@@ -0,0 +1,169 @@
...
+While the API shapes and data formats are designed for reuse, the registry implementation itself is not designed for self-hosting. Organizations needing private registries should:
+- Implement the same API shape
+- Use the same `server.json` format
+- Potentially mirror/filter the official registry data
+
+## Operations & Maintenance
+
+### What's the expected reliability?
+
+- 24-hour+ downtime tolerance is acceptable, so you shouldn't rely on the registry being always available.
```
toby, line 135: Same as above re: trying to improve this for the benefit of publishers.

tadasant:
Thanks for the feedback everyone! I'm about to head out for a week of OOO, so will be circling back to address when I'm back.

docs/design_principles.md
```diff
@@ -0,0 +1,39 @@
...
+
+## 3. Vendor Neutrality
```
patrick-rodgers, line 16: No interest in blocking anyone from building and publishing a server - but say we released an in-service MCP server for OneDrive, how would that be differentiated from the 100 other implementation that have OneDrive in the name/description/access OD files? Substitute OneDrive for GDrive or any other service. This is the app store problem of a bunch of very similar things being very confusing for users.

Say I am a consumer, I know nothing of tech, and I search for "OneDrive" in MCP registry and get 100 results - how would I make a choice?

jonathanhefner: I think that amounts to a curation problem, which would be the purview of aggregators rather than the registry itself.  We do plan to track download counts (https://github.com/modelcontextprotocol/registry/issues/95), so perhaps aggregators could sort by those numbers.  Aggregators can also implement their own ratings systems to recommend the most popular (and presumably official) servers.  And I suppose paid placement is another option. :skull:

joelverhagen: (I am not on the registry team, just focused on it because I'm lighting up NuGet based MCP servers)
Agreed that this mostly goes to the clients that read the data and "project" the data into whatever UX they want.
Additionally, the server name is a reverse DNS name so the official OneDrive MCP server could perhaps have a name like `com.microsoft.onedrive/mcp` or something like that. This would mean that the owner of microsoft.com is publishing this OneDrive MCP server. I am not sure when non-GitHub package names will come in. Maybe for the first release per https://github.com/modelcontextprotocol/registry/issues/100.

patrick-rodgers: Then from the registry standpoint the plan is to externalize the risk.

It would be worth thinking through the implications of that approach, where without some indicator of official releases the consumer (think non-tech person - what's DNS? what's a github package?) is potentially placing their content at risk.

Anyone should be able to build/publish servers, while also having a way to mark official releases would be nice.

jonathanhefner: > Anyone should be able to build/publish servers, while also having a way to mark official releases would be nice.

How would the registry determine which packages are official and which are not?  How would disputes be handled?

patrick-rodgers: npm has the organization model, which isn't flawless but groups things in a user understandable way.

docs/design_principles.md
tadasant: My biggest concern about sub-24h is that anything under 24h necessitates on on-call schedule. If we frame it as 24h, then we can rely on volunteers coming online once a day.

I know it's not great for productivity if publication is offline for 24h, but maybe we start with a 24h SLA and once we get a sense for adoption + other maintenance needs, we could then seek to create a permanent operational position that is funded/sponsored by someone?

docs/design_principles.md
tadasant: > Say I am a consumer, I know nothing of tech, and I search for "OneDrive" in MCP registry and get 100 results - how would I make a choice? ... (think non-tech person - what's DNS? what's a github package?) is potentially placing their content at risk.

I don't expect consumers to use the Registry API. They should only interface with MCP clients and their Registry mirrors, which they are responsible for augmenting and curating to the level of sophistication that matches their target persona. I don't think we can solve trust at the right level of granularity for all MCP uses through to the end-consumers, so the only scaleable way to approach this is to federate out to the clients, who own the end-user UI's and can present them as they see fit (after augmenting them with their own notions of trust and curation/filtering).

docs/faq.md
tadasant: Added a note about this here, we can figure it out separately: https://github.com/modelcontextprotocol/registry/issues/90#issuecomment-3006158877

I think it is actually unlikely that folks will want to bind the MCP server version with their package version. There may be cases where the MCP server version gets incremented even though the code doesn't change (for example, in the case of me deciding to update the `description`. It would be limiting myself to require that I make a code change to make that MCP server-level metadata change.

docs/faq.md
tadasant: Fair, I'll reword to say it's planned but perhaps not implemented at the time readers are reading

docs/faq.md
tadasant: That's fair, I think once per hour would be fine

docs/faq.md
tadasant: Fair, if we make it 10/day that probably does enough to still discourage mass spam and likely won't severely impact a company trying to launch a set of servers.

docs/faq.md
tadasant: To discuss in https://github.com/modelcontextprotocol/registry/pull/134/files#r2146246641

docs/roadmap.md
tadasant: Oops, just a typo, thanks.

docs/faq.md
tadasant: Noted here https://github.com/modelcontextprotocol/registry/issues/21 too

tadasant:
Sorry for the long delay while I was OOO, back on now.

I took a pass at addressing comments. The only notable sticking point is the SLA @toby. Happy to continue discussing if we're still not aligned, maybe we land this PR and carve that out to discuss separately?

jonathanhefner (approved):

docs/faq.md
toby: I like that and the "escape hatch", it should protect against spam but still give us a way to support enterprises that want to go all-in on MCP.

toby (approved):

toby:
Yeah @tadasant let's do it. I can open an issue for further SLA discussion, but let's merge this for now.

docs/faq.md
domdomegg: Discussion on this continued in https://github.com/modelcontextprotocol/registry/issues/158

Issue #136, opened 2025-06-13
Don't allow `unknown` in `registry_name`


tadasant:
@dmartinol [pointed out](https://github.com/modelcontextprotocol/registry/issues/123#issuecomment-2966939920) that we have an `unknown` value in the `registry_name` seed data [here](https://github.com/modelcontextprotocol/registry/blob/686fbe1e78117546c91dfb47051e405f0bd1762b/data/seed_2025_05_16.json#L6180C9-L6180C35).

`registry_name` should be specifically one of the supported registries; unknown should not be permitted.

PR #138, opened 2025-06-13
Resolve packages and remove unknown registries

connor4312:
Closes https://github.com/modelcontextprotocol/registry/issues/136

- For each unknown registry, try to resolve the name on npm, pypi, and docker, and use that when available
- Filter out unknowns emove any servers that lack packages afterwards

### DIFF
--- a/data/seed_2025_05_16.json
+++ b/data/seed_2025_05_16.json
@@ -96,7 +96,7 @@
     },
     "packages": [
       {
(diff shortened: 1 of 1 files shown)


tadasant:
Thanks for doing this! Would like for @sridharavinash to take a look as I would imagine there is some script somewhere that is actually generating these `unknown` values that needs to be modified too?

Issue #135, opened 2025-06-13
Search for MCP servers by keywords


pavinduLakshan:
**Is your feature request related to a problem? Please describe.**

Current registry API doesn't seem to have a search API endpoint.

**Describe the solution you'd like**

It should be possible to search MCP servers by keywords.

**Describe alternatives you've considered**

N/A

**Additional context**

n/a

tadasant:
Good point, basic keyword string matching search was meant to be in scope for the MVP here; I forgot to cut an Issue for this and we should do this.

I do think we should design the search endpoint such that it is easily extensible for other implementations to do more advanced things with it (e.g. filter, sort); and perhaps look to other registry specifications for inspiration (e.g. WoT Discovery, xRegistry) on how to design this well. Even though advanced search is out of scope for the official registry itself.

jonathanhefner:
Will the central registry actually implement search, or are we just defining an endpoint for API implementors?  My instinct was that, for scalability reasons, search would be the purview of registry consumers.

tadasant:
I think we should actually implement a search by name (via exact substring match). Reason being, I would want our CLI tool to be able to provide functionality like "give me a list of servers already published by com.microsoft so I don't try to republish the same one".

domdomegg:
I would propose that we:
- Agree on 'implement a substring match on the 'name' field' approach
- For now, keep it out of the registry API spec, so that...
  - subregistries don't need to implement it
  - easier to make breaking changes/change how search works in future to support more capable forms of search
- Change the label from 'product requirements work' to 'implementation work' (or at least to 'tech requirements work', if we think there is much more to discuss on this)

domdomegg:
Related: https://github.com/modelcontextprotocol/registry/issues/291 (search for MCP servers by updated at timestamp)

domdomegg:
Recording decision: in today's maintainer meeting we agreed on [this](https://github.com/modelcontextprotocol/registry/issues/135#issuecomment-3208729181) for go live, combined with #291.

PR #142, opened 2025-06-15
feat: add keyword search functionality to registry service

tuannvm:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ The MCP Registry service provides a centralized repository for MCP server entrie
 ## Features
 
 - RESTful API for managing MCP registry entries (list, get, create, update, delete)
+- **Keyword search** across server names and descriptions with case-insensitive matching
 - Health check endpoint for service monitoring
 - Support for various environment configurations
 - Graceful shutdown handling
@@ -104,11 +105,12 @@ Returns the health status of the service:
 GET /v0/servers
 ```
 
-Lists MCP registry server entries with pagination support.
+Lists MCP registry server entries with pagination support. Supports keyword search across server names and descriptions.
 
 Query parameters:
 - `limit`: Maximum number of entries to return (default: 30, max: 100)
 - `cursor`: Pagination cursor for retrieving next set of results
+- `search`: Search keyword to filter servers by name and description (case-insensitive)
 
 Response example:
 ```json
@@ -128,6 +130,23 @@ Response example:
     "count": 30
   }
 }
+```
+
+##### Search Examples
+
+Search for servers containing "redis":
+```
+GET /v0/servers?search=redis
+```
+
+Search with pagination:
+```
+GET /v0/servers?search=database&limit=10&cursor=123e4567-e89b-12d3-a456-426614174000
+```
+
+Case-insensitive search for "API":
+```
+GET /v0/servers?search=API
 ```
 
 #### Get Server Details
--- a/internal/api/handlers/v0/publish_test.go
+++ b/internal/api/handlers/v0/publish_test.go
@@ -25,6 +25,11 @@ func (m *MockRegistryService) List(cursor string, limit int) ([]model.Server, st
 	return args.Get(0).([]model.Server), args.String(1), args.Error(2)
 }
 
+func (m *MockRegistryService) Search(query string, cursor string, limit int) ([]model.Server, string, error) {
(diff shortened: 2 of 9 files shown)


tuannvm:
https://github.com/modelcontextprotocol/registry/issues/135

tadasant:
Thanks for working on this, but I think we need to step back and do some back and forth on designing this feature, starting with OpenAPI specs before we try to implement.

First, we should do this https://github.com/modelcontextprotocol/registry/issues/149

Then, we should define the OpenAPI specs for both in a way that is easily extendible. It'd be worth looking at some comparable schemas like xRegistry and/or WoT Discovery for inspiration / alternatives considered.

After that lands, we can do the implementation.

domdomegg:
@tadasant Now that #149 is complete, should we revisit this and what the API should be here? I'm pretty happy with this interface (maybe moving it under an /extensions/ path or something, to indicate it's not part of the very core registry spec) and would lean towards accepting this PR and iterating on it.

FWIW I do think a decent search (beyond substring search) would be useful here. But I appreciate this can end up being costly/adding complexity to what we might want to be very lightweight. It really does depend on how most users end up interacting with the registries etc.

tadasant:
> @tadasant Now that #149 is complete, should we revisit this and what the API should be here? I'm pretty happy with this interface (maybe moving it under an /extensions/ path or something, to indicate it's not part of the very core registry spec) and would lean towards accepting this PR and iterating on it.
> 
> FWIW I do think a decent search (beyond substring search) would be useful here. But I appreciate this can end up being costly/adding complexity to what we might want to be very lightweight. It really does depend on how most users end up interacting with the registries etc.

My thinking with the [generic registry spec](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-registry-api/openapi.yaml) is that we'd want a way to define search that can be constrained to just do simple substring search, but also extended to more complex opinionated search as well. The Registry would constrain to simple substring, versus other implementations (usually downstream) would likely allow for more "advanced" search.

Three reasons I am keen to keep Registry search "simple":
1. Discouraging spam. The more search vectors we add, the more ways to "game" the system. For example, if we enable vector search, everyone will be incentivized keyword-stuff their names and descriptions -- and that's a bad experience for everyone consuming it.
2. Limiting our complexity. I think server search is an extremely hard thing to do well (I have not seen any third party MCP registry out there do it well to-date, and there have been very many attempts), so I think for the sake of keeping our scope limited and moving quickly, we'd do well to just do the easy thing that keeps us moving.
3. Discouraging direct consumption of the Registry (versus using it as an upstream). If search is intentionally limited, folks will be less likely to abuse the registry and consume it directly. We don't want them doing that because (1) we want to avoid scale perf problems with lots of recurring hits (millions of hourly users vs. just thousands of mirrors doing daily ETL) (2) we can't support a great SLA anyway with an OSS team (3) encouraging direct usage introduces additional expectations on the registry to make it fit for that use -- I think we are better off limiting our scope as much as possible and only focusing on what the upstream registry can uniquely provide to the community.

Regarding the interface, I haven't thought super deeply on this but my initial reaction to the PR is that `search={query}`, while suitable for the upstream Registry needs, is not flexible enough for downstream implementations.

For example, [WoT Discovery](https://www.w3.org/TR/wot-discovery/#directory-api-spec) pushes something like this:

```
# Find all TDs with a specific property name
GET /search/jsonpath?query=$..properties.temperature
```

JSONPath is probably the other extreme end of the spectrum, where it can "search anything" but is a burden to implement. But maybe there's a case to be made that it's worth it and we bake the complexity into an SDK (and per above, maybe we'd lock the upstream Registry down to just `name` searches by only allowing `GET /search/jsonpath?query=$[?(@.name =~ /.*myquery.*/i)]`.

My gut is there's some happy medium like a

```
GET /v0/search?name=query
```

Where  we can allowlist keys in the query parameters (constrained in the upstream Registry but free for all in the generic OpenAPI).

If we just want to push something out for v0 I'd be keen to go with that last suggestion^

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to a git history rewrite mistake. The PR has been recreated as #244 with all the original content preserved.

Please continue any discussions or reviews on the new PR: #244

Again, I apologize for the inconvenience this has caused.

Issue #143, opened 2025-06-17
Support streamable HTTP via `packages`


connor4312:
It was brought up that packages implicitly are stdio servers, but people might want to publish servers that are only streamable HTTP, e.g. to use authentication mechanisms.

If we do this, I would suggest having having an optional `uriTemplate` on packages. This is defined as an RFC 6570 template (same format as resource templates in MCP itself) with variables that can reference arguments and environment variables used in that package. This makes it easy to variables like `host` and `port` that can be reused to form the template.

However, a common class of vulnerability is escalation of privilege attacks for services listening on the same device or network, and there is no precedent yet in MCP servers to guard against this (e.g. by having a secret token set in the server's environment variable, required as a header or param on the route, and sent by clients.) So it would be up to servers to guard against this individually, or it would be up to clients to isolate the network of such servers, such as by running them in docker. This makes a safe implementation more complex, in exchange for HTTP-only features (currently just auth.)

tadasant:
Accidentally made a dupe: https://github.com/modelcontextprotocol/registry/issues/148

I'm inclined to say we should just add a `transport` field within `packages`. If we're uncomfortable with the security hole in the official registry API, we could lock down that OpenAPI schema specifically to only allow `stdio` for now.

connor4312:
I'll get a PR going in the morning

BobDickinson:
I have a similar concern (not sure it's exactly the same, but given that #148 was closed as a dupe of this, it seems like we're grouping these concerns together).

The current server payload tells me all about the code, including where to get it and how to run it.  It tells me as a client what configuration information I need to collect from the user.  But in the case of a non-stdio transport, it doesn't tell me how to actually connect to it and use it.

This concern presupposes the case in this issue - that there will exist discoverable, installable MCP servers that present other transports besides stdio (maybe in addition to stdio).  So if a server presents streamable, as a client I need to understand that (that I can use the streamable client transport) and I need to know how to configure both the server itself and the client transport (especially the port and path) so that I can connect to it.

As far as I can tell the current assumption is that all non-stdio transport MCP servers will be "remote" (hosted by a third party at a specified endpoint), and I don't think that's the case.  I believe that MCP servers, even those intended to be installed locally, will increasingly take advantage of streamable (in part for things like auth), and that there will also be the need to run shared/multi-user streamable servers to serve a local community (and those still have to be discovered and installed).

tadasant:
[...]

Issue #139, opened 2025-06-13
Plan for schema version changes in the ServerDetail model


joelverhagen:
**Is your feature request related to a problem? Please describe.**

If I am missing a bigger vision here, feel free to close the issue.

The `server.json` request body provided to the `/v0/publish` endpoint is minimally versioned. The shape of the JSON appears to the follow the [ServerDetail](https://github.com/modelcontextprotocol/registry/blob/686fbe1e78117546c91dfb47051e405f0bd1762b/internal/model/model.go#L122C6-L126) model and the related model in the OpenAPI YAML.

There appears to be a coupling between the "input" (publish), "output" (get-by-ID), and DB models in the code (`ServerDetail`).

After the initial launch, I wonder how schema changes will be handled, whether breaking (e.g. property rename) or non-breaking (e.g. new optional property).

Suppose someone publishes a JSON to the registry at the very beginning, then we later introduce a new, required property. The existing JSON in MongoDB will not have the new property but will be returned as-is to the end user (or perhaps break on unmarshal). This could lead to variety on the schema returned to reads of the API.

If the intent is to flow the original schema (as the MCP extension author provided it) to the consumer, then I the current code will work as long as we don't introduce breaking changes.

But what about a property rename? Will a data fix-up/backfill be performed?

**Describe the solution you'd like**

Without trying to fix the whole schema versioning thing at once I propose a schema version be provided at publish time so the publish code/scripts can express concretely "I am giving you a server detail of version X". The server could reject or map the input if the version differs from what is expected. Or this could be included in MongoDB and used for smarter unmarshal at query/read time.

This schema version could be a root JSON property, a query parameter, or an `api-version=0.1` in the `Content-Type` (many options).

Given this is `v0` of the API I think we could allow ourselves to have breaking changes but accept a schema version as input and reject/map old schema versions that could come in to the publish endpoint.

**Describe alternatives you've considered**

Declare `/v0/publish` shape will never change (or never have breaking changes) and describe what "non-breaking changes" might look like.

**Additional context**

I want to describe the current `server.json` schema version in my NuGet + MCP authoring docs so that it's easy for authors to see that they are using an old docs/template version and to check MCP registry docs for what they need to fix up in their server.json.

In other words, I'm planning for potential schema changes :)

tadasant:
Yes, we should definitely version this schema. Thanks for flagging and writing this up.

[...]

Issue #125, opened 2025-06-05
Allow engine or runtime version to be expressed for local MCP servers


joelverhagen:
**Is your feature request related to a problem? Please describe.**

Currently, npm and Python-based MCP servers can be marked as such but there is no indication of the runtime version.  In an npm package, the package.json can declare an `engines` node and define what npm and node versions are needed to run the thing. Similarly there is a `Requires-Python` node in a Python wheel's metadata.

This is not surfaced in the MCP registry metadata, which means that runtime incompatibilities are not handled gracefully.

AFAIK, the best a MCP registry client can do is detect the package type is, say, `npm` and the user has no `npm` installed so it can prompt the user to do so. VS Code does this:

![Image](https://github.com/user-attachments/assets/fe2f048b-44d0-4f03-ad18-efad6d6053b3)
(tested without MCP server context, just with the manual npm install flow)

**Describe the solution you'd like**

Runtime information could be optionally expressed in the `server` node of the OpenAPI definition:
https://github.com/modelcontextprotocol/registry/blob/6b22cf09b376ed94c772c69951a2d125e495b26b/docs/openapi.yaml#L165C5-L200

Making it an arbitrary KVP like `engines` seems like a good start, or we could be more prescriptive on the format per ecosystem.

The client could then enrich the error experience of by noticing that the installed npm version is not compatible with the MCP server and, say, select an older version of the MCP server (and be clear about that) or prompt the user to install a newer npm version.

Tooling could be created to init the server.json from a package.json, filling out the mapped engine properties as needed.

The same concept applies to Python and .NET (the area I am looking at specifically).

**Describe alternatives you've considered**

This could be deferred entirely to the generated single-shot command. So, if `npm` is not installed, VS Code could prompt to install the latest LTS version (nodejs.org/en/download does this). Then, when npm determines the `engines` incompatibility after downloading the package, it could provide its own error message.

But this is at least 2 operations ("install MCP server" then "install npm") into the flow and could be confusing to users that are not expert in the node.js/npm ecosystem but want to use that cool MCP server written in JS. If the MCP client has this context, it could provide a more guided experience.

tadasant:
This is a good callout, thank you!

Should we consider folding it into within `runtime_hint`? i.e. turn that into an object (or array of objects) with command+engine pairings?

joelverhagen:
@connor4312 - do you have thoughts on this? I think you worked on `runtime_hint` also.

[...]

Issue #124, opened 2025-06-04
Support an array of commands to start an MCP server instead of just one


joelverhagen:
**Is your feature request related to a problem? Please describe.**

For some ecosystems, multiple commands are needed to install/update and then run an MCP server. For example, before the upcoming .NET 10 release, it is not possible to run a .NET tool in a single-shot manner.

We are working on [`dnx` analogous to `npx`](https://github.com/dotnet/designs/pull/334) but the requirement for single-shot essentially forced us to add a `dnx` experience. This is probably a net good but I am thinking about other ecosystems or complex MCP setups that would need multiple commands to get running.

**Describe the solution you'd like**

The current schema is pretty specific, not allowing the main command to be set explicitly, except via the `runtime_hint` (AFAIK).

What this means is private MCP registries adhering to the same protocol don't have a spot to put their language-specific runner.

Current:
https://github.com/modelcontextprotocol/registry/blob/6b22cf09b376ed94c772c69951a2d125e495b26b/docs/openapi.yaml#L182-L200

The OpenAPI YAML could express an array of commands instead of just one. A new `command_name` optional property be added (instead of implied by the registry type).

For .NET (before `dnx` exists), the multiple commands to launch a .NET MCP server would be:
```
dotnet tool install --global ErikEJ.DacFX.TSQLAnalyzer.Cli --version 1.0.25
tsqlanalyze -mcp
```

This is not to say the main, public MCP registry will not validate or restrict what commands could be run, but it does allow more flexibility for newer language ecosystems or private deployments of an MCP registry.

Multiple commands unlock these kinds of scenarios, generally:
- Interactive auth for the first command, starting the MCP server as the second
- Install, then run (like pre .NET 10).
- configure a private package source first (`npm config set registry`), then `npx`

Said another way, this removes the single-shot requirement for a language/runtime ecosystem to onboard to MCP server hosting.

**Describe alternatives you've considered**

- Run the setup steps out of band (manually). This is a burden on the end user and makes an single click "install" experience not possible.
- Supporting shell scripting as the entry point (which can internally do multiple commands)
  - This feels to flexible and hidden -- where to host the shell script?
  - It is likely OS specific

**Additional context**
I have opened a similar issue on VS Code (https://github.com/microsoft/vscode/issues/249370) but I now understand this is a more broad discussion than just VS Code.

connor4312:
[...]

PR #153, opened 2025-06-27
First cut at server.json

tadasant:
This is a first cut at properly defining the `server.json` file. We have been working out details of the shape for a while, but to-date it has only been implicitly documented as the OpenAPI response to `/servers/:id`, which is not an exact definition due to some registry-specific details (like `id` and `is_latest`).

Edit: as-written in this PR, this is a more constrained version of `server.json` that the Registry intends to support at-launch. The broader `server.json` may have looser constraints, for example no `enum` constraint on `source=github,gitlab`. See https://github.com/modelcontextprotocol/registry/issues/155

---

## Motivation

There are a variety of use cases where a _static representation of an MCP server_ is necessary:
- Discoverability on a centralized registry (i.e. our official Registry work)
- Discoverability on a decentralized `.well-known` endpoint
- As a response to an initialization call, so the client knows information about the MCP server to which it is connecting
- As an input into crafting a [DXT file](https://www.anthropic.com/engineering/desktop-extensions)
- Packaged in with the source code of an MCP server, so as to have a structured way context 

All of these scenarios (and more) would benefit from an agreed-upon, standardized format that makes it easy to port around and maintain a consistent experience for consumers and developers working with the data. At the end of the day, it's all the same data (or a subset of it). MCP server maintainers should have to manage one copy of this file, and all these use cases can serve that file (or a programmatic derivative/subset of it).

Please note: this is different from the file commonly referred to as `mcp.json`, which is _an MCP client's configuration file for **running** a specific set of MCP servers_. See [this issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/292).

## In this PR

There is a JSON schema for the file (`schema.json`) and a set of examples for how some servers might look (`examples.md`).

This PR does not add any new features/changes that weren't in the codebase already. There are a handful of improvements we can make quickly, but given a bunch of conversations flying about this concept I wanted to get something in place to point to that isn't an OpenAPI spec. I think eventually this spec might do better to live in `modelcontextprotocol/modelcontextprotol` instead of `modelcontextprotocol/registry`, but the registry is probably the most heavily dependent piece of work related to it, so may as well keep it here for now.

## Relevant Issues

Issue: https://github.com/modelcontextprotocol/registry/issues/90

Fast follows next up:
- https://github.com/modelcontextprotocol/registry/issues/155
- https://github.com/modelcontextprotocol/registry/issues/139
- https://github.com/modelcontextprotocol/registry/issues/159

There are definitely some pending in the [registry backlog](https://github.com/modelcontextprotocol/registry/issues) that could impact the shape further. Notably at least:
- https://github.com/modelcontextprotocol/registry/issues/25
- https://github.com/modelcontextprotocol/registry/issues/105
- https://github.com/modelcontextprotocol/registry/issues/143
- https://github.com/modelcontextprotocol/registry/issues/158

And some later issues that are not blockers for the registry work but maybe other folks will be interested in seeing them incorporated into this shape:
- https://github.com/modelcontextprotocol/registry/issues/84
- https://github.com/modelcontextprotocol/registry/issues/94
- https://github.com/modelcontextprotocol/registry/issues/124
- https://github.com/modelcontextprotocol/registry/issues/125

### DIFF
--- a/docs/server-json/README.md
+++ b/docs/server-json/README.md
@@ -5,7 +5,7 @@ There are a variety of use cases where a _static representation of an MCP server
 - Discoverability on a decentralized `.well-known` endpoint
 - As a response to an initialization call, so the client knows information about the MCP server to which it is connecting
 - As an input into crafting a [DXT file](https://www.anthropic.com/engineering/desktop-extensions)
-- Packaged in with the source code of an MCP server, so as to have a structured way context 
+- Packaged in with the source code of an MCP server, so as to have a structured way to identify a server given just its source code
 
 All of these scenarios (and more) would benefit from an agreed-upon, standardized format that makes it easy to port around and maintain a consistent experience for consumers and developers working with the data. At the end of the day, it's all the same data (or a subset of it). MCP server maintainers should have to manage one copy of this file, and all these use cases can serve that file (or a programmatic derivative/subset of it).
 
--- a/docs/server-json/README.md
+++ b/docs/server-json/README.md
@@ -1,0 +1,16 @@
+# server.json file
+
+There are a variety of use cases where a _static representation of an MCP server_ is necessary:
+- Discoverability on a centralized registry (i.e. our official Registry work)
+- Discoverability on a decentralized `.well-known` endpoint
+- As a response to an initialization call, so the client knows information about the MCP server to which it is connecting
+- As an input into crafting a [DXT file](https://www.anthropic.com/engineering/desktop-extensions)
+- Packaged in with the source code of an MCP server, so as to have a structured way context 
+
+All of these scenarios (and more) would benefit from an agreed-upon, standardized format that makes it easy to port around and maintain a consistent experience for consumers and developers working with the data. At the end of the day, it's all the same data (or a subset of it). MCP server maintainers should have to manage one copy of this file, and all these use cases can serve that file (or a programmatic derivative/subset of it).
+
+Please note: this is different from the file commonly referred to as `mcp.json`, which is _an MCP client's configuration file for **running** a specific set of MCP servers_. See [this issue](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/292).
+
+References:
+- [schema.json](./schema.json) - The official JSON schema specification for this representation
+- [examples.md](./examples.md) - Example manifestations of the JSON schema
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -1,0 +1,371 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "MCP Server Detail",
+  "description": "Schema for a static representation of an MCP server. Used in various contexts related to discovery, installation, and configuration.",
+  "type": "object",
+  "required": [
+    "name",
+    "description",
+    "version_detail"
+  ],
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Server name/identifier",
+      "example": "io.modelcontextprotocol/filesystem"
+    },
+    "description": {
+      "type": "string",
+      "description": "Human-readable description of the server's functionality",
+      "example": "Node.js server implementing Model Context Protocol (MCP) for filesystem operations."
+    },
+    "repository": {
+      "type": "object",
+      "description": "Canonical source repository information",
+      "required": [
+        "url",
+        "source",
+        "id"
+      ],
+      "properties": {
+        "url": {
+          "type": "string",
+          "format": "uri",
+          "description": "Repository URL",
+          "example": "https://github.com/modelcontextprotocol/servers"
+        },
+        "source": {
+          "type": "string",
+          "enum": [
+            "github",
+            "gitlab"
+          ],
+          "description": "Repository hosting service",
+          "example": "github"
+        },
+        "id": {
+          "type": "string",
+          "description": "Repository identifier unique and immutable in the corresponding source",
+          "example": "b94b5f7e-c7c6-d760-2c78-a5e9b8a5b8c9"
+        }
+      }
+    },
+    "version_detail": {
+      "type": "object",
+      "description": "Version information for this server",
+      "required": [
+        "version",
+        "release_date",
+        "is_latest"
+      ],
+      "properties": {
+        "version": {
+          "type": "string",
+          "description": "Server version (equivalent to Implementation.version in MCP spec)",
+          "example": "1.0.2"
+        },
+        "release_date": {
+          "type": "string",
+          "format": "date-time",
+          "description": "When this version was published to the registry",
+          "example": "2023-06-15T10:30:00Z"
+        }
+      }
+    },
+    "packages": {
+      "type": "array",
+      "description": "Available package distributions",
+      "items": {
+        "type": "object",
+        "required": [
+          "registry_name",
+          "name",
+          "version"
+        ],
+        "properties": {
+          "registry_name": {
+            "type": "string",
+            "enum": [
+              "npm",
+              "docker",
+              "pypi",
+              "homebrew"
+            ],
+            "description": "Package registry type",
+            "example": "npm"
+          },
+          "name": {
+            "type": "string",
+            "description": "Package name in the registry",
+            "example": "@modelcontextprotocol/server-filesystem"
+          },
+          "version": {
+            "type": "string",
+            "description": "Package version",
+            "example": "1.0.2"
+          },
+          "runtime_hint": {
+            "type": "string",
+            "description": "Hint for appropriate runtime (e.g., npx, uvx)",
+            "examples": [
+              "npx",
+              "uvx"
+            ]
+          },
+          "runtime_arguments": {
+            "type": "array",
+            "description": "Arguments for the runtime command (docker, npx, etc.)",
+            "items": {
+              "oneOf": [
+                {
+                  "type": "object",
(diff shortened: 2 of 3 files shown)


tadasant:
cc @joelverhagen @felixrieseberg

atrawog:
That's quite a sophistication list that's covering almost everything. But the one thing that could be still useful to add is some info about the supported OAuth Identify Provider(s).

Because we are  likely going to see a rise of MCP Server with some form of OAuth authentication. Who either require the user to create a user account on a certain website. Or require an OAuth callback configuration from an Identify provider for user identification.

tadasant:
@atrawog yes, that is in:

> And there are definitely some pending in the [registry backlog](https://github.com/modelcontextprotocol/registry/issues) that could impact the shape further. Notably at least:
> [OAuth Config (Optional) for Remotes #25](https://github.com/modelcontextprotocol/registry/issues/25)

I think that would cover what you are describing?

docs/server-json/schema.json
```diff
@@ -0,0 +1,371 @@
+{
```
connor4312, line 1: Can we have a script that auto-gens the overlapping portion of this from the openapi spec or vise versa? The schema is a little involved and I would not want them to accidentally diverge

connor4312:
I think this looks good overall, one comment :)

docs/server-json/schema.json
tadasant: Definitely feel that pain: I think I'd rather we have a script that's in CI validating that they are consistent though (e.g. that the OpenAPI spec's /servers/:id returns a shape that is compliant with server.json), rather than literally autogenerate one or the other. Let me know if you feel strongly otherwise

Will add that validation as a third fast follow here

connor4312: If we make changes to one or the other, I would prefer not to have to update both YAML and JSON schema to do so. I think a nice autogen story could be good, but we could also go with a validation step and then add that later on if it's a pain point.

docs/server-json/README.md
```diff
@@ -0,0 +1,16 @@
+# server.json file
+
+There are a variety of use cases where a _static representation of an MCP server_ is necessary:
+- Discoverability on a centralized registry (i.e. our official Registry work)
+- Discoverability on a decentralized `.well-known` endpoint
+- As a response to an initialization call, so the client knows information about the MCP server to which it is connecting
+- As an input into crafting a [DXT file](https://www.anthropic.com/engineering/desktop-extensions)
+- Packaged in with the source code of an MCP server, so as to have a structured way context
```
jonathanhefner: Missing words... Perhaps?

```suggestion
- Packaged in with the source code of an MCP server, so as to have a standard way to preload server information
```

docs/server-json/schema.json
```diff
@@ -0,0 +1,371 @@
...
+                    "value_hint"
+                  ],
+                  "properties": {
+                    "type": {
+                      "const": "positional"
+                    },
+                    "value_hint": {
+                      "type": "string",
+                      "description": "Identifier hint for the value"
+                    },
```
jonathanhefner: I wasn't fully clear on the functionality of `value_hint`, so I dug and found its [original description](https://github.com/modelcontextprotocol/registry/pull/33/files#diff-6c08e65eef0374988297d041bd50f87b385613a7e478c3812cb5312cc3e1f1e3R266):

> An identifier-like hint for the value. This is not part of the command line, but can be used by client configuration and to provide hints to users.

Is there a historical reason to prefer the terser description?

docs/server-json/schema.json
toby: I think we should be able to use `go generate` for this, [link](https://go.dev/blog/generate).

docs/server-json/README.md
tadasant: Oops thank you!

docs/server-json/schema.json
tadasant: No, good callout, bringing back the full description. Thanks!

tadasant:
Addressed feedback and made some simplifications in 8041f5f359d241862a5de1ddc8574fe79c4c44b0. Thanks all for taking a look!

docs/server-json/schema.json
```diff
@@ -0,0 +1,306 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "MCP Server Detail",
```
joelverhagen, line 3: Could we add an `$id` to this JSON schema, ideally pointing to where the this schema hosted (URL & URI), and maybe including a version, such as the date of release (maybe today). This could be included optionally at publish time for improved error experience (other than complaining of some missing property).

docs/server-json/schema.json
```diff
@@ -0,0 +1,306 @@
...
+      "properties": {
+        "url": {
+          "type": "string",
+          "format": "uri",
+          "description": "Repository URL",
+          "example": "https://github.com/modelcontextprotocol/servers"
+        },
+        "source": {
+          "type": "string",
+          "enum": [
```
joelverhagen, line 39: Is this a description of what the main public registry will support? Or what any MCP registry can support including private ones? It seems to me like this should just be a string and the set of allowed values is enforced by a specific registry implementation. 

Asked another way, do you expect this JSON schema to be used by other deployments or just the main one?

docs/server-json/schema.json
```diff
@@ -0,0 +1,306 @@
...
+      "type": "object",
+      "description": "Version information for this server",
+      "required": [
+        "version",
+        "release_date"
+      ],
+      "properties": {
+        "version": {
+          "type": "string",
+          "description": "Server version (equivalent to Implementation.version in MCP spec)",
```
joelverhagen, line 63: Is this supposed to be a SemVer style version? Or is it any versioning scheme? Based on the implementation today appears to be lex order compared to determine latest.

tadasant: Probably SemVer but I haven't thought deeply on the options and nuances here

> Add a version to the scheme of server.json itself (https://github.com/modelcontextprotocol/registry/issues/139)

If you have strong opinions and are willing to make a PR laying out the pro's/con's, would love for you to take that pending followup to this PR :) Seemed like @jonathanhefner probably has an opinion here too. This decision likely has much wider implications than just registry so we should probably carve it out to a separate issue and be thorough with it

docs/server-json/schema.json
tadasant: Good question. I guess we're going to need two schemas here, similar to how I was thinking we need to separate out the [official vs. broader OpenAPI schemas](https://github.com/modelcontextprotocol/registry/issues/149). As-written, this is moreso the public registry one (e.g. constraints like enum on `source` don't make sense for the generic server.json). 

Added ticket: https://github.com/modelcontextprotocol/registry/issues/155

docs/server-json/schema.json
tadasant: Yes, that's the intent of this fast follow:

> Add a version to the scheme of server.json itself (https://github.com/modelcontextprotocol/registry/issues/139)

docs/server-json/schema.json
jonathanhefner: Actually, from the end user's perspective, I think it would be preferable to allow any format.  I think the core issue is how the registry tracks version history.

The FAQ says version numbers should be ["bumped"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L66) or ["incremented"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L72), which, obviously, requires version numbers to have an ordering scheme.  However, since the registry [only tracks which version is "latest"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L68) (i.e. doesn't support branching), we could just base "latest" off of the submission date, and change the version number requirement to "unique" instead of "incremented".  But perhaps the intention was to leave the door open to support branching in the future?

joelverhagen: @tadasant - sorry for two comments on version :D. This thread is talking about the server version scheme, as @jonathanhefner is responding about. I saw this part of the code:
https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/internal/database/mongo.go#L204-L206
I am not a Go expert, but this appears to do lex comparison which is pretty strange for a version string. `2.0.0` would be > `10.0.0`. 

I think it's reasonable to say latest ~= most recent, not "greatest version per some arbitrary versioning scheme". That would be simpler. I like SemVer and prefer it, but do not know the needs of the broader ecosystem here.

tadasant: Oops sorry I pasted the wrong link in my last comment. Meant to paste

> Propose a specific versioning scheme (https://github.com/modelcontextprotocol/registry/pull/134/files#r2145287494)

Regarding the code -- it's all just very POC at the moment. These docs are the stronger source of truth (and we'll eventually bring the implementation to align).

tadasant: https://github.com/modelcontextprotocol/registry/issues/158

tadasant:
@toby or @connor4312 when you get a chance to 👍 we can move this in and then I'll move onto:
- https://github.com/modelcontextprotocol/registry/issues/155
- https://github.com/modelcontextprotocol/registry/issues/159

PR #167, opened 2025-07-09
Split "general" server.json from "registry-specific" server.json

### LINKED ISSUE #155: Separate out Registry-specific `server.json` out from generic `server.json`
Analogous to #149, we should have a (more constrained) schema for the official Registry vs. a generic `server.json` that will serve other use cases.

See comment: https://github.com/modelcontextprotocol/registry/pull/153/files/8041f5f359d241862a5de1ddc8574fe79c4c44b0#r2172966211

Related: https://github.com/modelcontextprotocol/registry/pull/153

tadasant:
Closes #155

The "generalized" `server.json` is the one that can be used in a variety of situations:

- Discoverability on a centralized registry (i.e. our official Registry work)
- Discoverability on a decentralized .well-known endpoint
- As a response to an initialization call, so the client knows information about the MCP server to which it is connecting
- As an input into crafting a [DXT file](https://www.anthropic.com/engineering/desktop-extensions)
- Packaged in with the source code of an MCP server, so as to have a structured way context 

Of these, the official Registry is but one use case. The Registry has a unique set of concerns for which we are optimizing, like whitelisting of specific registries we trust to host MCP server source code for the public community (`npm`, `pypi`, etc). It is not necessary to have such a constraint for the generalized server.json, which may be used in, for example, an internal context where such a validation doesn't make sense.

So I've done the following at a high level:
- Renamed the prior `schema.json` to be `registry-schema.json`
- Introduced a more generic `schema.json` to serve as the schema for the generalized `server.json`
- Refactored `registry-schema.json` to be simply a constraint on the base `schema.json`

I made a few small changes in this transition:
- Removed `gitlab` as an option in `repository.source` for `registry-schema.json`. I'm fine going live with just GitHub support; would be happy to take a contribution from e.g. Gitlab employees or a motivated Gitlab customer to expand anytime.
- Removed `homebrew` as an option in `packages.registry_name`; never really made sense, not a place where folks are publishing MCP servers
- Added `docker` to `packages.runtime_hint`; was just a missing oversight
- Gave both schemas a global `$id`
~- Removed `release_date` from both schemas. Certainly not something that makes sense for the generalized schema. It actually doesn't make sense for the Registry schema either, because this is the server.json that _users will submit_, i.e. the _input_ into the Registry publication API. `release_date` will be a piece of _output_ metadata. The Registry API will be free to append such extra metadata for GET /:id requests of servers, and the data can be present in the OpenAPI spec, but it does not need to be in the registry-server.json JSON schema.~
  - Note this is no longer in this PR; rebased on #168 which included it
- Removed unnecessary enums from the generalized `schema.json`

### DIFF
--- a/docs/server-json/registry-schema.json
+++ b/docs/server-json/registry-schema.json
@@ -1,0 +1,40 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://modelcontextprotocol.io/schemas/draft/2025-07-09/registry-server.json",
+  "title": "MCP Server Detail - Registry Schema",
+  "description": "Registry-specific constraints for MCP server representation. Extends the base schema with additional validation rules.",
+  "$ref": "https://modelcontextprotocol.io/schemas/draft/2025-07-09/server.json",
+  "properties": {
+    "repository": {
+      "properties": {
+        "source": {
+          "enum": [
+            "github"
+          ]
+        }
+      }
+    },
+    "packages": {
+      "items": {
+        "properties": {
+          "registry_name": {
+            "enum": [
+              "npm",
+              "pypi",
+              "docker",
+              "nuget"
+            ]
+          },
+          "runtime_hint": {
+            "enum": [
+              "npx",
+              "uvx",
+              "docker",
+              "dnx"
+            ]
+          }
+        }
+      }
+    }
+  }
+}
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -1,5 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://modelcontextprotocol.io/schemas/draft/2025-07-09/server.json",
   "title": "MCP Server Detail",
   "$ref": "#/$defs/ServerDetail",
   "$defs": {
@@ -18,10 +19,7 @@
         },
         "source": {
           "type": "string",
-          "enum": [
-            "github",
-            "gitlab"
-          ],
+          "description": "Repository hosting service",
           "example": "github"
         },
         "id": {
@@ -32,6 +30,7 @@
     },
     "VersionDetail": {
       "type": "object",
+      "description": "Version information for this server. Defined as an object to allow for downstream extensibility (e.g. release_date)",
       "required": [
         "version"
       ],
@@ -79,13 +78,6 @@
       "properties": {
         "registry_name": {
           "type": "string",
-          "enum": [
-            "npm",
-            "docker",
-            "pypi",
-            "homebrew",
-            "nuget"
-          ],
           "description": "Package registry type",
           "example": "npm"
         },
--- a/docs/server-json/README.md
+++ b/docs/server-json/README.md
@@ -14,3 +14,4 @@ Please note: this is different from the file commonly referred to as `mcp.json`,
 References:
 - [schema.json](./schema.json) - The official JSON schema specification for this representation
 - [examples.md](./examples.md) - Example manifestations of the JSON schema
+- [registry-schema.json](./registry-schema.json) - A more constrained version of `schema.json` that the official registry supports


tadasant:
cc @joelverhagen @felixrieseberg

docs/server-json/schema.json
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://modelcontextprotocol.io/schemas/draft/server.json",
```
joelverhagen: nit: maybe put the current month in here like
```suggestion
  "$id": "https://modelcontextprotocol.io/schemas/draft/2025-07/server.json",
```

We can perhaps make breaking changes to the schema and leave the `$id` as is (or change it and forget about the old date) until the Registry is live.

docs/server-json/schema.json
```diff
@@ -52,22 +49,15 @@
...
-        "release_date"
+        "version"
       ],
       "properties": {
         "version": {
           "type": "string",
           "description": "Server version (equivalent to Implementation.version in MCP spec)",
           "example": "1.0.2"
-        },
-        "release_date": {
```
joelverhagen: Oh, haha, I just found this in my PR. https://github.com/modelcontextprotocol/registry/pull/168.

I can rebase my PR after you merge this.

docs/server-json/registry-schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://modelcontextprotocol.io/schemas/draft/registry-server.json",
+  "title": "MCP Server Detail - Registry Schema",
+  "description": "Registry-specific constraints for MCP server representation. Extends the base schema with additional validation rules.",
+  "$ref": "https://modelcontextprotocol.io/schemas/draft/server.json",
+  "properties": {
```
joelverhagen, line 7: Cool, I didn't know you could do overrides like this.

joelverhagen:
Looks good to me.

docs/server-json/schema.json
tadasant: My idea here would be to align `server.json` with the official MCP versions. And since 2025-06-18 is already out, it's only conceivable to be building for `draft` right now, with an aim to lock it in with the next MCP release (which I know the core team is eager to make a regular, time-based release). I think we pretty much need to do this if we want server.json to have the broad applicability we are aiming for here.

In theory, we could have separate versioning for `registry-schema.json` and evolve it off-MCP-cycle, so maybe there's merit to considering that separately... but I still think keeping it `draft` before we have an "0.1 version release" is fine though no? What's the purpose of versioning explicitly if we're calling it draft anyway?

joelverhagen: Ah, makes sense. I was thinking that having the date here would allow people to put `"$schema": "https://modelcontextprotocol.io/schemas/draft/2025-07/server.json"` in the server.json and then they could know that they are using an old version, when the next draft comes out (or perhaps as the draft changes prior to "0.1 version release"). It seems like your approach is effectively stating "prerelease/draft = unversioned". I think it could still be versioned, but the word "draft" signified breaking changes could come.

At a high level, I think it would be nice to know what version people coded their server.json against even if it is prerelease/draft but maybe I am overthinking it.

tadasant: Fair - you might be right; given we have been iterating on this for quite a while now, I think it'd be nice to get some identifier in there so we're not blocking folks from starting to do work like you are doing downstream of server.json. I'm ok with us putting a date where you suggested (I think date makes more sense than month, as we are very likely to make more-frequent-than-monthly updates). Will add, thanks!

tadasant: https://github.com/modelcontextprotocol/registry/pull/167/commits/177770e94611cf51c4fd8139d32893202220a058

tadasant:
Did a rebase on #168 which makes this PR a lot simpler :)

sridharavinash (approved):

Issue #159, opened 2025-06-30
Add a validation script that ensures OpenAPI, example, server.json, examples are all in sync with each other


tadasant:
See discussion: https://github.com/modelcontextprotocol/registry/pull/153#discussion_r2170807733

domdomegg:
I think this is done now, as per #198 -> #105 -> #199 🎉 

Reopen if I've misunderstood :)

Issue #158, opened 2025-06-30
Decide & document server versioning approach for server.json


tadasant:
`server.json` has a `version_detail.version` field that is currently just defined as a string, with a rough expectation that it is some flavor of SemVer.

@jonathanhefner [raised](https://github.com/modelcontextprotocol/registry/pull/134/files#r2145287494):

> Is there any place that we specify the version number format? For example, would we support "1.0.0-alpha" => "1.0.0-beta"?

@joelverhagen:

> I had assumed SemVer, but this is a nuanced problem. I imagine folks will want to use the same version string as the underlying package for simplicity. And there is no guarantee of a unified version model for N package ecosystems. For example NuGet is SemVer-ish with case insensitive and a 4th version digit supported.
>
> Perhaps complaint SemVer 2.0.0 for MCP Registry, and build metadata rejected? (1.0.0-alpha but not 1.0.0-alpha+foo which SemVer 2.0.0 supports but adds complexity).

Me:

> I think it is actually unlikely that folks will want to bind the MCP server version with their package version. There may be cases where the MCP server version gets incremented even though the code doesn't change (for example, in the case of me deciding to update the description. It would be limiting myself to require that I make a code change to make that MCP server-level metadata change.

@jonathanhefner [here](https://github.com/modelcontextprotocol/registry/pull/153#discussion_r2173341792):

> Actually, from the end user's perspective, I think it would be preferable to allow any format. I think the core issue is how the registry tracks version history.
>
> The FAQ says version numbers should be ["bumped"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L66) or ["incremented"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L72), which, obviously, requires version numbers to have an ordering scheme. However, since the registry [only tracks which version is "latest"](https://github.com/modelcontextprotocol/registry/blob/443159fed1d70c7d202e54c363a80775fcdf9b27/docs/faq.md#L68) (i.e. doesn't support branching), we could just base "latest" off of the submission date, and change the version number requirement to "unique" instead of "incremented". But perhaps the intention was to leave the door open to support branching in the future?

I had said this would be a fast-follow to https://github.com/modelcontextprotocol/registry/pull/153 but I think it merits its own discussion as an Issue first

domdomegg:
[...]

PR #168, opened 2025-07-09
Align the server.json schema with the YAML

joelverhagen:
## Motivation and Context

This is in support of https://github.com/modelcontextprotocol/registry/issues/159, to make it easier, I hope. My team noticed some differences in the JSON schema, most notably the "choices" property was missing. I tried to write a tool to do this but slowed down since I am not experienced with Go. This is a first pass, maybe someone else can finish the tooling.

I restructured the JSON schema to break out child objects like the YAML file, and brought the descriptions and example into sync. The only differences are two extra properties in the YAML which are computed by the server:

- `VersionDetail.is_latest` is not present in the JSON schema, computed by the registry based on https://github.com/modelcontextprotocol/registry/issues/158
- `Server.id` is not present in the JSON schema, generated by the registry

The idea is the a tool could look at ServerDetail in the YAML and then generate the schema.json with minimal fix-ups (add a root title, remove `VersionDetail.is_latest` and `Server.id`.

~Question for @tadasant - should `VersionDetail.release_date` be in the JSON schema? I think _no_ (and let it be computed) but it's there right now.~ - aligned with https://github.com/modelcontextprotocol/registry/pull/167 and removed

## How Has This Been Tested?

I have tested the JSON schema against the sample server.json in the repo (files and markdown blocks).

## Breaking Changes

None.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -33,20 +33,13 @@
     "VersionDetail": {
       "type": "object",
       "required": [
-        "version",
-        "release_date"
+        "version"
       ],
       "properties": {
         "version": {
           "type": "string",
           "example": "1.0.2",
           "description": "Equivalent of Implementation.version in MCP specification."
-        },
-        "release_date": {
-          "type": "string",
-          "format": "date-time",
-          "example": "2023-06-15T10:30:00Z",
-          "description": "Datetime that the MCP server version was published to the registry."
         }
       }
     },


joelverhagen:
@connor4312 - does this look reasonable to you?

tadasant (approved):
This all makes sense to me! Thanks for restructuring this.

I intend to take a whack at the generation soon, will see how I go wielding Go code with Claude Code...

Issue #169, opened 2025-07-09
Unclear schema for fixed (constant) CLI arguments


joelverhagen:
**Describe the bug**

Suppose you have an MCP server that has `mcp start` as required CLI arguments. Perhaps the MCP server is a CLI tool package with multiple purposes, not just MCP and `mcp start` invokes the MCP server subroutine.

I believe these would be expressed as positional arguments with a `value`. But `value_hint` is required.
https://github.com/modelcontextprotocol/registry/blob/f6fa98c9145e25dac01e9f9177155236e9be3b7d/docs/server-json/schema.json#L233-L243

VS Code appears to prefer `value` and tolerate a missing `value_hint`:
https://github.com/microsoft/vscode/blob/1ea38ea6831a8c05cb071e9219f11ad546b1b748/src/vs/platform/mcp/common/mcpManagementService.ts#L275-L281

So perhaps `value_hint` can become optional? Or maybe we need a `ConstantArgument` which has a required `value` and `type = constant` but no `name` or `runtime_hint`?

**To Reproduce**

Here is a sample server.json that shows the clunkiness.

```json
{
  "description": "A sample MCP server using the MCP C# SDK. Generates random numbers and random weather.",
  "name": "io.github.joelverhagen/Knapcode.SampleMcpServer",
  "packages": [
    {
      "registry_name": "nuget",
      "name": "Knapcode.SampleMcpServer",
      "version": "0.4.0-beta",
      "package_arguments": [
        {
          "type": "positional",
          "value_hint": "mcp",
          "value": "mcp"
        },
        {
          "type": "positional",
          "value_hint": "start",
          "value": "start"
        }
      ]
    }
  ],
  "version_detail": {
    "version": "0.4.0-beta"
  }
}
```

**Expected behavior**

There should be clear guidance on fixed (non-input) positional arguments.

PR #172, opened 2025-07-09
Make value_hint optional to allow constant positional arguments

joelverhagen:
## Motivation and Context

This is one attempt to resolve https://github.com/modelcontextprotocol/registry/issues/169.

Instead of adding a new "constant" argument type, I expanded the existing positional one and just dropped the `value_hint` requirement. I am not sure if this has VS Code impact @sandy081, @connor4312.

## How Has This Been Tested?

I tested the current server.json samples in the repo against the new (less strict) schema. Of course they passed (aside from an unrelated bug in my previous NuGet sample 🤦).

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -204,8 +204,7 @@
         {
           "type": "object",
           "required": [
-            "type",
-            "value_hint"
+            "type"
           ],
           "properties": {
             "type": {
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -7,8 +7,7 @@
       "type": "object",
       "required": [
         "url",
-        "source",
-        "id"
+        "source"
       ],
       "properties": {
         "url": {
--- a/docs/server-json/schema.json
+++ b/docs/server-json/schema.json
@@ -7,7 +7,8 @@
       "type": "object",
       "required": [
         "url",
-        "source"
+        "source",
+        "id"
       ],
       "properties": {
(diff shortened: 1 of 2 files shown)


docs/server-json/schema.json
```diff
@@ -204,8 +204,7 @@
         {
           "type": "object",
           "required": [
-            "type",
-            "value_hint"
```
connor4312, line 200: `value_hint` is technically extraneous for 'constant' arguments, but we should definitely have it required otherwise since that's how clients can index and reference those arguments when forming the template

joelverhagen: Could you explain how value_hint is used for this? This sounded kind of like an HTML `placeholder` value to default an input, instead of a key used for referencing. Or maybe I misunderstand you.

Maybe something like `value` could be used as a fallback? essentially a positional argument must have either a value or a value_hint?

connor4312: `value_hint` is needed to provide a name for positional arguments in cases where the user needs to input them and the client needs to remember them. Without this we only have the index which is quite subject to change (and breakage) between versions of the package.

> essentially a positional argument must have either a value or a value_hint?

Yes, I think updating the schema to reflect this would be okay

joelverhagen: I have updated the schema to reflect this.

tadasant (approved):

PR #176, opened 2025-07-12
Add validation tools for server.json schemas and examples

tadasant:
This is a first PR for #159 - adding some validation scripts. I want to get to a place where we are confident that making schema changes is in sync with the rest of our code. Step one is making sure the schemas are in sync with our examples. Step two will be making sure the schemas are in sync with our actual Registry implementation code.

This is just for the JSON right now; if we're feeling good about this approach I'll do an analogous change for the OpenAPI schemas as well.

I'll also follow with adding usage of these scripts to our CI pipeline so we have automated checks on them with every change.

## Summary
- Added CLI validation tools for JSON schemas and examples in `docs/server-json/`
- Fixed some minor validation errors found in `examples.md`
- Ensures schema consistency and example validity in CI/CD pipelines

### New Validation Tools
1. **`tools/validate-schemas`** - Validates that `schema.json` and `registry-schema.json` are valid JSON Schema documents
2. **`tools/validate-examples`** - Validates all JSON examples in `examples.md` against both schemas

### Key Features
- ✅ Validates JSON schema syntax and structure
- ✅ Validates all examples against both base and registry schemas
- ✅ Tracks validation counts to prevent silent failures
- ✅ Proper exit codes for CI integration
- ✅ Clear error messages for debugging

### Fixes to examples.md
- Added missing `repository.id` field in NuGet example
- Changed `repository.source` from "gitlab" to "github" in Docker example

### Usage
```bash
# From repository root
./tools/validate-schemas.sh
./tools/validate-examples.sh
```

## Testing

I ran these scripts locally with minor (breaking) changes, which they each correctly reported. As submitted, they are both passing.

🤖 Co-authored with [Claude Code](https://claude.ai/code)


validate-examples
jonathanhefner, line 1: Is this binary file a compiled version of `tools/validate-examples/main.go`?  If so, should it be removed or added to `.gitignore`?

validate-schemas
jonathanhefner, line 1: Ditto for this binary.

validate-examples
tadasant: Yes, thanks for catching! https://github.com/modelcontextprotocol/registry/pull/176/commits/4a8044777ce2f707960191dc604b206055725779

validate-schemas
tadasant: https://github.com/modelcontextprotocol/registry/pull/176/commits/4a8044777ce2f707960191dc604b206055725779

tools/validate-examples/main.go
```diff
@@ -0,0 +1,191 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to compile registry-schema.json: %w", err)
+	}
+
+	allValid := true
+	validatedCount := 0
+	for i, example := range examples {
+		log.Printf("Example %d:", i+1)
+
+		var data interface{}
```
chlowell:
```suggestion
		var data any
```

tools/validate-schemas/main.go
```diff
@@ -0,0 +1,108 @@
...
+	return nil
+}
+
+func validateSchema(compiler *jsonschema.Compiler, path string) error {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	var schemaData interface{}
```
chlowell:
```suggestion
	var schemaData any
```

tools/validate-schemas.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Validate JSON schema files
+# For more information, see docs/server-json/README.md
+
+set -e
+
+cd "$(dirname "$0")/.."
+exec go run tools/validate-schemas/main.go "$@"
```
chlowell:
```suggestion
exec go run tools/validate-schemas/main.go "$@"

```

tools/validate-examples.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Validate examples in docs/server-json/examples.md
+# For more information, see docs/server-json/README.md
+
+set -e
+
+cd "$(dirname "$0")/.."
+exec go run tools/validate-examples/main.go "$@"
```
chlowell:
```suggestion
exec go run tools/validate-examples/main.go "$@"

```

tools/validate-schemas/main.go
```diff
@@ -0,0 +1,108 @@
...
+	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	log.SetFlags(0) // Remove timestamp from logs
+
+	var rootCmd = &cobra.Command{
```
chlowell: `cobra` feels excessive for a program having no parameters

tools/validate-schemas/main.go
```diff
@@ -0,0 +1,108 @@
...
+	schemas := []struct {
+		name string
+		path string
+	}{
+		{"schema.json", filepath.Join(basePath, "schema.json")},
+		{"registry-schema.json", filepath.Join(basePath, "registry-schema.json")},
+	}
+
+	expectedSchemaCount := len(schemas)
+	compiler := jsonschema.NewCompiler()
```
chlowell: consider moving this into `validateSchema`; it's cheap to construct and used only in that function

tools/validate-schemas/main.go
```diff
@@ -0,0 +1,108 @@
...
+	compiler.Draft = jsonschema.Draft7
+
+	allValid := true
```
chlowell: Do you need this? `validatedCount` counts valid schemas, so I expect `allValid == (validatedCount == expectedSchemaCount)`

tools/validate-schemas/main.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("failed to read file: %w", err)
+	}
+
+	var schemaData interface{}
+	if err := json.Unmarshal(data, &schemaData); err != nil {
+		return fmt.Errorf("invalid JSON: %w", err)
+	}
+
+	// For registry-schema.json, we need to register the base schema it references
+	if strings.Contains(path, "registry-schema.json") {
```
chlowell, line 76: It would be better to make the additional resource the caller's responsibility so that this function doesn't need foreknowledge of its inputs. I see that may not be worth the extra effort though, so I'm fine with this

tools/validate-examples/main.go
```diff
@@ -0,0 +1,191 @@
...
+					line:    startLine,
+				})
+			}
+		case inCodeBlock:
+			if currentExample.Len() > 0 {
+				currentExample.WriteString("\n")
+			}
+			currentExample.WriteString(line)
+		}
+	}
```
chlowell: This might be easier to maintain as a regex

tools/validate-examples/main.go
tadasant: Thanks, added lint rule too https://github.com/modelcontextprotocol/registry/pull/176/commits/7b2fd9f76fa1d36efb0f28ac667887f371d6bc1a

tools/validate-schemas/main.go
tadasant: https://github.com/modelcontextprotocol/registry/pull/176/commits/7b2fd9f76fa1d36efb0f28ac667887f371d6bc1a

tools/validate-schemas/main.go
tadasant: Fair, thanks, removed from both https://github.com/modelcontextprotocol/registry/pull/176/commits/88da4dcc28085b2910f094d8c1716bd400ba205e

tools/validate-schemas/main.go
tadasant: No you're right, I added validatedCount after my first pass and should have cleaned up. Fixed both: https://github.com/modelcontextprotocol/registry/pull/176/commits/59dc57fe7e83d5ee105a61c731d776d3d76bb894

tools/validate-examples/main.go
tadasant: That's fair: https://github.com/modelcontextprotocol/registry/pull/176/commits/3042bf75e13f303c07f27ef80f576209299c6526

tools/validate-schemas/main.go
tadasant: https://github.com/modelcontextprotocol/registry/pull/176/commits/72798816b14f8283ac3a925519654759e6374c51

tools/validate-schemas/main.go
tadasant: I will keep this for now as I don't expect these relationships between the files to change much, if at all, but can revisit later on

tadasant:
Thanks for the review @chlowell ! This is my first-ever Go PR (coming from a TypeScript/Ruby and some Python career), so all the feedback is very welcome :)

tools/validate-examples/main.go
```diff
@@ -0,0 +1,172 @@
...
+				// This should never happen since we found the match, but be safe
+				continue
+			}
+			beforeMatch := content[:index]
+			lineNumber := strings.Count(beforeMatch, "\n") + 2 // +2 for ```json line and 1-based indexing
+
+			examples = append(examples, example{
+				content: match[1],
+				line:    lineNumber,
+			})
```
chlowell: [FindAllStringSubmatchIndex](https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatchIndex) makes this a little simpler because it returns indices of matching content instead of that content itself:
```suggestion
	matches := re.FindAllStringSubmatchIndex(content, -1)

	var examples []example
	for _, match := range matches {
		if len(match) < 4 {
			// should never happen
			return nil, fmt.Errorf("invalid match - expected at least 4 indices but got %d", len(match))
		}
		start, end := match[2], match[3]
		// line numbers start at 1
		line := 1 + strings.Count(content[:start], "\n")
		examples = append(examples, example{
			content: content[start:end],
			line:    line,
		})
```

tadasant:
Thanks @chlowell - one more 👍 when you have a sec

PR #206, opened 2025-07-29
feat: implement 128-bit cryptographically secure token generation for…

aphansal123:
Closes [#22191](https://github.com/github/copilot/issues/22191)

This PR implements cryptographically secure 128-bit token generation for domain ownership verification in the MCP Registry. The package provides a single, focused function for generating base64url-encoded tokens suitable for both DNS TXT record and HTTP-01 challenge verification methods.

## Motivation and Context

The MCP Registry requires a robust domain verification system to ensure that server publishers actually own the domains they claim. This package provides the foundational token generation capability needed for:

1. **DNS TXT Record Verification**: Publishers add `mcp-verify=<token>` to their domain's DNS
2. **HTTP-01 Web Challenge**: Publishers serve tokens at `https://domain/.well-known/mcp-challenge/<token>`

The implementation follows security best practices with 128-bit entropy and base64url encoding for maximum compatibility with both DNS and HTTP standards.

## How Has This Been Tested?

- **Unit tests**: Comprehensive test suite covering token generation, uniqueness, entropy validation, and error handling
- **Security tests**: Verified 128-bit entropy, cryptographically secure randomness, and proper base64url encoding
- **Compatibility tests**: Ensured tokens are safe for both DNS TXT records and HTTP URLs
- **Performance tests**: Benchmarked token generation (~196ns/op on Apple M4 Max)
- **Integration tests**: Verified the package builds and integrates correctly with the larger registry codebase
- **Linting**: Passed all Go linting tools (go vet, gofmt, golangci-lint)

## Breaking Changes

No breaking changes. This is a new package that doesn't affect existing functionality.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
**Design Decisions:**
- **Simplified API**: Based on PR feedback, removed `ValidateTokenFormat` and `TokenInfo` struct to keep the API minimal and focused on core functionality
- **String comparison**: Token validation is handled through simple string comparison rather than format validation, as suggested in code review
- **128-bit entropy**: Chosen to provide cryptographically secure randomness while maintaining reasonable token length (22 characters)
- **Base64url encoding**: Ensures tokens are safe for both DNS TXT records and HTTP URLs without escaping

**Security Considerations:**
- Uses `crypto/rand` for cryptographically secure random number generation
- Tokens should be treated as single-use and time-limited
- Always use HTTPS when transmitting tokens
- 2^128 possible values make collision probability negligible

**Performance:**
- Token generation is extremely fast (~196ns/op) and suitable for real-time web applications
- No external dependencies


internal/verification/README.md
```diff
@@ -0,0 +1,197 @@
...
+
+Benchmark results on Apple M4 Max:
+
+```
+BenchmarkGenerateVerificationToken-16    5726528    196.1 ns/op
+BenchmarkValidateTokenFormat-16          98329761   12.31 ns/op
+BenchmarkGenerateTokenWithInfo-16        4017357    290.5 ns/op
+```
+
+Note: These benchmark results are provided as examples and were obtained on an Apple M4 Max system. Performance may vary significantly on different hardware configurations.
```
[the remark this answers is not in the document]
aphansal123: The performance tests were run on an Apple M4 system which currently exists.

internal/verification/token.go
```diff
@@ -0,0 +1,143 @@
...
+//
+// Parameters:
+// - token: The token string to validate
+//
+// Returns:
+// - true if the token format is valid, false otherwise
+//
+// Note: This function only validates format, not authenticity or ownership.
+// Actual verification requires checking the token against stored values.
+func ValidateTokenFormat(token string) bool {
```
thomas-sickert: What do you think about just using the base64 decode method to validate the format instead of a bespoke validation method? In fact, I'm not sure that we need to explicitly validate the format of the token at all. All that really matters is the comparison when validating the token itself. We might be able to remove some code if we only rely on the comparison during the validation.

internal/verification/token.go
```diff
@@ -0,0 +1,143 @@
...
+			char == '-' || char == '_') {
+			return false
+		}
+	}
+
+	return true
+}
+
+// TokenInfo holds metadata about a generated verification token
+type TokenInfo struct {
```
thomas-sickert: Just wondering if we need the whole `TokenInfo` data structure...

My line of thinking is that we don't need to decode the token since there's no data encoded, so we can do a string comparison from the data fetched via TXT record or .well-known and we don't operate on the byte array used to create the base64 encoded token, so the encoding information isn't necessary.

thomas-sickert (commented):
This is very thorough! I'm a bit concerned that it might be doing a lot of extra work though. 

Since these tokens are public and don't encode data, I think we might be adding a few extra steps and additional information that we don't need. I've left some comments below about it, but I'm thinking we could probably simplify this and reduce the amount of code we have to maintain.

Open to discussion and would love to hear your thoughts if you disagree 😄

internal/verification/token.go
aphansal123: That's a good point about the validation function - We could probably remove it if we're only comparing the expected token to the TXT/well-known record rather than validating the format.

internal/verification/token.go
aphansal123: Will we be mapping the token to the DNS TXT/well-known path in our database? If so then I think it's fine to remove this extra info.

internal/verification/README.md
```diff
@@ -0,0 +1,152 @@
+# Domain Verification Package
+
+This package provides cryptographically secure token generation for domain ownership verification in the MCP Registry. It implements the requirements specified in the Server Name Verification system.
+
+## Overview
+
+The verification package generates 128-bit cryptographically secure random tokens used for proving domain ownership through two verification methods:
+
+1. **DNS TXT Record Verification**: Add `mcp-verify=<token>` to your domain's DNS
+2. **HTTP-01 Web Challenge**: Serve the token at `https://domain/.well-known/mcp-challenge/<token>`
```
trent-j:
```suggestion
2. **HTTP-01 Web Challenge**: Serve the token at `https://domain/.well-known/mcp-verify`
```

trent-j (approved):
LGTM nice work!

internal/verification/token.go
```diff
@@ -0,0 +1,57 @@
...
+// suitable for domain ownership verification.
+const TokenLength = 16
+
+// GenerateVerificationToken generates a cryptographically secure 128-bit (16 bytes)
+// random token for domain ownership verification. The token is encoded using base64url
+// (RFC 4648) which is both URL-safe and DNS TXT record safe.
+//
+// This function is designed for use in both DNS TXT record verification
+// (mcp-verify=<token>) and HTTP-01 web challenge verification
+// (https://domain/.well-known/mcp-challenge/<token>).
```
thomas-sickert:
```suggestion
// (https://domain/.well-known/mcp-verify).
```

internal/verification/token.go
```diff
@@ -0,0 +1,57 @@
...
+// - An error if the system's entropy source is unavailable
+//
+// Example usage:
+//
+//	token, err := GenerateVerificationToken()
+//	if err != nil {
+//	    return fmt.Errorf("failed to generate verification token: %w", err)
+//	}
+//	// Use token in DNS: mcp-verify=<token>
+//	// Or HTTP: /.well-known/mcp-challenge/<token>
```
thomas-sickert:
```suggestion
//	// Or HTTP: /.well-known/mcp-verify
```

thomas-sickert (changes requested):
Nice! Found a few old strings for the mcp-verify path but other than that, this looks good 👍

Issue #181, opened 2025-07-13
Include mechanism for tagging servers as "archived"


tadasant:
For example, all the servers here: https://github.com/modelcontextprotocol/servers-archived

At some point, they would have been published to the registry as stable servers. Now, they need to be sunset in some way.

We'll probably want this concept baked into the top level `server.json` because this concept of "archived" is probably useful across `server.json` use cases.

rdimitrov:
This was one of the things we were about to bring up so happy to see that issue 😃 

Related to this is should this also be available for setting on per package/remote entry too? For example, the whole server might not be archived/deprecated but one of the ways it was packaged so far might have been. Is this a valid usecase or that is expected to be handled by just not including this package anymore in the new version bump?

tadasant:
@rdimitrov it's a good question - my gut is that your latter take is the right one:

> that is expected to be handled by just not including this package anymore in the new version bump?

It seems unnecessary to introduce the complexity of keeping an archived state per-package. It's necessary at the top server level because deleting at the server level would erase history - vs. we at the lower level we have the more elegant option to publish a new version without the archived package without losing any data. Let me know if you have a use case where that doesn't bear out elegantly.

rdimitrov:
Yeah, I think the same 👍 

The use cases that I can think of for having it are:
* This would allow a vendor to mark something as deprecated in order to notify people in advance before removing it completely.
* Should we support registry clients having reproducible environments? For example, if I set up an MCP server two months ago, could I recreate that setup today even if its package, remote, or the server itself was deprecated? Probably not as supporting this would increase memory usage significantly and deprecated servers were likely removed for good reasons.

That said maybe it's even worth considering to have something like a garbage collection that cleans all deprecated servers older than X time to keep the registry as fresh as possible.

tadasant:
Fair thoughts, though in the spirit of keeping it simple and not building ahead of need here, I think:

> This would allow a vendor to mark something as deprecated in order to notify people in advance before removing it completely.

The consumer could address this by just sticking to an older server version after the new version removes the old package

> Should we support registry clients having reproducible environments? For example, if I set up an MCP server two months ago, could I recreate that setup today even if its package, remote, or the server itself was deprecated? Probably not as supporting this would increase memory usage significantly and deprecated servers were likely removed for good reasons.

[...]

PR #210, opened 2025-07-30
Add a status field indicating if a server is deprecated or active

rdimitrov:
The following PR adds a new server `status` field to support the status of servers, like deprecated or active

Details:
* Add status field to server.json schema with 'active' and 'deprecated' values
* Update OpenAPI spec to include status field in Server responses
* Add ServerStatus enum and field to Go model structs
* Update examples to demonstrate status field usage (`active` and `deprecated`)

Since this is my first PR there's a good chance I might have missed something so I'll appreciate any feedback to make this and any PRs in the future easier for you to review 😃 

Fixes: #181

## Motivation and Context

The main goal is to have a way for server authors to notify registry consumers that a server may no longer be maintained and shouldn't be considered for using.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/model/model.go
+++ b/internal/model/model.go
@@ -10,6 +10,16 @@ const (
 	AuthMethodNone AuthMethod = "none"
 )
 
+// ServerStatus represents the lifecycle status of a server
+type ServerStatus string
+
+const (
+	// ServerStatusActive represents an active server recommended for use
+	ServerStatusActive ServerStatus = "active"
+	// ServerStatusDeprecated represents a server no longer recommended for new usage
+	ServerStatusDeprecated ServerStatus = "deprecated"
+)
+
 // Authentication holds information about the authentication method and credentials
 type Authentication struct {
 	Method  AuthMethod `json:"method,omitempty"`
@@ -114,6 +124,7 @@ type Server struct {
 	ID            string        `json:"id" bson:"id"`
 	Name          string        `json:"name" bson:"name"`
 	Description   string        `json:"description" bson:"description"`
+	Status        ServerStatus  `json:"status,omitempty" bson:"status,omitempty"`
 	Repository    Repository    `json:"repository" bson:"repository"`
 	VersionDetail VersionDetail `json:"version_detail" bson:"version_detail"`
 }
--- a/internal/model/model.go
+++ b/internal/model/model.go
@@ -14,9 +14,9 @@ const (
 type ServerStatus string
 
 const (
-	// ServerStatusActive represents an active server recommended for use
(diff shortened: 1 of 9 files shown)


internal/model/model.go
```diff
@@ -10,6 +10,16 @@ const (
...
 )
 
+// ServerStatus represents the lifecycle status of a server
+type ServerStatus string
+
+const (
+	// ServerStatusActive represents an active server recommended for use
+	ServerStatusActive ServerStatus = "active"
+	// ServerStatusDeprecated represents a server no longer recommended for new usage
+	ServerStatusDeprecated ServerStatus = "deprecated"
```
tadasant:
```suggestion
	// ServerStatusActive represents an actively maintained server (as asserted by the publisher)
	ServerStatusActive ServerStatus = "active"
	// ServerStatusDeprecated represents a server that is no longer actively maintained
	ServerStatusDeprecated ServerStatus = "deprecated"
```

tadasant (changes requested):
This looks great as far as the docs are concerned, thank you!

The one thing I think we'd want to include in this PR is integration with the publisher tool (tools/publisher). I'm not as familiar with that part of the codebase, but the basic idea is that we want to be able to use the `./bin/mcp-publisher create` command to generate a server.json that includes this new field

Ideally we'd have some sort of CI step that is asserting we don't have gaps like this, so next time someone makes a new contribution like this we don't risk missing that. Not asking you to include that enhancement for this PR, but maybe good for a followup?

rdimitrov:
> The one thing I think we'd want to include in this PR is integration with the publisher tool (tools/publisher). I'm not as familiar with that part of the codebase, but the basic idea is that we want to be able to use the `./bin/mcp-publisher create` command to generate a server.json that includes this new field

Thanks! Completely missed that part 😃 Added the necessary changes 👍 

> Ideally we'd have some sort of CI step that is asserting we don't have gaps like this, so next time someone makes a new contribution like this we don't risk missing that. Not asking you to include that enhancement for this PR, but maybe good for a followup?

Sounds good 👍 I guess one of the first things we should do is actually not try to replicate the same Go object twice but reuse a single one. What I mean is instead of using this -  https://github.com/modelcontextprotocol/registry/blob/651c8dfee340a8c04cb93bcb28e1fe3430809a09/tools/publisher/main.go#L55, we should be using - https://github.com/modelcontextprotocol/registry/blob/651c8dfee340a8c04cb93bcb28e1fe3430809a09/internal/model/model.go#L113

tadasant (approved):
This is great, thank you!

I think making that object DRY makes sense

PR #220, opened 2025-08-05
Verification API

thomas-sickert:
## Types of changes

- [ ] Documentation update

## Checklist


internal/service/fake_service.go
```diff
@@ -123,6 +124,62 @@ func (s *fakeRegistryService) Publish(serverDetail *model.ServerDetail) error {
 	return s.db.Publish(ctx, serverDetail)
 }
 
+// ClaimDomain generates a verification token for a domain and stores it as pending
+func (s *fakeRegistryService) ClaimDomain(domain string) (*model.VerificationToken, error) {
+	// Create a timeout context for the database operation
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	const maxAttempts = 10
```
aphansal123, line 133: Should we create a constant for this? I'm wondering if we may want to adjust this later.

aphansal123: Also I'm not sure if we need the retries in this case since we're using a 128-bit token, so I think the probability of a token collision occurring should be negligible (since there's 2^128 possible values), but I'm fine with keeping it.

internal/docs/swagger.yaml
```diff
@@ -200,6 +204,94 @@ paths:
...
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: Failed to claim domain
+
+  /v0/domains/status:
+    post:
```
aphansal123: I'm curious if this should be a GET operation instead of POST since we're checking the domain status?

internal/api/handlers/v0/verification.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// If no protocol, try adding one and parsing again
+	if !strings.Contains(domain, "://") {
+		if u, err := url.Parse("https://" + domain); err == nil && u.Host != "" {
+			return strings.ToLower(u.Host), nil
+		}
+	}
+
+	// If we get here, the input is not a valid domain/URL
+	return "", errors.New("invalid domain format")
```
aphansal123, line 35: Should we include the value of `err` in the error message here to give users more specific info on why the domain normalization fails?

aphansal123 (commented):
Looks great overall! I'm curious if we need a strategy to clean up old pending tokens that were never verified? This could prevent the database from growing indefinitely and the database queries from slowing down over time.

aphansal123 (approved):

domdomegg:
meta: would be fairly keen on getting this merged, then continuing to work on it - just to avoid large long running branches being hard to merge later, and clashing with things like #228

domdomegg:
@claude can you:
- address aphansal123's comment regarding changing the endpoint from get -> post, and push that to this branch
- add a github issue for cleaning up old tokens from the database

domdomegg:
@claude sorry i meant post -> get!

also regarding the gh access - any chance you can raise a PR to add this permission to the claude workflow or show me how, so you can do this in future?

domdomegg (approved):

PR #229, opened 2025-08-08
Clean up API documentation

domdomegg:
- Move API examples from docs into swagger.yaml
- Align openapi.yaml docs with implementation

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #236, opened 2025-08-08
Clean up API documentation

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to an unintended git history rewrite operation that broke the connection between branches. The operation has been reverted and I'm now recreating the affected PRs.

Original PR: #229

Clean up API documentation


domdomegg:
@claude review please

internal/docs/swagger.yaml
```diff
@@ -243,3 +348,266 @@ components:
...
+          properties:
+            packages:
+              type: array
+              items:
+                $ref: '#/components/schemas/Package'
+            remotes:
+              type: array
+              items:
+                $ref: '#/components/schemas/Remote'
+      examples:
```
crondinini-ant: I got an error when rendering this so I looked it up and it seems this only accepts a single "example".

domdomegg: Huh you're right, odd [from the spec](https://swagger.io/docs/specification/v3_0/adding-examples/) that this is an exception:

> Note that schemas and properties support single example but not multiple examples.

Will have a think about what we want to do here - because I think we do ideally want to have these examples...

domdomegg: I think these are all already present in examples.md, so I don't think we need them here too. Will simply delete :)

Thanks for the good catch!

camirondi (commented):
✅

domdomegg:
// approving based on @.camirondi's review - she's joining the MCP team but just hasn't got the GitHub perms quite set up yet :)

docs/server-registry-api/README.md
```diff
@@ -10,5 +10,4 @@ The centralized, publicly available catalog ("Official MCP Registry") needs addi
 
 References:
 - [openapi.yaml](./openapi.yaml) - A reusable API specification (MCP Server Registry API) that anyone building any sort of "MCP server registry" should consider adopting / aligning with.
-- [examples.md](./api_examples.md) - Example manifestations of the OpenAPI specification
-- [official-registry-openapi.ayml](./official-registry-openapi.yaml) - The specification backing the Official MCP Registry; a derivative of the MCP Server Registry API specification.
\ No newline at end of file
+- [swagger.yaml](../../internal/docs/swagger.yaml) - The specification backing the Official MCP Registry; a derivative of the MCP Server Registry API specification.
```
tadasant, line 13: @domdomegg do you have some replacement for `api_examples.md` we can reference instead? We've been referencing this file very heavily while iterating on the shapes - it's a lot easier to grok and talk about examples than it is to read the schemas.

domdomegg: A difficulty was they get out of date very quickly - https://registry.modelcontextprotocol.io/docs does have examples that people can poke around in.

And 90% of the complexity is in `server.json`, where I've kept the examples.

tadasant: > And 90% of the complexity is in server.json, where I've kept the examples.

Got it - wasn't sure if you were planning to do the same for those, but if we're keeping those then I'm ok dropping these for maintainability.

I'll also close #239 as I think this change supersedes that

domdomegg: Sounds good!

PR #221, opened 2025-08-05
Remote verification

thomas-sickert:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #238, opened 2025-08-08
Remote verification

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to an unintended git history rewrite operation that broke the connection between branches. The operation has been reverted and I'm now recreating the affected PRs.

Original PR: #221

Remote verification

### DIFF
--- a/cmd/registry/main.go
+++ b/cmd/registry/main.go
@@ -55,7 +55,7 @@ func main() {
 		defer cancel()
 
 		// Connect to MongoDB
-		db, err = database.NewMongoDB(ctx, cfg.DatabaseURL, cfg.DatabaseName, cfg.CollectionName)
+		db, err = database.NewMongoDB(ctx, cfg.DatabaseURL, cfg.DatabaseName, cfg.CollectionName, cfg.VerificationCollectionName)
 		if err != nil {
 			log.Printf("Failed to connect to MongoDB: %v", err)
 			return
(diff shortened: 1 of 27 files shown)


domdomegg:
@claude read this and PR #221, then:
- produce a concise, concrete explainer for what this PR does, and why / how it contributes to the registry project. don't assume readers have much context on the registry.
- explain the remaining blockers to merging this clearly. i think there's maybe some security things (not sure if false positive) and probably merge conflicts to resolve. if there are multiple ways to proceed, lay them out. if possible, try to break this down into distinct clear tasks.

domdomegg:
After digging in a bit more, I think this PR needs substantial work to get it into a good state. Things don't seem to be glued up correctly, and the code is starting to become quite messy and verbose for what it does.

I think key issues missing here are:

 1. No Authentication/Authorization
    - Tokens aren't associated with users (internal/api/handlers/v0/verification.go:94)
    - No way to enforce namespace ownership
    - Anyone who verifies a domain owns it forever
  2. No Verification Trigger
    - VerifyDNSRecord() exists but is never called
    - /v0/domains/status only checks database, doesn't perform DNS lookup
    - No endpoint to actually verify DNS records after claiming
  3. Incomplete Data Model
    - Missing user/publisher association in verification tokens
    - No link between verified domains and publishing permissions
    - No way to revoke or transfer domain ownership

domdomegg:
I think superceded by #279? Reopen if not!

PR #217, opened 2025-07-31
Fix struct duplication between internal/model and tools/publisher

rdimitrov:
## Motivation and Context

The following PR consolidates duplicate struct definitions between the publisher tool and the internal model package to eliminate code duplication and ensure consistency:
- Removed duplicate structs from `tools/publisher/main.go` (Repository, VersionDetail, Package, etc.)
- Updated publisher to import `github.com/modelcontextprotocol/registry/internal/model`
- Modified `createServerStructure()` to return `model.ServerDetail` and populate the full model structure
- The generated JSON now includes all model fields but only sets values for relevant ones. **Others like ID, ReleaseDate and IsLatest are left empty and so automatically set to their default values when marshalling.** I assume this shouldn't be a problem given these will be set/overwritten by the registry, but it's worth to note it.

**Example output:**

```bash
➜  publisher git:(use-model-types) ./bin/mcp-publisher create --name "io.github.example/simple-server" --description "A simple MCP server" --repo-url "https://github.com/example/simple-server" --output "simple.json"
2025/08/01 01:35:27 Successfully created simple.json
2025/08/01 01:35:27 You may need to edit the file to:
2025/08/01 01:35:27   - Add or modify package arguments
2025/08/01 01:35:27   - Set environment variable requirements
2025/08/01 01:35:27   - Add remote server configurations
2025/08/01 01:35:27   - Adjust runtime arguments
➜  publisher git:(use-model-types) cat simple.json 
{
  "id": "",
  "name": "io.github.example/simple-server",
  "description": "A simple MCP server",
  "status": "active",
  "repository": {
    "url": "https://github.com/example/simple-server",
    "source": "github",
    "id": ""
  },
  "version_detail": {
    "version": "1.0.0",
    "release_date": "",
    "is_latest": false
  },
  "packages": [
    {
      "registry_name": "npm",
      "name": "io.github.example/simple-server",
      "version": "1.0.0",
      "runtime_hint": "npx"
    }
  ]
}%   
```

```bash
➜  publisher git:(use-model-types) ./bin/mcp-publisher create --name "io.github.example/complex-server" --description "A complex MCP server with configuration" --repo-url "https://github.com/example/complex-server" --version "2.1.0" --status "deprecated" --execute "node server.js --port 8080 --config /etc/config.json" --env-var "API_TOKEN:Your API authentication token" --env-var "LOG_LEVEL:Logging level (debug, info, warn, error)" --package-arg "config-file:Path to configuration 
[...]

PR #240, opened 2025-08-08
Fix struct duplication between internal/model and tools/publisher

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to an unintended git history rewrite operation that broke the connection between branches. The operation has been reverted and I'm now recreating the affected PRs.\n\nOriginal PR: #217


rdimitrov:
@domdomegg - It's alright, no worries 🍻 I'll make sure to address your feedback and rebase it 👍

rdimitrov:
@domdomegg - hey, I don't have the rights to mark the PR out of draft and edit the PR description so I'll place it as a comment here -

## Motivation and Context

• Remove duplicated structs in `tools/publisher` and ensure the CLI uses the same schema as the registry.
• Reduce maintenance burden and keep publisher output aligned with the public API.

## What’s changed

• Switched publisher to import `github.com/modelcontextprotocol/registry/pkg/model` (no `internal/...` dependency).
• Removed local struct definitions in `tools/publisher/main.go`.
• `createServerStructure(...)` now returns `model.ServerDetail`; only necessary fields are populated.
• Output is minimal due to omitempty tags in pkg/model. **Note**: repository.id currently lacks omitempty, so it appears as an empty string.

### Example output

```json
{
  "name": "io.github.example/simple-server",
  "description": "A simple MCP server",
  "status": "active",
  "repository": {
    "url": "https://github.com/example/simple-server",
    "source": "github",
    "id": ""
  },
  "version_detail": {
    "version": "1.0.0"
  },
  "packages": [
    {
      "registry_name": "npm",
      "name": "io.github.example/simple-server",
      "version": "1.0.0",
      "runtime_hint": "npx"
    }
  ]
}
```

## How Has This Been Tested?

• Built the publisher and generated a sample server.json; verified unset fields are omitted where omitempty is present.

## Breaking Changes

• None.

## Types of changes

• Refactor / code hygiene.

## Related

• Add omitempty to repository.id or introduce a minimal DTO for publisher output to prevent accidental inclusion of registry-managed fields (already addressed via https://github.com/modelcontextprotocol/registry/pull/285)
• Continues the work from the closed PR #217

rdimitrov:
Regarding this https://github.com/modelcontextprotocol/registry/pull/217#pullrequestreview-3092393107 comment - I assume now that we have the `omitempty` tags set and they work as intended it shouldn't be needed to do the split you suggested, right?

domdomegg:
Yep I think this is correct! I think also because of the extensions spec that we've agreed on the model can be exactly the same between publisher and registry :)

rdimitrov:
@domdomegg - hey, fyi I'll work on rebasing this and do some housekeeping like better variable names this evening 👍 One of my goals is to have the server and the API objects public (as part of pkg/) and not be internal.

rdimitrov:
Closing in favour of #318 (@domdomegg can you help with that as GitHub doesn't let me do it)

PR #44, opened 2025-05-15
fix: ignore data directory in docker build

pboyd:
## Motivation and Context

I get this error when building the docker image after running the server in docker compose:

```
ERROR: failed to solve: error from sender: open data/diagnostic.data: permission denied
```

The issue is that `docker-compose.yaml` mounts the data directory in the mongo container which creates files that my user doesn't have access too. Except for seed.json, there doesn't appear to be a reason for those files to be in the docker build layer anyway.

This adds a `.dockeringore` file to exclude everything in data except `seed.json`.

## How Has This Been Tested?
`docker build`

## Breaking Changes
No

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #246, opened 2025-08-08
fix: ignore data directory in docker build

domdomegg:
I sincerely apologize for the disruption. This PR was accidentally closed due to an unintended git history rewrite operation that broke the connection between branches. The operation has been reverted and I'm now recreating the affected PRs.\n\nOriginal PR: #44

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+data
+!data/seed.json


domdomegg:
I think it's not clear that this is needed any more? If it is, please reopen!

Issue #97, opened 2025-05-27
Sanitization, limits on all free-form fields


tadasant:
Fields like `name, `description`, and anything else free-form in the `server.json` file should be strictly limited to minimum necessary content so as to limit the possible vectors for spam and/or security risks (e.g. injection attacks).

Avish34:
Hey @tadasant , Can I pick this up? Also, do we have a document capturing the constraint on fields like name, description etc.?

tadasant:
@Avish34 thank you! Yes happy for you to take this on. Some more guidance:
1. First step is to enumerate and add those constraints to the [registry-schema.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/registry-schema.json). We don't want them in the core [schema.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/schema.json) - the latter is fine to be more permissive.
2. Then it's to actually implement those enforcements in the Go code

Note that #198 is not yet done, so I expect you'll find inconsistencies in the implementation vs. the docs. The docs are the source of truth, so any work you do to get them closer to alignment is much appreciated.

> do we have a document capturing the constraint on fields like name, description etc.?

We don't have this, so part of the scope for addressing this issue is to review and make recommendations. Recommend just going through field by field and asking yourself what constraints might be reasonable to add, and document your thinking for each one.

tadasant:
@Avish34 this issue is specifically about "free-form fields". At the moment, that is:
- `Repository.url`
- `Repository.id`
- `VersionDetail.version` though there is an issue floating around to better define this, probably blocking
- `Server.name`
- `Server.description`
- `Package.name`
- `Package.version`
- `Input.*`... need to think through whether we can/should lock this one down at all
- `PositionalArgument.value_hint`
- `NamedArgument.name`
- `KeyValueInput.name`
- `Remote.url`

Basically we want to add guards for those in the same way we already added guards for `Repository.source`, `Package.registry_name`, `Package.runtime_hint`.

Some of these are obvious, e.g. URL's need to be valid URL's. `Repository.id` should be locked down to the format of GitHub's format (until we expand beyond GitHub being a Repositoryoption).

The goal here is to protect against spam/malicious activity that might try to inject data, by locking down the free form input to the things we expect to be there. It's ok if we don't have a mechanism for every single field; might not make sense to be foolproof here.

Avish34:
[...]

PR #272, opened 2025-08-14
Add formatting and validation for free-form fields

Avish34:
The changes are intended to create validators for all the free-form fields. Please find the details of the validation of different field below.

- Repository.Source: only github and gitlab is accepted.
- Repository.Url - It should be a valid github or gitlab URL. Created regex for both.
- Server.Name -  Max 200 char can be there. Reverse domain name verification is part of another PR (not owned by me).
- Package.Name - Max 200 char and no spaces.
- Server.Description - Max 100 char.
- Remote.Url - It should be valid URL.

TODOs
- Respository.ID - It's a int64 as mentioned in the github API docs. I will confirm again and make the relevant changes.

## Motivation and Context

The motivation was to make the code clean and add base code for more validator for different objects like Repository etc.
https://github.com/modelcontextprotocol/registry/issues/97

## How Has This Been Tested?
It's tested via UTs. Added UT's for all the validations.

## Breaking Changes
Yes

## Types of changes

- [ ] Documentation update

## Checklist


Avish34:
@tadasant please review.

domdomegg:
[EDIT: see comment below...]

I think where possible, I'd want to push this into the JSON schema spec as much as possible - this means:
- it's well documented to users of the spec
- we have fewer places we're defining the spec
- it's automatically validated by Huma (I think?)

Of course there will be exceptions to this where we really can't do it with JSON schema. But a lot of this I think could be replicated with e.g. length or regex requirements on the JSON schema etc.

domdomegg:
Hmm actually just read up on #97, and that maybe we only want this to be on the official registry API spec (rather than the generic registry API spec). Based on @.tadasant's comment:

> First step is to enumerate and add those constraints to the [registry-schema.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/registry-schema.json). We don't want them in the core [schema.json](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/schema.json) - the latter is fine to be more permissive.

I guess in that case the approach you've gone for here is fine - I'll leave it up to @tadasant to decide / review :)

rdimitrov:
@domdomegg - isn't it preferred to still have the validation in the schema even if it's just for registry-server.json?

domdomegg:
@rdimitrov possibly true? I guess then it'd be harder to take the official-registry-api-schema.json and turn it into the registry-api-schema.json. But maybe this is fine and/or we shouldn't maintain them this way.

Avish34:
@domdomegg / @tadasant can we conclude on the discussion whether we want to add the validation in the spec and let huma handle it or we should go ahead with the code based validation? It's fine if we need more time to think it through.

tadasant:
Sorry I'm well behind on reviews and discussions and trying to focus where I'm blocking @domdomegg - trust his call on this one if he wants to jump ahead and get this merged (otherwise I will get to this eventually, just will be slow)

This is correct:

> maybe we only want this to be on the official registry API spec

The intent is for this formatting/validation to only apply to the official registry. Most importantly at publication-time (we don't want people publishing data that is noncomformant to these expectations)

I haven't yet formed an opinion on how that cascades to the implementation here

Avish34:
No worries @tadasant, take your time. @domdomegg you can also review if you want to.

domdomegg:
Okay I've had a think about this, and my conclusion is that I think it doesn't matter too much either way 😅

I have a weak preference towards doing it as schema annotations in huma, rather than custom code validations. I think this documents it better for users of the API, and usually is less verbose/puts all the schema info in one place.

I'd be happy to accept this PR, or one doing it with schema annotations, once the code is cleaned up a little (e.g. linting and tests pass). @Avish34 are you good to take up implementing it one of these ways, then pinging me once you have CI passing?

Avish34:
Thanks @domdomegg , Yes I'm picking up the implementation. Sure!!

Avish34:
@domdomegg CI is passing now. Can you start reviewing please?

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// ServerValidator validates server details
+type ServerValidator struct {
+	*RespositoryValidator // Embedded RespositoryValidator for repository validation
+}
+
+// Validate checks if the server details are valid
+func (v *ServerValidator) Validate(obj *model.ServerDetail) error {
+	if obj.Name == "" {
```
domdomegg: I think this could be made a requirement on the main server.json schema (which should be pushed up into the huma validation)

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+		return ErrNameRequired
+	}
+
+	// Add format validation according to https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1086 in the next PR
+	if len(obj.Name) > MaxLengthForServerName {
+		return ErrServerNameTooLong
+	}
+
+	// Version is required
+	if obj.VersionDetail.Version == "" {
```
domdomegg: I think this could be made a requirement on the main server.json schema (which should be pushed up into the huma validation)

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+
+// RepositoryValidator validates repository details
+type RespositoryValidator struct {
+	validSources map[RepositorySource]bool
+}
+
+// Validate checks if the repository details are valid
+func (rv *RespositoryValidator) Validate(obj *model.Repository) error {
+	// empty repository object, this check is needed because we get empty repo object (if not present in request) everytime we unmarshal the request json
+	if len(obj.URL) == 0 && len(obj.Source) == 0 && len(obj.ID) == 0 {
```
domdomegg: nit: This feels potentially fragile - any way we can avoid this check? No worries if not.

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+func NewPackageValidator() *PackageValidator {
+	return &PackageValidator{}
+}
+
+// RemoteValidator validates remote connection details
+type RemoteValidator struct{}
+
+// Validate checks if the remote connection details are valid
+func (rv *RemoteValidator) Validate(obj *model.Remote) error {
+	if !IsValidURL(obj.URL) {
```
domdomegg, line 79: I think this could be made a requirement on the main server.json schema (which should be pushed up into the huma validation)

domdomegg (commented):
Looking mostly good, with a few comments - can merge after they are resolved :)

Avish34:
I will address the comments, thanks for reviewing @domdomegg. Btw is this https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/schema.json are you talking about in the comments right?

domdomegg:
@Avish34 yeah for the comments like 'I think this could be made a requirement on the main server.json schema' I think we'd want to add them to [the server.json schema](https://github.com/modelcontextprotocol/registry/blob/main/docs/server-json/schema.json) and as annotations on the Huma API probably [here](https://github.com/modelcontextprotocol/registry/blob/main/internal/model/model.go) following [these docs](https://huma.rocks/features/model-validation/).

Avish34:
@domdomegg Got it. You mean adding minLength and maxLength to the openapi schema of server.json for the relevant fields and adding the same annotation to model.go to allow huma to validate. Is the understanding right?

domdomegg:
Yes that's correct, and similar for other validations that should be universal to server.json (i.e. not specific to the official MCP registry)

internal/validators/validators.go
Avish34: Maybe we need to make Repository as pointer. I will take this up in next PR it that's fine.

Avish34:
@domdomegg  I have added the validation in the schema for server.json and annotations in model.go, I tried enabling huma validation but it's failing some UTs. I think enabling I can take up in next PR if not a problem, we can merge this till then.

internal/validators/constants.go
```diff
@@ -0,0 +1,38 @@
...
+	ErrPackageNameTooLong   = errors.New("package name is too long")
+	ErrPackageNameHasSpaces = errors.New("package name cannot contain spaces")
+
+	// Remote validation errors
+	ErrInvalidRemoteURL = errors.New("invalid remote URL")
+)
+
+// Constants for validation limits
+const (
+	MaxLengthForServerName  = 255
```
domdomegg: I think these should be aligned with the schema.json limits?

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Validate checks if the server details are valid
+func (v *ServerValidator) Validate(obj *model.ServerDetail) error {
+	if obj.Name == "" {
+		return ErrNameRequired
+	}
+
+	// Add format validation according to https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1086 in the next PR
+	if len(obj.Name) > MaxLengthForServerName {
```
domdomegg: these probably aren't needed if we already have huma validation on this

internal/validators/validators.go
domdomegg: same on this

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+		return ErrServerNameTooLong
+	}
+
+	// Version is required
+	if obj.VersionDetail.Version == "" {
+		return ErrVersionRequired
+	}
+
+	if len(obj.Description) > MaxLengthForDescription {
```
domdomegg: and this (sorry!)

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Validate checks if the repository details are valid
+func (rv *RespositoryValidator) Validate(obj *model.Repository) error {
+	// empty repository object, this check is needed because we get empty repo object (if not present in request) everytime we unmarshal the request json
+	if len(obj.URL) == 0 && len(obj.Source) == 0 && len(obj.ID) == 0 {
+		return nil
+	}
+	// validate the repository URL
+	repoSource := RepositorySource(obj.Source)
+	if _, ok := rv.validSources[repoSource]; !ok {
```
domdomegg: I think this if block isn't necessary, already validated in `IsValidRepositoryURL`

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+	return &RespositoryValidator{
+		validSources: map[RepositorySource]bool{SourceGitHub: true, SourceGitLab: true},
+	}
+}
+
+// PackageValidator validates package details
+type PackageValidator struct{}
+
+// Validate checks if the package details are valid
+func (pv *PackageValidator) Validate(obj *model.Package) error {
```
domdomegg, line 61: I think this isn't necessary, already validated in Huma

docs/server-json/server.schema.json
```diff
@@ -53,12 +53,16 @@
         "name": {
           "type": "string",
           "description": "Server name/identifier",
-          "example": "io.modelcontextprotocol/filesystem"
+          "example": "io.modelcontextprotocol/filesystem",
+          "minLength": 1,
+          "maxLength": 200
```
domdomegg, line 58: I was not sure if these limitations were to be applied to the server schema. Maybe okay to apply them here and see how people react - probably helpful the more things we can push into the server spec for ecosystem consistency.

internal/validators/validators.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	"fmt"
+
+	"github.com/modelcontextprotocol/registry/internal/model"
+)
+
+// ServerValidator validates server details
+type ServerValidator struct {
+	*RespositoryValidator // Embedded RespositoryValidator for repository validation
```
domdomegg: typo: RespositoryValidator -> RepositoryValidator

domdomegg (approved):

Avish34:
Thanks a lot @domdomegg !!

PR #276, opened 2025-08-15
feat: Add creds for GitHub auth, add "none" auth to publisher

domdomegg:
## Motivation and Context

- Adds credentials for the new MCP Registry Login GitHub Apps (local, staging, prod)
- Adds support for "none" auth to publisher

## How Has This Been Tested?

- Tested publisher workflows locally
- Unit and integration tests passing

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist


domdomegg:
@claude please review

.env.example
```diff
@@ -16,8 +16,8 @@ MCP_REGISTRY_SEED_IMPORT=true
 MCP_REGISTRY_VERSION=dev
 
 # GitHub OAuth configuration
-MCP_REGISTRY_GITHUB_CLIENT_ID=
-MCP_REGISTRY_GITHUB_CLIENT_SECRET=
+MCP_REGISTRY_GITHUB_CLIENT_ID=Iv23licy3GSiM9Km5jtd
+MCP_REGISTRY_GITHUB_CLIENT_SECRET=0e8db54879b02c29adef51795586f3c510a9341d
```
maxisbey, line 22: I'd prefer these to be left blank, and maintainers can create it themselves if needed for testing. I'm not super familiar with the best practice around these specific type of credentials, but general rule is to never commit secrets of any kind to a repo (even if for testing).

If this is normal, or it is best practice I'm keen to learn about that, but first look feels like we shouldn't have these.

domdomegg: [just noting down for sake of recording discussion we had in person]

These values are for the local / testing version of the GitHub App, which was intentionally created to be shared like this. These therefore aren't really "secrets", and are intentionally made public. It doesn't grant access to any resources, and is just used so people can get a GitHub token that they can fetch their own details with (e.g. what their username is + the organisations they are associated with).

Otherwise every contributor needs to create a GitHub App, which is a bit of a pain navigating the documentation etc. and configuring. And scaling this up to the other auth methods means actually they might need to set up lots of apps, which adds a greater barrier to entry to contributors.

This is similar to what other respected organisations do, e.g.:
- Stripe provides a public testing key in their docs: https://docs.stripe.com/api/authentication
- Supabase provides public testing keys in their docker setup: https://github.com/supabase/supabase/blob/master/docker/.env.example
- OrchardCMS has public testing stripe keys: https://github.com/OrchardCMS/OrchardCore.Commerce/blob/main/docs/features/stripe-payment.md

Note: the staging and prod credentials should be kept private, and are - the secrets there are only included in encrypted form.

---

Claude's (neutrally prompted) analysis:

> Yes, it's safe to include local GitHub app details in the codebase. Here's why:
> 
> 1. Separate local development app: The credentials in deploy/Pulumi.local.yaml (Iv23licy3GSiM9Km5jtd) are for a dedicated local development GitHub app, separate from staging (Iv23liB2r5oUw6PxiMOC) and
> production (Iv23liUydBbI7Z2Q9bOZ).
> 2. Open source project: Since the entire codebase is open source, having development credentials visible is acceptable and common practice.
> 3. Local environment only: The local credentials are only used for:
>     - docker-compose.yml (local development)
>     - .env.example (developer setup template)
>     - deploy/Pulumi.local.yaml (local deployment)
> 4. Production security maintained: Staging/prod credentials are properly encrypted using Pulumi's secret management (secure: fields).
> 5. Standard practice: Many open source projects include development/testing credentials in their repositories for easy contributor onboarding.
> 
> The PR correctly separates concerns - local development convenience vs production security.

tadasant: Maybe worth a comment in the code explaining this? Being that it caught the attention of both Max and I (in DMs), guessing other folks will see this and wonder the same 😅

PR #279, opened 2025-08-18
auth: Add DNS auth

### LINKED ISSUE #233: Clean up expired verification tokens from database
## Problem
The verification API creates pending tokens in the database when domains are claimed, but there's currently no mechanism to clean up old tokens that were never verified. This could lead to:

- Database growth over time
- Slower query performance as the tokens collection grows
- Potential confusion with multiple pending tokens for the same domain

## Proposed Solution
Implement a cleanup strategy for expired verification tokens. Options to consider:

1. **TTL-based cleanup**: Set a reasonable expiration time (e.g., 30 days) for pending tokens
2. **Periodic cleanup job**: Run a background task to remove old unverified tokens
3. **Database-level TTL**: Use MongoDB TTL indexes to automatically expire old tokens
4. **Cleanup on new claims**: When claiming a domain, remove any existing pending tokens first

## Implementation Considerations
- Should preserve verified tokens indefinitely 
- Need to decide on appropriate token lifetime (suggested: 30-90 days)
- Consider impact on existing tokens in production
- May want to add token creation timestamps if not already present

## Related
- Part of the domain verification API implementation in PR #220
- Suggested during code review by @aphansal123

### LINKED ISSUE #270: auth: DNS priv/pub key auth
Based on #264

## Current plan

We've thought a bit DNS verification. One idea we have is to have a separate DNS verification subsystem, where:
- the user has to authenticate e.g. via GitHub login
- they request a challenge token, which we generate and stores in a database
- the user then has to add a TXT record with this token
- the user then has to request a DNS refresh, where if we find the token we associate their github user with being able to publish to this domain
- we have to occasionally check the DNS still has this token

This requires users to have an existing auth provider account, causes problems with org permissions (e.g. do you authorise just that user, or all github users in their orgs (and which orgs!), to publish to the domain), a database table setup, and requires continual rechecking of DNS.

## Alternative proposal

What if instead of all that, we just had people generate a cryptographic keypair and upload the public key to A DNS txt record. Like:

```
example.com  TXT  "v=MCPv1; k=ed25519; p=MCowBQYDK2VwAyEAja8L6VAIuWaXJLxDXBFf..."
```

Then people can hit the registry auth API with:

```http
POST /v0/auth/dns
Content-Type: application/json

{
  "timestamp": "2025-05-16T18:56:49Z",
  "signedTimestamp": "<timestamp string signed with private key>"
}
```

(possibly instead there should be a challenge string generated by the registry that is signed, to reduce the risk of pre-computation attacks)

We verify that the timestamp was within the last 15 seconds, and verify the signature with the public key. Then this gets them a registry token authorising them to that domain's namespace.

This means:
- From the user perspective, it's simpler:
  - Just upload your key and use it straight away. No need to auth with a different method and bind your domain to that account.
  - They also don't need an existing auth provider, so it makes it more auth agnostic.
  - While generating a keypair is a little more complex, it matches other processes developers are used to, e.g. GitHub's SSH key registraton process, or DKIM signing. We could also build it into the publisher CLI.
- From the user perspective, it's more powerful:
  - Organisations can manage access to keys as they wish, including: granting access to teams, CI etc. and rotating keys immediately when people leave.
- From the registry perspective, it's simpler:
  - We don't have database state for DNS tokens, or looking up users <> domains.
  - We don't have to handle user/team management.
  - We don't have to do DNS rechecking because we do it at publish time.

domdomegg:
Fixes #270
Closes #233

Adds support for DNS-based authentication for proving ownership of a particular domain. E.g. if you can add a TXT record to `anthropic.com`, you can publish anything at `com.anthropic/*` or `com.anthropic.*`.

Notes:
- Uses Ed25519 signatures with 15-second timestamp windows to prevent replay attacks
- Supports multiple public keys in DNS TXT records for key rotation scenarios
- Publisher CLI supports DNS auth via --dns-domain and --dns-private-key flags
- Includes key generation commands in publisher README for easy setup

## Motivation and Context

This implements DNS-based public/private key authentication to allow people to actually publish at their own domains. Currently, authentication requires GitHub accounts which limits publishing to the io.github.* namespace.

DNS-based auth enables:
- Publishing under your actual domain (e.g., com.anthropic/* for anthropic.com)
- Independence from GitHub or other external auth providers
- Organizations managing their own keys and access to publishing

## How Has This Been Tested?

- Unit tests cover signature verification, DNS TXT parsing, timestamp validation, and error cases
- Tested locally by editing my own DNS records
- All existing auth methods covered by passing tests

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Will be adding HTTP challenge auth in a similar way as a separate PR, e.g. for verifying you own `github.com` by hosting a `github.com/.well-known/mcp-server-challenge` or similar. Just mentioning this as we discussed HTTP challenges as closely related to DNS challenges elsewhere, and want to reassure reviewer that this hasn't been forgotten with this approach. :) Update: tracking the HTTP auth in https://github.com/modelcontextprotocol/registry/issues/280

### DIFF
--- a/internal/api/handlers/v0/auth/dns.go
+++ b/internal/api/handlers/v0/auth/dns.go
@@ -196,7 +196,9 @@ func (h *DNSAuthHandler) buildPermissions(domain string) []auth.Permission {
 			Action:          auth.PermissionActionPublish,
 			ResourcePattern: fmt.Sprintf("%s/*", reverseDomain),
 		},
-		// Grant permissions for all subdomains (e.g., com.example.*)
+		// DNS implies a hierarchy where subdomains are treated as part of the parent domain,
+		// therefore we grant permissions for all subdomains (e.g., com.example.*)
+		// This is in line with other DNS-based authentication methods e.g. ACME DNS-01 challenges
 		{
 			Action:          auth.PermissionActionPublish,
 			ResourcePattern: fmt.Sprintf("%s.*", reverseDomain),


domdomegg:
@claude can you please review this PR?

jerome3o-anthropic (approved):

Issue #85, opened 2025-05-27
DNS-based server name verification


tadasant:
We should allow for MCP servers to be published without requiring them to be prefixed with a `io.github` name prefix (which is the default fallback).

Work to be done here:
- [ ] Decide on DNS TXT-based verification or some other process
- [ ] Implement as an alternative to the `io.github` publication approach
- [ ] Allow for an individual user to verify a domain on behalf of a whole GitHub org
- [ ] The DNS check should happen on _every_ publish, so that if the verification is no longer valid, the corresponding users/orgs are no longer allowed to publish

steiza:
I think there's quite a few open questions here.

As I understand it, the current authentication mechanism for publishing to the MCP registry is to have the user get a GitHub OAuth token, and ensuring the `server.json` `repository.url` matches that user. This works for MCP servers you run locally, but not for a remote server?

I think this proposal would allow for authorization for MCP servers that run on a remote server (but not locally, because there's no domain the submitter controls for a MCP server intended to run locally). Then you need to figure out a way to provide a challenge (to be set as the DNS TXT record on that domain), what properties that challenge should cover, what to do if multiple submissions use different paths on the same domain, etc.

There's a lot here - I think it might be helpful to write up a threat model, or at least some "happy path" and malicious user examples, and how the DNS flow enables the "happy path" users while preventing attacks from malicious users.

tadasant:
Sorry, I can tell from reading your response that we have not sufficiently documented this. Will take an action to do that better ASAP (https://github.com/modelcontextprotocol/registry/issues/131).

> As I understand it, the current authentication mechanism for publishing to the MCP registry is to have the user get a GitHub OAuth token, and ensuring the server.json repository.url matches that user. 

This is not the case. The goal of the auth mechanism (+ optional DNS verification) is to facilitate a namespace for the `name` field, in reverse-DNS fashion. So it is unrelated to the `repository.url` field. It may make sense to cut another issue for making sure the repo at `repository.url` wants to be associated with this MCP server entry, similar to https://github.com/modelcontextprotocol/registry/issues/96 for `packages`. And maybe another analogous mechanism for the entries in `remotes` (alluded to [here](https://github.com/modelcontextprotocol/registry/pull/130#issuecomment-2946397809); need to cut another ticket to follow up).

Hopefully from that starting point, the downstream considerations make more sense.

domdomegg:
Linking up a few related threads: #221, #270, #279

[...]

PR #282, opened 2025-08-19
auth: Add HTTP-based authentication support to validate domain ownership

### LINKED ISSUE #280: auth: HTTP priv/pub key auth
Add HTTP-based authentication to prove domain ownership, similar to #238's DNS authentication but using HTTPS well-known URIs instead of DNS TXT records.

### Motivation

Some organizations don't have easy access to DNS records, but can easily deploy web assets. They would therefore prefer to be able to complete a HTTP challenge rather than a DNS one to authenticate to the registry.

This would in some way mirror ACME challenges: where you can do a DNS-01 or a HTTP-01 challenge.

#### 1. Public Key Hosting

Domain owners host their Ed25519 public key at:
```
https://example.com/.well-known/mcp-registry-auth
```

(better names/ideas for paths welcome! although I don't think blocks implementation as is probably a one line change if we want to fix later before go-live)

Response format (plaintext, format to match the DNS records):
```txt
v=MCPv1; k=ed25519; p=MCowBQYDK2VwAyEAja8L6VAIuWaXJLxDXBFf...
```

Multiple keys can be specified by just having multiple lines

(Alternative considered:
- JSON format of the above e.g. `{ "v": "MCPv1", "keys": [ { "id": "key-2025-01", "algorithm": "ed25519", "publicKey": "MCowBQYDK2VwAyEAja8L6VAIuWaXJLxDXBFf..." } ] }`. But I think there's something nice about it being the same as DNS given how similar the two implementations will be.
- [JWKS format](https://datatracker.ietf.org/doc/html/rfc7517))

#### 2. Authentication Flow

Basically the same as DNS, except you only get access to the challenge domain (rather than the domain + its subdomains). This is to reflect the difference between DNS and HTTP, where DNS implies hierarchical control but HTTP does not. This mirrors similar protocols as well, e.g. like ACME challenges.

```http
POST /v0/auth/http
Content-Type: application/json

{
  "domain": "example.com",
  "timestamp": "2025-01-16T18:56:49Z",
  "signature": "<hex-encoded signature of timestamp>"
}
```

Registry verifies by:
1. Fetching `https://example.com/.well-known/mcp-registry`
2. Validating timestamp is within 15 seconds
3. Verifying signature with the specified public key
4. Returning JWT token with `com.example/*` permissions

#### 3. Publisher CLI Support

Would basically work same as DNS, but pass in `--http-domain` and `--http-private-key`. Docs should also provide an example of how to set it up, in the same way that we already do for DNS.

domdomegg:
Add HTTP-based authentication to prove domain ownership, similar to https://github.com/modelcontextprotocol/registry/pull/238's DNS authentication but using HTTPS well-known URIs instead of DNS TXT records.

Fixes #280

## Motivation and Context

Some organizations don't have easy access to DNS records, but can easily deploy web assets. They would therefore prefer to be able to complete a HTTP challenge rather than a DNS one to authenticate to the registry.

This would in some way mirror ACME challenges: where you can do a DNS-01 or a HTTP-01 challenge.

## How Has This Been Tested?

- Tested locally by stubbing out the HTTP getter
- Added unit tests

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

I concern I had when building this that it might expose us to SSRF attacks.

I think on reflection this is not a significant problem:
- We only GET a fixed file `.well-known/mcp-registry-auth` (redirects disallowed) after sanitizing the domain, and then also don't expose the response to the user. This means the user shouldn't be able to hit anything very interesting.
- The information they might get is what domains/servers exist in the network. Our infrastructure is already all open-source, so attackers aren't getting any additional real info by prodding at the environment this way.

### DIFF
--- a/internal/api/handlers/v0/auth/http.go
+++ b/internal/api/handlers/v0/auth/http.go
@@ -45,7 +45,7 @@ func NewDefaultHTTPKeyFetcher() *DefaultHTTPKeyFetcher {
 			Timeout: 10 * time.Second,
 			// Disable redirects for security purposes:
 			// Prevents people doing weird things like sending us to internal endpoints at different paths
-			CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			CheckRedirect: func(_ *http.Request, _ []*http.Request) error {
 				return http.ErrUseLastResponse
 			},
 		},
--- a/internal/api/handlers/v0/auth/http.go
+++ b/internal/api/handlers/v0/auth/http.go
@@ -74,6 +74,9 @@ func (f *DefaultHTTPKeyFetcher) FetchKey(ctx context.Context, domain string) (st
 		return "", fmt.Errorf("HTTP %d: failed to fetch key from %s", resp.StatusCode, url)
 	}
 
+	// Limit response size to prevent DoS attacks
+	resp.Body = http.MaxBytesReader(nil, resp.Body, 4096)
+
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return "", fmt.Errorf("failed to read response body: %w", err)
--- a/internal/api/handlers/v0/auth/http.go
+++ b/internal/api/handlers/v0/auth/http.go
@@ -43,6 +43,11 @@ func NewDefaultHTTPKeyFetcher() *DefaultHTTPKeyFetcher {
 	return &DefaultHTTPKeyFetcher{
 		client: &http.Client{
 			Timeout: 10 * time.Second,
+			// Disable redirects for security purposes:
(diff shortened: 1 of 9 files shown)


domdomegg:
@claude can you review this PR?

domdomegg:
Response to Claude review:
1. I think it's fine to reuse these functions as they're in the same package - Go seems to complain if I duplicate them as requested. Am not a Go expert so might be doing something dumb here, human pushback welcome!
2. Good idea, will implement

domdomegg:
// rebased on main, after squash merging #85

PR #290, opened 2025-08-20
102 Otel instrumentation

pree-dew:
## Motivation and Context
- It is important to have basic observability in place so that availability issues can be figured out easily

## How Has This Been Tested?
- Registration of metrics is tested
- Prometheus endpoint `/metrics` is included in test cases to test the metrics exposed
- Test locally to understand any regression.

## Breaking Changes
- No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- Telemetry implementation is currently done by keeping metrics in mind as of now, other types of telemetry can be added on top of it later.
- Instrumentation takes care of these problems already:
  - Cardinality control by limiting latency buckets and not including labels that grows with throughput.
  - Latency buckets are added as per SLA guidelines
  - Each metric prefixed with `mcp` to easily identify.
  - Pulling mechanism is used since the idea is to scrape metrics via an agent like vmagent, prometheus agent or otel-collector.
  - Metrics follow standard approach of defining counters, histogram and gauges as per opentelemetry guidelines.
  
Screenshot 2025-08-20 at 9 57 47 PM


internal/api/router/router.go
```diff
@@ -23,8 +104,16 @@ func NewHumaAPI(cfg *config.Config, registry service.RegistryService, mux *http.
 	// Create a new API using humago adapter for standard library
 	api := humago.New(mux, humaConfig)
 
+	// Add metrics middleware with options
+	api.UseMiddleware(MetricTelemetryMiddleware(metrics,
+		WithSkipPaths("/health", "/metrics", "/ping", "/docs"),
```
domdomegg, line 109: I'm not sure if we need to exclude the health and ping endpoints, seems find to collect metrics for them? Maybe fine for all of them really, even if it is a bit meta to collect metrics on /metrics haha.

Also if we do want this, are these the correct paths for health and ping - would it not be `/v0/health` and `/v0/ping`?

internal/api/handlers/v0/health.go
```diff
@@ -32,3 +39,15 @@ func RegisterHealthEndpoint(api huma.API, cfg *config.Config) {
 		}, nil
 	})
 }
+
+// recordHealthMetrics records the health check metrics
+func recordHealthMetrics(ctx context.Context, metrics *telemetry.Metrics, path string, version string) {
```
domdomegg, line 44: On reflection, I think maybe this up metric is not needed? I think prometheus will automatically generate an up metric of whether it can hit the metrics endpoint.

But I don't mind too much either way :)

domdomegg (approved):
This is great, thank you so much @pree-dew for the contribution.

We can discuss my minor suggestions here, and if we decide we do want to do something about them we can deal with them in a follow-up PR, so I'm happy to merge now.

pree-dew:
Thank you @domdomegg for allowing me to contribute, I am completely okay to discuss the changes you suggested in the follow-up PR as per your convenience.

Issue #260, opened 2025-08-13
feature: Support for DXTs


domdomegg:
Add support for DXTs (or [whatever they're going to be named](https://github.com/anthropics/dxt/issues/81)). We discussed this in a call last week, but we were thinking the format might be like this in a server.json:

```json
{
  ...
  "packages": [
    {
      "registry_name": "dxt",
      "name": "https://some-file-serving-place.com/my-dxt/1-0-2.dxt",
      "file_hashes": {
        "sha-256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce"
      },
      "version": "1.0.2"
    }
  ]
}
```

We could provide guidance on people on how to host it on GitHub releases etc.

The file hash is necessary so that clients can verify the file hasn't changed since e.g. a downstream repository has stamped it as 'verified' or 'probably free of malware' etc.

domdomegg:
cc @tadasant @felixrieseberg

tadasant:
Makes sense to add `dxt` as an option here, I like that elegant solution to the "what do we do with DXT" problem.

I could also see us moving to a similar approach for `container/oci` as per Jim's thoughts [here](https://github.com/modelcontextprotocol/registry/discussions/249#discussioncomment-14077383).

`file_hashes` is interesting but is this DXT-specific? Isn't this a concern for basically any `package` type? I think it's a good idea but maybe worth considering separately from DXT; I like that thought that downstream registries could verify it and it provides a way for folks to zip the data back together even if we're not storing that downstream data in the upstream directly.

tadasant:
Cross-pointing some thoughts from #249

> After thinking about this a little more, I'm actually not sure we'd want to make a significant change here. There are at least two reasons it is valuable to us to be explicit about `registry_name=dockerhub` rather than doing something generic like `package_uri=docker.io/mcp/filesystem`:
> - By including an explicit whitelist of well known registries, rather than an open-ended URI that could point to any (malicious) registry, we lean on existing registry standards for combatting spam and malicious code (it's one of our [design principles](https://github.com/modelcontextprotocol/registry/blob/main/docs/design_principles.md#4-meets-industry-security-standards)). So someone can't go and stand up maliciousregistry.org and start hosting packages they submit all over the MCP Registry and try to inject into MCP clients.
[...]

PR #295, opened 2025-08-21
feat: Add support for MCP Bundles (MCPB) in registry

joan-anthropic:
## Motivation and Context
- Add MCPB as a new registry_name option
- Add file_hashes field to package model for integrity verification
- Implement URL validation restricting MCPB packages to GitHub/GitLab hosts
- Require SHA-256 hash for all MCPB packages
- Add example MCPB package in documentation
- Update schemas and OpenAPI specification

See issue #260

## Breaking Changes
N/A, registry has not yet been launched

## Types of changes

- [ ] Documentation update

## Checklist


internal/service/registry_service.go
```diff
@@ -84,6 +87,51 @@ func (s *registryServiceImpl) GetByID(id string) (*model.ServerDetail, error) {
...
+// validateMCPBPackage validates MCPB packages to ensure they meet requirements
+func validateMCPBPackage(pkg *model.Package) error {
+	// Validate that the URL is from an allowlisted host
+	parsedURL, err := url.Parse(pkg.Name)
+	if err != nil {
+		return fmt.Errorf("invalid MCPB package URL: %w", err)
+	}
+
+	// Allowlist of trusted hosts for MCPB packages
+	allowedHosts := []string{
```
joan-anthropic, line 75: to verify

domdomegg: lgtm

internal/service/registry_service.go
```diff
@@ -84,6 +87,51 @@ func (s *registryServiceImpl) GetByID(id string) (*model.ServerDetail, error) {
...
+		"github.com",
+		"www.github.com",
+		"raw.githubusercontent.com",
```
domdomegg: possibly this one (raw.githubusercontent.com) is not necessary

domdomegg (approved):
LGTM!

Would like a review from one of @tadasant or @toby to verify they are happy with the direction and server.json spec implementation too :)

docs/server-json/examples.md
```diff
@@ -389,6 +389,38 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
...
+    "source": "github",
+    "id": "mcpb-123ab-cdef4-56789-012ghi-jklmnopqrstu"
+  },
+  "version_detail": {
+    "version": "1.0.2"
+  },
+  "packages": [
+    {
+      "registry_name": "mcpb",
+      "name": "https://github.com/modelcontextprotocol/text-editor-mcpb/releases/download/v1.0.2/text-editor.mcpb",
```
tadasant: This is right with our current schema. But a thought for potentially renaming some details here: `name` doesn't really make sense for a URL like this, and `registry_name` doesn't for `mcpb` either (it's more of a package `type`).

I think I'd be in favor of something more like this here:

```json
{
      "location": {
        "type": "mcpb",
        "url": "https://github.com/modelcontextprotocol/text-editor-mcpb/releases/download/v1.0.2/text-editor.mcpb"
      },
      "runtime_hint": ...
      ...
}
```

And that's it - drop `version`, because the only reason we have `version` is to effectively "construct" the `url` in the `registry_name` case.

Basically, my proposal is to add a nested `PackageLocation` object inside here, where we currently just store [package_name, name, and version](https://github.com/modelcontextprotocol/registry/blob/ebaf4c9c99a13f214c3f72ebeed7c1cc0947033f/docs/server-json/schema.json#L76-L98) in the top level of the `Package` object. It would be an "OR" situation where either the original fields (registry_name, name, and version) or the new fields (type, url) are required. `mcpb` is the only valid entry for `type`.

Then the original case for other examples, like the one just before this, becomes:

```json
  "packages": [
    {
      "location": {
        "registry_name": "npm",
        "name": "@example/hybrid-mcp-server",
        "version": "1.5.0"
      },
      "runtime_hint": "npx",
      "package_arguments": [
        {
          "type": "named",
          "name": "--mode",
          "description": "Operation mode",
          "default": "local",
          "choices": ["local", "cached", "proxy"]
        }
      ]
    }
  ]
```

It's a breaking change but I think at this point, pre-release, is a good time to make this rather than shoehorn values into `registry_name` and `name`. What do you all think? cc @domdomegg @joan-anthropic @toby

docs/server-json/examples.md
```diff
@@ -389,6 +389,38 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
...
+      "version": "1.0.2",
+      "file_hashes": {
+        "sha-256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce"
+      }
```
tadasant: Should we consider lifting this inside the `PackageLocation` object I am proposing above, because it is only a concern relevant to non-package-registry entries?

tadasant (commented):
I'm a little uneasy about shoehorning `registry_name` and `name` to this purpose, while we still have the opportunity to make pre-release breaking changes. Thoughts on the proposed clearer modeling?

docs/server-json/examples.md
tadasant: I guess there is the possibility we add some registry in the future that does not do its own checks. So I'm fine keeping it here as an optional field.

I assume we need to rely on the MCP Registry to generate this file hash at publish-time, right? Or is the idea it will get generated by the CLI tool en route to creating a server.json, and the registry just validates it's correct and not tampered with in the publish flow?

I think we should map out and document this feature in a little more detail, maybe a dedicated doc in `docs/` explaining the flows, who's responsible for what. I worry a bit about how `file_hashes` is something we are expecting in the immutable server.json file, but also need to verify they're actually valid hashes. Seems doable but would like to see a step by step explanation somewhere.

docs/server-json/examples.md
joan-anthropic: @tadasant thanks for the quick turnaround. Agree that `registry_name` and `name` feel a bit overloaded in the current proposal - I like the distinction.

It'd be nice to keep version as a separate field - to better enable future auto-update workflows for DXTs. Thoughts?

docs/server-json/examples.md
joan-anthropic: I'm not opinionated here and will defer to you all re: what makes sense for the workflow the registry committee would prefer to support; happy add docs to reflect.

tadasant:
@joan-anthropic another thought just came to me (sorry for the thrashing): what if instead of `PackageLocation` we just unified the registry-based location and the arbitrary file-upload based one, but rather than shoehorning into `registry_name` and `name`, we refactored those to be agnostic of the registries?

For example, this old format:

```json
    {
      "registry_name": "npm",
      "name": "@example/hybrid-mcp-server",
      "version": "1.5.0",
      "runtime_hint": "npx",
      "package_arguments": [
        {
          "type": "named",
          "name": "--mode",
          "description": "Operation mode",
          "default": "local",
          "choices": ["local", "cached", "proxy"]
        }
      ]
    }
```

Becomes:

```json
    {
      "url": "https://www.npmjs.com/package/@example/hybrid-mcp-server/v/1.5.0",
      "type": "javascript",
      "runtime_hint": "npx",
      "package_arguments": [
        {
          "type": "named",
          "name": "--mode",
          "description": "Operation mode",
          "default": "local",
          "choices": ["local", "cached", "proxy"]
        }
      ]
    }
```

So then we have just one way of doing things.

We can still properly sanitize and logic branch on `npmjs.com` as if `registry_name=npm` in everything we care about downstream.

Running out of time here today on my end to fully think through this, but this seems like a potentially cleaner approach here and drawbacks aren't coming to top of mind.

I also like how it helps us remove the schema-level dependency on enumerating all these branded centralized registry names `npm`, `pypi`; that just becomes a in-code validation-level concern.

Curious for @domdomegg's take here

docs/server-json/examples.md
tadasant: Not sure I follow the use case, could you explain auto-update workflows for DXTs? Is there a reason you wouldn't be able to use the `https://github.com/modelcontextprotocol/text-editor-mcpb/releases/download/v1.0.2/text-editor.mcpb` value for that (which has the version in it)? 

We've had some feedback from @toby that all these `version` fields throughout the various server.json levels are causing confusion, so my thinking is that this is a good opportunity to potentially eliminate one of them as it's not actually necessary (but maybe I'm missing something).

domdomegg: - Maybe package_type and package_identifier seems good? From https://github.com/modelcontextprotocol/registry/discussions/249#discussioncomment-14118379. I think also as a meta thing I'd prefer them to stay in the package object rather than a nested location object, because I think unnecessary nesting is painful for many ecosystems (e.g. parsing nested structs like this in Java creates a lot of boilerplate).

So together I think I might propose something like:

```jsonc
"packages": [
    {
      "package_type": "mcpb",
      "package_identifier": "https://github.com/modelcontextprotocol/cool-mcp/releases/download/v1.2.3/cool-mcp.mcpb",
      // there's something that aesthetically feels a little worse about this not being an object, but I think it will make it easier on consumers. and very unlikely to really need other hash functions unless sha256 is broken
      "package_hash_sha256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce"
    },
    {
      "package_type": "docker",
      // docker implicitly has registry name in package_identifier, I think we should just have package_identifier because of this
      "package_identifier": "ghcr.io/domdomegg/cool-mcp:1.2.3@sha256:4a7f307d128da53e99ea7f45064df440c742aa3f648a09a09b51dfd1cd55378b"
    },
    {
      "package_type": "commonjs",
      // i think for now only allowlist npmjs. but subregistries could use e.g. https://npm.pkg.github.com or https://artifactory.corp.internal/npm
      "registry_name": "https://registry.npmjs.org/",
      "package_identifier": "@domdomegg/cool-mcp@1.2.3",
      "runtime_hint": "npx",
      "package_arguments": [
        {
          "type": "named",
          "name": "--mode",
          "description": "Operation mode",
          "default": "local",
          "choices": ["local", "cached", "proxy"]
        }
      ]
    }
  ]
```

- I think I agree with @tadasant that the version is not necessary in the `packages` array for DXT. There will still be a version at the top level in server.json. For auto-update I think we'd use this top-level version, and maybe `is_latest`.

domdomegg:
> what if instead of PackageLocation we just unified the registry-based location and the arbitrary file-upload based one, but rather than shoehorning into registry_name and name, we refactored those to be agnostic of the registries?

Yep I like this! I ended up [commenting something similar](https://github.com/modelcontextprotocol/registry/pull/295#discussion_r2295239560) before I scrolled far enough down to see this.

I think I prefer the direction of this approach. I don't mind _too much_ between a URL and slightly more structure (e.g. package type, registry name, identifier). I think I have a weak preference towards the slightly more structure just because I think clients might have an easier time parsing it to programmatically get packages etc. But fine to go with URL, especially if we provide guidance on the formats of URLs. Perhaps also some small SDKs in common languages for parsing and running server.jsons, as I imagine clients will want something like that anyway.

joan-anthropic:
@domdomegg I took @tadasant 's suggestion here but happy to replace with `package_type`, `registry_name` etc if preferred - still trying to figure out test/lint failures but just want to make sure the broad structure looks good to you

joan-anthropic:
Sorry - I'm missing a few things here; please hold off on review!

joan-anthropic:
@domdomegg would welcome a review here - I'm not totally sure how these changes ought to fit into the migration script and leaning heavily on Claude there.

Re: schema changes, I try to reflect both goals of you and Tadas - providing some structure to ease downstream parsing, but preferring a flattened structure that can be somewhat registry-agnostic.

```

  Before:
  {
    "registry_name": "npm",
    "name": "@21st-dev/magic"
  }

  After:
  {
    "package_type": "javascript",
    "registry_name": "npm",
    "identifier": "@21st-dev/magic"
  }

  {
    "package_type": "mcpb",
    "registry_name": "github-releases",
    "identifier": "https://github.com/example/mcp-bundle/releases/download/v1.0.0/bundle.mcpb"
  }


```
Changes:
  1. Added package_type field - Explicitly identifies the package ecosystem (javascript, python, docker, dotnet, mcpb, binary)
  2. Renamed `name` to `identifier` - More clearly indicates this is the package identifier within its registry
  3. Kept `registry_name` - Specifies which registry hosts the package (npm, pypi, docker-hub, nuget, github-releases, etc.) - would love a sense-check here whether eg github-releases is reasonable package ID for mcpb, else can remove

joan-anthropic:
Also, I remove the file hash generation / validation workflow from this PR and would love to treat that as a follow-up

domdomegg:
Don't worry about migrate-seed! It's just a temporary script we made for ourselves for doing a bit of work with the seed file previously, can (should?) probably be binned.

Regarding field naming I have thought a bit more and maybe have a slight minor preference to registry_type. This is because clients will want to download the packages from somewhere, so need to know _how_ (registry_type) and _where_ (registry_name / registry_identifier). I think generally registry_type (e.g. npm) implies package_type (e.g. javascript), so we don't need the latter.

This is (sadly) needed because we could actually have multiple registry_names per programming language. E.g. `npm` or `jsr` for javascript, or `pypi` vs `conda` in python land.

(a confusing point is that often the dominant registry API type mirrors the name of the official registry in that world, e.g. the NPM spec vs the NPM registry, and same for PyPi, etc.)

And finally for registry name to avoid ambiguity maybe it'd be best to specify the base url, especially as this makes it easier to swap out for custom ones etc. in subregistries.

---

I'm happy to implement the changes above either on this branch, or another branch leading into this one, if that would make it clearer?

domdomegg:
Thoughts about alts to registry_name, as actually this is a bit weird after staring at the examples when updating them on a branch locally:

  1. registry_name - This does sound like it should be a name rather than a URL, so I can see why they're concerned
  2. registry_identifier - This could be confusing as it might sound like it's identifying the package on the registry rather than the registry itself
  3. Other options could be:
    - registry_url - Most explicit about what it contains
    - registry_base_url - Very clear but a bit verbose
    - registry_endpoint - Technical but clear
    - registry_host - Simple and clear

I think I am most keen on registry_base_url. That would be:

  - npm → https://registry.npmjs.org
  - pypi → https://pypi.org
  - docker-hub → https://registry-1.docker.io
  - nuget → https://api.nuget.org
  - github-releases → https://github.com
  - gitlab-releases → https://gitlab.com

domdomegg:
have raised https://github.com/joan-anthropic/registry/pull/1

domdomegg (approved):

docs/server-json/examples.md
```diff
@@ -128,8 +131,9 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
         ]
       },
       {
-        "registry_name": "docker",
-        "name": "mcp/filesystem",
+        "registry_type": "docker-hub",
+        "registry_base_url": "https://docker.io",
+        "identifier": "mcp/filesystem:1.0.2",
```
tadasant, line 136: @joan-anthropic @domdomegg it seems odd to me to have 1.0.2 both in the identifier here and below in `version`. How should consumers decide which version to use (e.g. in the case they are submitted as different)?

Should we collapse `version` to be part of `identifier` and get rid of `version` as a key? Or just not allow version data in `identifier`?

docs/server-json/examples.md
tadasant: @joan-anthropic @domdomegg did we work out who is responsible for what in this file hash verification flow somewhere? Sorry if I missed it, still working through catching up on emails and messages

docs/server-json/examples.md
domdomegg: Yeah, this isn't really supposed to be here. Have removed from identifier in #332

There were some discussions about how best to identify docker images in https://github.com/modelcontextprotocol/registry/discussions/249#discussioncomment-14259927

docs/server-json/examples.md
joan-anthropic: @tadasant apologies for the delay - I'm OOO through the rest of the week so might be a bit slow to respond.

I ultimately wanted to leave the hash validation workflow as a question for you / @domdomegg / @toby as it'll be supported by the registry committee (iiuc). I'm happy to open a follow-up PR to reflect whatever decision ends up being made, though let me know if you're un-opinionated and I'm happy take a stab at it.

Just so I understand: do we actually _need_ to generate/verify the hash at publish time? My impression was that it was a guard against _future_ tampering with mcpb contents - so could we trust that the publication is correct and have registry clients take responsibility for hash verification before they install the mcpb? Ie, make this an contract between mcpb author <> registry client, and not MCP registry <> registry client?

tadasant: Adding a ticket to think through this to make sure we tackle before go-live (apologies if I'm still missing context somewhere, haven't given this field much thought yet): https://github.com/modelcontextprotocol/registry/issues/351

PR #296, opened 2025-08-22
feat: implement server versioning approach with semantic versioning support

### LINKED ISSUE #158: Decide & document server versioning approach for server.json

domdomegg:
This implements the versioning approach agreed upon in issue #158:

### Documentation
- Added comprehensive versioning guide (docs/versioning.md)
- Updated FAQ with versioning guidance and best practices 
- Enhanced schema description with SHOULD recommendations

### Schema Updates
- Added 255-character limit for version strings in schema.json
- Enhanced version field description with semantic versioning guidance

### Implementation 
- New isSemanticVersion() function for proper semver detection
- Enhanced compareSemanticVersions() with prerelease support
- Implemented compareVersions() strategy:
  * Both semver: use semantic comparison
  * Neither semver: use timestamp comparison
  * Mixed: semver versions always take precedence
- Updated publish logic to determine "latest" using new strategy
- Added version length validation (255 char max)

### Versioning Strategy
1. Version MUST be string up to 255 characters
2. SHOULD use semantic versioning for predictable ordering
3. SHOULD align with package versions to reduce confusion 
4. MAY use prerelease labels for registry-specific versions
5. Registry attempts semver parsing, falls back to timestamp ordering
6. Clients SHOULD follow same comparison logic

Resolves #158

Generated with [Claude Code](https://claude.ai/code)


domdomegg:
@claude I think the logic for ordering etc. shouldn't happen in database/memory.go - it should happen a layer higher at the service layer, not persistence layer.

docs/versioning.md
```diff
@@ -0,0 +1,137 @@
...
+## Version Ordering and "Latest" Determination
+
+### For Semantic Versions
+The registry attempts to parse versions as semantic versions. If successful, it uses semantic version comparison rules to determine:
+- Version ordering in lists
+- Which version is marked as `is_latest`
+
+### For Non-Semantic Versions
+If version parsing as semantic version fails:
+- The registry will always mark the version as latest (overriding any previous version)
```
domdomegg, line 71: This feels reasonable in isolation. but actually means that `version`'s total ordering is inconsistent with `is_latest`, for packages with a mix of versions.

Not sure what is best to do here. Maybe this is fine. Maybe we should only update is_latest if it's now the latest in the total ordering spec below, idk.

docs/versioning.md
```diff
@@ -0,0 +1,137 @@
...
+"2021.03.15"     // Date-based versioning
+"snapshot"       // Development snapshots
+"latest"         // Custom versioning scheme
+```
+
+### Alignment Examples
+```json
+{
+  "version_detail": {
+    "version": "1.2.3-gke.1"
```
domdomegg:
```suggestion
    "version": "1.2.3-1"
```

internal/service/registry_service.go
```diff
@@ -84,6 +84,62 @@ func (s *registryServiceImpl) GetByID(id string) (*model.ServerDetail, error) {
 	return serverDetail, nil
 }
 
+// getVersionsByName retrieves all versions of a server by name
+func (s *registryServiceImpl) getVersionsByName(ctx context.Context, name string) ([]*model.ServerDetail, error) {
+	// Use the database's List method to find all servers with the same name
+	entries, _, err := s.db.List(ctx, map[string]any{"name": name}, "", 1000) // Large limit to get all versions
```
domdomegg: fixed limit may cause problems, although no limit may cause dos. think this is okay for now.

internal/service/versioning.go
```diff
@@ -0,0 +1,125 @@
+package service
```
domdomegg, line 1: should have unit tests

docs/versioning.md
jonathanhefner: I think this behavior is necessary because the publisher does not indicate whether they are actually using semantic versioning.  For example, if we said "semantic versions always take precedence", and a publisher is using non-semantic version numbers that sometimes appear to be semantic, then they are going to have a bad time. :laughing:

Unfortunately, as you noted, this means there is no natural sort order.  Instead, we have to maintain an insertion order based on this logic.

docs/versioning.md
```diff
@@ -0,0 +1,137 @@
...
+
+### For Non-Semantic Versions
+If version parsing as semantic version fails:
+- The registry will always mark the version as latest (overriding any previous version)
+- Clients should fall back to using publish timestamp for ordering
+
+### Mixed Scenarios
+When comparing versions where one is semantic and one is not:
+- The semantic version is always considered higher
+- This ensures semantic versions take precedence in ordering
```
jonathanhefner: I think this conflicts with the above section ("If version parsing as semantic version fails"), and I think it may potentially result in undesirable behavior (see my comment on the above section).

docs/versioning.md
```diff
@@ -0,0 +1,137 @@
...
+1. Attempt to interpret versions as semantic versions when possible
+2. Use the following ordering rules:
+   - If both versions are valid semver: use semver comparison
+   - If neither are valid semver: use publish timestamp  
+   - If one is semver and one is not: semver version is considered higher
+
+## Examples
+
+### Valid Semantic Versions
+```json
```
jonathanhefner:
``````suggestion
```javascript
``````

File: docs/versioning.md [hunk 7]
```diff
@@ -0,0 +1,137 @@
...
+"1.0.0"          // Basic semantic version
+"2.1.3-alpha"    // Prerelease version  
+"1.0.0-beta.1"   // Prerelease with numeric suffix
+"3.0.0-rc.2"     // Release candidate
+```
+
+### Non-Semantic Versions (Allowed)
+```json
```
jonathanhefner:
``````suggestion
```javascript
``````

**Comment by domdomegg** [comment]
Requesting review from @rdimitrov :)

**Comment by domdomegg** [comment]
@claude can you review this?

**Comment by domdomegg** [comment]
Response to Claude review:

1. is wrong, that is a semver
2. accepted for now
3. will fix!
4. #272
5. this is fine
6. accepted

File: internal/service/versioning.go [hunk 8]
```diff
@@ -0,0 +1,180 @@
...
```
rdimitrov, line 1: Shouldn't we consider using semver instead? It's from the official package - https://pkg.go.dev/golang.org/x/mod/semver and would save us from maintaining/testing that part

internal/service/registry_service.go
```diff
@@ -81,11 +83,59 @@ func (s *registryServiceImpl) Publish(req model.PublishRequest) (*model.ServerRe
...
+	
+	// Check all existing versions to determine if new version should be latest
+	for _, server := range existingServers {
+		if server.RegistryMetadata.IsLatest {
+			existingLatestID = server.RegistryMetadata.ID
+			existingVersion := server.ServerJSON.VersionDetail.Version
+			existingTime, _ := time.Parse(time.RFC3339, server.RegistryMetadata.ReleaseDate)
+			
+			// Compare versions using the proper versioning strategy
+			comparison := CompareVersions(newVersion, existingVersion, currentTime, existingTime)
```
rdimitrov, line 121: Should we exit early here in case the request is trying to publish a version that is already there (which I think we don't want to allow) or older?

rdimitrov (commented):
Overall it looks good 👍 I added a few comments but my main one is around the custom semver implementation (less code == less problems 😃)

internal/service/registry_service.go
domdomegg: Good point! I think publishing old versions is fine, but publishing duplicate version is definitely a no-no. Fixing...

internal/service/versioning.go
domdomegg: Yeah, that is probably better.

There are some Anthropic-internal reasons why I didn't do this initially, but realistically this is definitely the better approach.

PR #306, opened 2025-08-26
feat: implement DNS namespace validation and remote URL uniqueness checks

### LINKED ISSUE #180: Enforce requirement: remote URL's in official registry must match reverse-DNS namespace
This means that if you have an MCP server for which you want to declare a `remotes` entry at e.g. `subdomain.microsoft.com/mcp`, your MCP server _must_ be in the namespace `com.microsoft.subdomain`.

Secondarily, this requirement enables enforcement that if someone else has published a remote server at that URL, you cannot also do so (no duplicates). This prevents spam attempts to list the same server many times.

domdomegg:
## Summary

- Add reverse-DNS namespace format validation for server names (namespace/name)
- Validate remote URLs match the DNS namespace derived from hostname  
- Prevent duplicate remote URLs across different servers
- Support localhost/development URLs without namespace restrictions
- Add comprehensive test coverage for validation logic
- Update database layer to support remote_url filtering

Fixes #180 by ensuring proper namespace governance and preventing URL conflicts.

### DIFF
--- a/internal/model/model.go
+++ b/internal/model/model.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 )
@@ -302,27 +303,6 @@ func validateRemoteURLMatchesNamespace(remoteURL, namespace string) error {
 	return nil
 }
 
-// convertHostnameToReverseDNS converts a hostname to reverse-DNS format
-func convertHostnameToReverseDNS(hostname string) string {
-	// Skip localhost and local development URLs
-	if hostname == "localhost" || strings.HasSuffix(hostname, ".localhost") || hostname == "127.0.0.1" {
-		return ""
-	}
-
-	parts := strings.Split(hostname, ".")
-	if len(parts) < 2 {
-		// For single-part hostnames, return empty to skip validation
-		return ""
-	}
-
-	// Reverse the parts
-	for i, j := 0, len(parts)-1; i < j; i, j = i+1, j-1 {
-		parts[i], parts[j] = parts[j], parts[i]
-	}
-
-	return strings.Join(parts, ".")
-}
-
 // extractPublisherDomainFromNamespace converts reverse-DNS namespace to normal domain format
 // e.g., "com.example" -> "example.com"
 func extractPublisherDomainFromNamespace(namespace string) string {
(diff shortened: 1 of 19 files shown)


domdomegg:
@claude please review

internal/model/model.go
```diff
@@ -244,22 +243,137 @@ func ParseServerName(serverDetail ServerDetail) (string, error) {
...
+	// Check if the remote URL hostname matches the publisher domain or is a subdomain
+	if !isValidHostForDomain(hostname, publisherDomain) {
+		return fmt.Errorf("remote URL host %s does not match publisher domain %s", hostname, publisherDomain)
+	}
+
+	return nil
+}
+
+// convertHostnameToReverseDNS converts a hostname to reverse-DNS format
+func convertHostnameToReverseDNS(hostname string) string {
```
crondinini-ant: Is this for the future / used anywhere? Maybe I'm just missing it but I can't find the usage other than in the test.

internal/model/model.go
```diff
@@ -244,22 +243,137 @@ func ParseServerName(serverDetail ServerDetail) (string, error) {
...
+	// Split into parts and reverse them to get normal domain format
+	parts := strings.Split(namespacePart, ".")
+	if len(parts) < 2 {
+		return ""
+	}
+
+	// Reverse the parts to convert from reverse-DNS to normal domain
+	for i, j := 0, len(parts)-1; i < j; i, j = i+1, j-1 {
+		parts[i], parts[j] = parts[j], parts[i]
+	}
```
crondinini-ant: Nit: could also use slices.Reverse(parts)

internal/service/registry_service.go
```diff
@@ -154,3 +165,28 @@ func (s *registryServiceImpl) Publish(req model.PublishRequest) (*model.ServerRe
...
 	return &response, nil
 }
+
+// validateNoDuplicateRemoteURLs checks that no other server is using the same remote URLs
+func (s *registryServiceImpl) validateNoDuplicateRemoteURLs(ctx context.Context, serverDetail model.ServerDetail) error {
+	// Check each remote URL in the new server for conflicts
+	for _, remote := range serverDetail.Remotes {
+		// Use filter to find servers with this remote URL
+		filter := map[string]any{
+			"remote_url": remote.URL,
```
crondinini-ant, line 175: Would it make sense to add name to the filter here?
```suggestion
			"remote_url": remote.URL,
			"name": serverDetail.Name
```

Just trying to understand if we can move the "Check if any conflicting server has a different name" to the db query level and if it makes sense.

domdomegg: I think it's not possible (with our current filtering) to filter for the name _not_ being a particular value unfortunately.

PR #307, opened 2025-08-26
Disable ireturn linter rule

domdomegg:
## Summary

- Disable the `ireturn` linter rule in golangci-lint configuration
- Remove all related `nolint:ireturn` comments from the codebase

The `ireturn` rule was flagging factory functions that intentionally return interfaces for dependency injection, which is a common and valid Go pattern. Since this pattern is used extensively throughout the codebase for good architectural reasons, it makes sense to disable this rule entirely.

References: https://stackoverflow.com/questions/72385337/golang-linter-always-complains

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,7 +27,6 @@ linters:
     - gosec
     - grouper
     - importas
-    - ireturn
     - makezero
     - misspell
     - nakedret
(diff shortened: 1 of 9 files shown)


domdomegg:
@claude please review

PR #308, opened 2025-08-26
Add  $schema to server.json specification to enable versioning, IDE autocomplete

### LINKED ISSUE #139: Plan for schema version changes in the ServerDetail model

domdomegg:
## Summary
- Add optional `$schema` property to ServerDetail specification
- Update JSON Schema, OpenAPI spec, Go models, examples, and publisher template
- Enable schema versioning and IDE validation support

## Changes
- Added `$schema` as optional string property with URI format in specs
- Updated Go ServerDetail struct with new Schema field
- Added `$schema` to examples and publisher template
- Copy schema file to ../static for GitHub Pages hosting

Fixes #139

### DIFF
--- a/docs/server-json/repository_references.md
+++ b/docs/server-json/repository_references.md
@@ -1,6 +1,6 @@
 # Repository References in server.json
 
-The [`server.json` schema](schema.json) MAY contain a `repository` property at the root of the JSON object. The `repository` object provides metadata about the MCP server's source code. This enables users and security experts to inspect the code of the MCP service, thereby improving the transparency of what the MCP server is doing at runtime.
+The [`server.json` schema](server.schema.json) MAY contain a `repository` property at the root of the JSON object. The `repository` object provides metadata about the MCP server's source code. This enables users and security experts to inspect the code of the MCP service, thereby improving the transparency of what the MCP server is doing at runtime.
 
 The inclusion of the `repository` object is RECOMMENDED for both local and remote MCP servers.
 
--- a/docs/new_package_registry.md
+++ b/docs/new_package_registry.md
@@ -4,7 +4,7 @@ The MCP Registry project is a **metaregistry**, meaning that it hosts metadata f
 
 For local MCP servers, the MCP Registry has pointers in the `packages` node of the [`server.json`](server-json/README.md) schema that refer to packages in supported package managers.
 
-The list of supported package managers for hosting MCP servers is defined by the `properties.packages[N].properties.registry_name` string enum in the [`server.json` schema](server-json/schema.json). For example, this could be "npm" (for npmjs.com packages) or "pypi" (for PyPI packages).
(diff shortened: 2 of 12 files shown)


docs/server-json/repository_references.md
```diff
@@ -1,6 +1,6 @@
 # Repository References in server.json
 
-The [`server.json` schema](schema.json) MAY contain a `repository` property at the root of the JSON object. The `repository` object provides metadata about the MCP server's source code. This enables users and security experts to inspect the code of the MCP service, thereby improving the transparency of what the MCP server is doing at runtime.
+The [`server.json` schema](server.schema.json) MAY contain a `repository` property at the root of the JSON object. The `repository` object provides metadata about the MCP server's source code. This enables users and security experts to inspect the code of the MCP service, thereby improving the transparency of what the MCP server is doing at runtime.
```
johnw188, line 3: Should this be `The [`server.schema.json` schema]`?

domdomegg: Good q - we expect people to create `server.json` files, but then we want a schema to describe these server.json files. Which is 'the server.json schema', or 'server.schema.json'. So this is correct :)

Issue #283, opened 2025-08-19
auth: Configurable OIDC authentication method


domdomegg:
## Summary

Add a configurable OIDC authentication method that allows deployers to configure OIDC login with environment variables, enabling integration with any OIDC provider.

## Motivation

1. **Admin access for @modelcontextprotocol.io accounts**: We need a way for people to get admin permissions for management endpoints (#182). This could enable doing this with @modelcontextprotocol.io Google Cloud Identity accounts. 
2. **Flexibility for other deployments**: While not a primary goal, making the codebase more generic allows others to deploy their own registries with their preferred OIDC provider.

## Proposed Solution

Add a new OIDC endpoints to enable an OIDC flow:
- `/v0/auth/oidc/start`, which constructs an auth URL with required OIDC parameters (client_id, redirect_uri, scope, state, nonce) and redirects user to the identity provider's authorization endpoint
- `/v0/auth/oidc/callback`, which receives the authorization code from the identity provider, exchanges code for tokens via backchannel POST to token endpoint and validates the ID token (signature, issuer, audience, expiry, nonce)

I'm not quite sure what's best to do on the callback - maybe it could do two things:
- If no redirect set in start, it just displays the registry token for easy copy paste into tools (this could change to redirecting to the signed-in UI at a much later date if/when we have a frontend)
- If redirect set, redirect with the registry token to that. This could then be something like localhost:<random-port>, which could enable CLI tools like publisher (/ a new special tool for admin purposes) to get the token.

Alternative: just have `/v0/auth/oidc` to exchange the oidc id token for a registry token. And then have some separate frontend/tool responsible for getting the id token.

### Configuration (Environment Variables)

```bash
# OIDC Provider Configuration
MCP_REGISTRY_OIDC_ENABLED=true
MCP_REGISTRY_OIDC_ISSUER=https://accounts.google.com  # any OIDC that provides a `.well-known/openid-configuration`
MCP_REGISTRY_OIDC_AUDIENCE=mcp-registry

# Require special claims (this feels a little jank, open to suggestions about how we format this better...)
MCP_REGISTRY_OIDC_EXTRA_CLAIMS=[{"hd":"modelcontextprotocol.io"}]

# Some Permission Configuration (this feels a little jank, open to suggestions about how we format this better...)
MCP_REGISTRY_OIDC_EDIT_PERISSIONS=*
MCP_REGISTRY_OIDC_PUBLISH_PERISSIONS=*  # Or grant specific patterns
```

[...]

Issue #182, opened 2025-07-13
"Admin" panel for the official registry API


tadasant:
We need to work out the scope and "who would have this permission" type of questions, but in general I believe there will be some need for an escape hatch of this nature for a while. e.g. in the case of some spam attack, malicious attack, or even just an edge case bug that gets our data into a bad state. It would be good to have something simple in place here so we don't have to make direct changes to infra via privileged access to e.g. a Postgres database.

domdomegg:
As part of this issue we should add docs on how to use this API to takedown a malicious server, with an easy to follow runbook for maintainers to follow.

PR #315, opened 2025-08-27
feat: add admin edit endpoint for server moderation

domdomegg:
## Summary

Implements `PUT /v0/servers/{id}` admin edit endpoint to address part of issue #182, providing an API-based solution for server moderation without requiring direct database access.

- **New endpoint**: `PUT /v0/servers/{id}` for editing existing servers
- **Security**: Uses existing JWT auth with `PermissionActionEdit` permissions  
- **Anti-undelete protection**: Added `ServerStatusDeleted` status that cannot be reverted
- **Full implementation**: Database, service, API, and comprehensive test coverage

## Key Use Cases

- Mark malicious servers as `deleted` for takedown
- Update server metadata for moderation actions

Related to: https://github.com/modelcontextprotocol/registry/issues/182 (but that is not _done_ yet because we need admins to be able to login (probably with #283), and need to add docs on how to do the takedown, and probably guidelines on what we will/won't takedown)

### DIFF
--- a/docs/server-json/server.schema.json
+++ b/docs/server-json/server.schema.json
@@ -66,9 +66,9 @@
         },
         "status": {
           "type": "string",
-          "enum": ["active", "deprecated"],
+          "enum": ["active", "deprecated", "deleted"],
           "default": "active",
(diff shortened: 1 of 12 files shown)


domdomegg:
@claude please review

PR #316, opened 2025-08-27
feat: implement OIDC authentication for admin users

### LINKED ISSUE #283: auth: Configurable OIDC authentication method

domdomegg:
## Summary

This PR implements OIDC authentication support for admin users, specifically configured for Google Cloud Identity. The implementation includes:

- Configurable OIDC authentication endpoints and validation
- Google Cloud Identity integration for admin access
- JWT token validation and user session management
- Deployment configuration updates for OIDC provider settings

Fixes #283

### DIFF
--- a/internal/api/handlers/v0/auth/oidc.go
+++ b/internal/api/handlers/v0/auth/oidc.go
@@ -285,7 +285,6 @@ func RegisterOIDCEndpoints(api huma.API, cfg *config.Config) {
 
 // ExchangeToken exchanges an OIDC ID token for a Registry JWT token
 func (h *OIDCHandler) ExchangeToken(ctx context.Context, oidcToken string) (*auth.TokenResponse, error) {
-
 	// Validate OIDC token
 	claims, err := h.validator.ValidateToken(ctx, oidcToken)
 	if err != nil {
@@ -318,7 +317,6 @@ func (h *OIDCHandler) ExchangeToken(ctx context.Context, oidcToken string) (*aut
 
 // StartAuth initiates the OIDC authorization flow
 func (h *OIDCHandler) StartAuth(_ context.Context, redirectURI string) (string, error) {
-
 	// Generate state and nonce for security
 	state, err := generateRandomString(32)
 	if err != nil {
@@ -350,7 +348,6 @@ func (h *OIDCHandler) StartAuth(_ context.Context, redirectURI string) (string, 
 
 // HandleCallback handles the OIDC callback
 func (h *OIDCHandler) HandleCallback(ctx context.Context, code, state string) (*auth.TokenResponse, error) {
-
 	// Validate state and retrieve session
 	session, exists := h.sessions[state]
 	if !exists {
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,3 @@ MCP_REGISTRY_OIDC_EXTRA_CLAIMS=[{"hd":"modelcontextprotocol.io"}]
 # Grant admin permissions to OIDC-authenticated users
 MCP_REGISTRY_OIDC_EDIT_PERMISSIONS=*
 MCP_REGISTRY_OIDC_PUBLISH_PERMISSIONS=*
-
-# Note: OAuth2 client credentials (CLIENT_ID/CLIENT_SECRET) not needed for direct token exchange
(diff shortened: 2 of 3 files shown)


domdomegg:
@claude please review

internal/api/handlers/v0/auth/oidc.go
```diff
@@ -0,0 +1,443 @@
...
+		Method:      http.MethodPost,
+		Path:        "/v0/auth/oidc",
+		Summary:     "Exchange OIDC ID token for Registry JWT",
+		Description: "Exchange an OIDC ID token from any configured provider for a short-lived Registry JWT token",
+		Tags:        []string{"auth"},
+	}, func(ctx context.Context, input *OIDCTokenExchangeInput) (*v0.Response[auth.TokenResponse], error) {
+		response, err := handler.ExchangeToken(ctx, input.Body.OIDCToken)
+		if err != nil {
+			return nil, huma.Error401Unauthorized("Token exchange failed", err)
+		}
```
crondinini-ant, line 237: Would it make sense to have separate codes depending on the error at some point for the endpoints? At first glance, without all the context, 401 seems too generic.

crondinini-ant (approved):

jerome3o-anthropic (approved):

internal/api/handlers/v0/auth/oidc.go
domdomegg: Fair - I think at some point we might want to look at improving the status codes and API ergonomics generally. I also think the way we send back errors to the client is a bit weird and undersanitized. But I think both of these are probably a bigger separate task.

PR #318, opened 2025-08-28
Deduplicate server JSON structs between publisher and registry

rdimitrov:
## Motivation and Context

The following PR eliminates duplicate struct definitions by migrating the publisher tool to use canonical types from pkg/model, creating a single source of truth across the codebase. The publisher tool previously had its own copies of structs that were duplicated, ie. ServerJSON, Package, Repository, VersionDetail, EnvironmentVariable, RuntimeArgument and a few others which was not optimal

**Changes in the PR:**
  - Refactored the structs to single canonical types located at pkg/model
  - Refactor the implementation to use/build canonical structures directly
  - Remove all duplicate definitions (5 structs eliminated)
  - Publisher now uses model.ServerJSON, model.Package, model.Repository, etc.
  - Renamed some of the structs to better reflect their use case, i.e. ServerJSON instead of ServerDetail
  - Fixes an issue where the Remote property was missing from the previous publisher copy of the Server

**Benefits:**
  - Single source of truth - Schema changes in one place
  - Type safety - Compiler catches mismatches
  - Enhanced validation - Field constraints and bson tags
  - Zero breaking changes - Same CLI interface and JSON output
  - Future-proof - New canonical fields automatically available

Supersedes: #217 and #240

## How Has This Been Tested?

Locally

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist


rdimitrov:
As a follow up I am thinking to generate the server.json schema out of the go struct describing it. This should make things much more consistent and easy. This way the go object would be the source of truth for everything which is also a good practice instead of keeping in sync the schema and the object manually.

pkg/api/v1/types.go
```diff
@@ -0,0 +1,42 @@
+package v1
```
domdomegg, line 1: I think we should have this in v0, because all our current APIs are under `/v0/`

domdomegg: https://github.com/modelcontextprotocol/registry/pull/320

rdimitrov: Ah, my bad, thanks for fixing it 🙏

PR #323, opened 2025-08-28
feat: add admin operations documentation and scripts for server management

### LINKED ISSUE #182: "Admin" panel for the official registry API

domdomegg:
## Summary
- Added comprehensive admin operations documentation
- Included scripts for common server management tasks

Together with #315, #316, this fixes #182


rpb-ant (commented):
Approved

PR #328, opened 2025-08-30
102 infra for monitoring

pree-dew:
This includes infra setup for metrics collection and visualisation

## Motivation and Context
It is important to have basic observability setup, so that common issues can be identified quickly

## How Has This Been Tested?
- Tested on local
- Tested on one of the production setup

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- This includes set up of following components:
  - Victoriametrics single node cluster for storing metrics data with persistent storage configuration for local and production env.
  - Vmagent with target discovery for mcp-registry pods with scrape interval of 30s.
  - Grafana for visualising metrics and setting alerts.
     - Datasource of victoriametrics is pre-configured as config map
     - Persistent volume for basic configuration
     - Sqlite for local and Postgres for production setup for storing dashboard, alerts etc details
 
 ## Metrics in action
Screenshot 2025-08-31 at 4 35 28 AM


deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+
+// getEnvironmentConfig returns environment-specific configuration
+func getEnvironmentConfig(ctx *pulumi.Context) *EnvironmentConfig {
+	conf := config.New(ctx, "mcp-registry")
+	provider := conf.Get("provider")
+	if provider == "" {
+		provider = "local"
+	}
+
+	isProduction := provider != "local"
```
domdomegg: I think we should avoid having things be very different locally and in production: ideally we should be able to run the whole stack locally

deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+			"feature_toggles": pulumi.Map{
+				"enable": pulumi.String("ngalert"), // Modern alerting system
+			},
+			// SMTP disabled - no email alerts needed
+			"smtp": pulumi.Map{
+				"enabled": pulumi.Bool(false),
+			},
+			"security": pulumi.Map{
+				"admin_user":       pulumi.String("admin"),
+				"secret_key":       pulumi.String("SW2YcwTIb9zpOOhoPsMm"),
```
domdomegg: I'm not super familiar with what this value is for, but seems concerning to have something called `secret_key` hardcoded (possibly it isn't but wanted to flag)!

domdomegg:
@claude can you review?

deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+
+	}
+
+	// Add namespace to dependencies
+	grafanaDependsOn = append(grafanaDependsOn, ns)
+
+	// Deploy Grafana with environment-aware configuration
+	_, err = helm.NewChart(ctx, "grafana", helm.ChartArgs{
+		Chart:   pulumi.String("grafana"),
+		Version: pulumi.String("7.0.19"),
```
domdomegg: I think this is a pretty old version of grafana - is there a reason we're not using a later one?

deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+			"accessModes": pulumi.Array{pulumi.String("ReadWriteOnce")},
+		}
+	} else {
+		persistenceConfig = pulumi.Map{
+			"type":    pulumi.String("emptyDir"),
+			"enabled": pulumi.Bool(false),
+		}
+	}
+
+	grafanaValues := pulumi.Map{
```
domdomegg: I'm not quite sure what's going on, but I reckon it's somewhere around here - it seems like deploying this seems to want to _always_ act like there's a change, which creates noise in the diffs. Are we able to avoid that somehow?

```
     Type                                                       Name                      Status               Info
     pulumi:pulumi:Stack                                        mcp-registry-infra-local                       
     └─ kubernetes:helm.sh/v3:Chart                             grafana                                        
 ~      ├─ kubernetes:rbac.authorization.k8s.io/v1:ClusterRole  grafana-clusterrole       updated (0.71s)      [diff: +rules]
 ~      ├─ kubernetes:rbac.authorization.k8s.io/v1:Role         monitoring/grafana        updated (1s)         [diff: +rules]
 +-     ├─ kubernetes:core/v1:ConfigMap                         monitoring/grafana        replaced (0.57s)     [diff: ~data]
 ~      └─ kubernetes:apps/v1:Deployment                        monitoring/grafana        updated (18s)        [diff: ~spec]
```

deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+	} else {
+		persistentVolumeConfig = pulumi.Map{
+			"enabled": pulumi.Bool(false),
+		}
+	}
+
+	// Install VictoriaMetrics Single using Helm chart
+	_, err := helm.NewChart(ctx, "victoria-metrics-single", helm.ChartArgs{
+		Chart:     pulumi.String("victoria-metrics-single"),
+		Version:   pulumi.String("0.19.0"),
```
domdomegg: this is also an old version

deploy/pkg/k8s/monitoring.go
```diff
@@ -0,0 +1,587 @@
...
+
+	return nil
+}
+
+func deployVMAgent(ctx *pulumi.Context, cluster *providers.ProviderInfo, namespace *corev1.Namespace, environment string, envConfig *EnvironmentConfig) error {
+	agentConfig := envConfig.VMAgentConfig
+
+	_, err := helm.NewChart(ctx, "victoria-metrics-agent", helm.ChartArgs{
+		Chart:     pulumi.String("victoria-metrics-agent"),
+		Version:   pulumi.String("0.12.0"),
```
domdomegg: this is also an old version

deploy/pkg/k8s/monitoring.go
pree-dew: @domdomegg Ack,  let me check this.

deploy/pkg/k8s/monitoring.go
pree-dew: @domdomegg Ack,  let me check this.

deploy/pkg/k8s/monitoring.go
pree-dew: @domdomegg Ack,  let me check this.

deploy/pkg/k8s/monitoring.go
pree-dew: @domdomegg I also wanted to keep things same but in terms of resource requirements for persistent storage it might create problem to run it locally as resources will be high for production. Any suggestion for this?

domdomegg:
@pree-dew I've raised https://github.com/pree-dew/modelcontextprotocol-registry/pull/2, which I think addresses my comments above - if happy with those changes, can you merge that in and then I can get this all merged in.

pree-dew:
@domdomegg
- Keeping postgres for all env makes sense. 
- Switching to google auth also makes sense. 

I am not sure about these changes:
1. Persistent config are not part of Victoriametrics, it is 16Gi bydefault which will be required for prod and local setup both, it is on higher side. 
2. Resource requirements are not given for limit and requests, not sure about the reason behind this change.
3. By default retention for victoriametrics is 1d, we will loose data after a data.
4. Persistent config for grafana stores some part of config, not sure about that change for storage size.

domdomegg:
1. I think this is fine locally - I believe both minikube and kind don't really allocate the storage, they just cap it at that amount. So local dev doesn't actually take up that space.
2. Fine to add these back, especially the limits. I was probably overcutting here.
3. Fine to hardcode this to a longer retention period across all envs e.g. 14d - I think I'd just want consistency across environments.
4. I think if it's given a postgres database, it uses that instead - and also means it can store dashboards etc. I think it also means it's captured in our backups automatically, which is nice. So I don't think we need grafana's custom persistence.

pree-dew:
1, 2 and 3 makes sense, for 4th generally for file provisioning we use PVC even if we have external database but what I found is that provisioning files are only read only input so configmap along with external database will be fine for us.

@domdomegg Should I make changes for 2, and 3rd?

domdomegg:
Yes please!

pree-dew:
Btw, good catch on local setup storage, I thought the storage class will be same as EKS but it turns out it is local path which keeps PVC as bounded 
```
kubectl get storageclass
NAME                   PROVISIONER             RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
local-path (default)   rancher.io/local-path   Delete          WaitForFirstConsumer   false                  2d2h
```

It's a miss from my side, I will keep this kind of check in mind for upcoming tasks.

pree-dew:
@domdomegg Requested changes are pushed.

domdomegg (approved):

PR #335, opened 2025-09-01
Database implementation cleanup

domdomegg:
## Motivation and Context

The database logic was getting a bit unwieldy and fragile - especially because business logic that should really have lived at the registry service level was being pushed into the persistence level.

This PR tightens the interface, simplifies the code there, as well as rips out a bunch of mocks etc. we don't need. 

## How Has This Been Tested?

Locally everything works

## Breaking Changes

The list endpoint was surprisingly always filtering for isLatest only, which seems like a bug? I have removed this filter.


domdomegg:
@claude please review

domdomegg:
@claude please review

jerome3o-anthropic (approved):
approving to unblock

tadasant:
> The list endpoint was surprisingly always filtering for isLatest only, which seems like a bug? I have removed this filter.

@domdomegg that seems right to me, no? In general, a consumer would expect a "list of all MCP servers" to just include one entry per unique MCP server. Versus if we don't filter for isLatest, we' might get hundreds of entries per MCP server (for those that have gone through a lot of version bumps)? Maybe I'm not quite following

PR #339, opened 2025-09-02
fix: remove id property from Server schema

### LINKED ISSUE #336: The 'id 'property should be required in OpenAPI spc
In https://github.com/modelcontextprotocol/registry/blob/main/docs/server-registry-api/openapi.yaml, there is 'id' property in both ‘Server’ and 'io.modelcontextprotocol.registry', however none of them are marked as 'required'.

Expected:

- The 'id' property should only be defined in ‘Server’ or 'io.modelcontextprotocol.registry' 
- The 'id' property should be 'required'

domdomegg:
The Server schema should not have its own `id` property as the ID is managed by the registry system through the `_meta["io.modelcontextprotocol.registry"].id` field.

Fixes #336

### DIFF
--- a/docs/server-registry-api/openapi.yaml
+++ b/docs/server-registry-api/openapi.yaml
@@ -203,10 +203,6 @@ components:
         - description
         - version_detail
       properties:
-        id:
-          type: string
-          format: uuid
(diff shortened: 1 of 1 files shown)


rdimitrov:
lgtm 👍

PR #345, opened 2025-09-02
Reflect the latest registry changes in openapi.yaml

rdimitrov:
## Motivation and Context

The following PR:
* Adds a new `transport` type used in Package and Remotes
* Added `docker` to the `runtime_hint` examples
* Added support for templated URLs. It is available for streamable-http and sse transports for Packages. I didn't thought of reasons to have this for Remotes, but we can enable it if needed.
* Added validation for templated URLs so we can ensure all referenced placeholder variables are available. 
* For streamable-http and SSE transport types the URL is now a required property
* Updated the publisher init command to reflect these changes
* Updated the seed.json file
* Added unit tests for the transport type related changes
* Updated the server.json schema and the external pkg structs to reflect the new transport type
* Updated the examples

Thanks to @kkkarthik for spotting it! 🙏 

## How Has This Been Tested?

Ran the tests locally

## Breaking Changes

Yes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/server-json/examples.md
+++ b/docs/server-json/examples.md
@@ -21,6 +21,9 @@
       "registry_base_url": "https://registry.npmjs.org",
       "identifier": "@modelcontextprotocol/server-brave-search",
       "version": "1.0.2",
+      "transport": {
+        "type": "stdio"
+      },
       "environment_variables": [
         {
           "name": "BRAVE_API_KEY",
@@ -60,6 +63,9 @@ Suppose your MCP server application requires a `mcp start` CLI arguments to star
       "registry_base_url": "https://api.nuget.org",
       "identifier": "Knapcode.SampleMcpServer",
       "version": "0.4.0-beta",
+      "transport": {
+        "type": "stdio"
+      },
       "package_arguments": [
         {
           "type": "positional",
@@ -109,6 +115,9 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
       "registry_base_url": "https://registry.npmjs.org",
       "identifier": "@modelcontextprotocol/server-filesystem",
       "version": "1.0.2",
+      "transport": {
+        "type": "stdio"
+      },
       "package_arguments": [
         {
           "type": "positional",
@@ -132,6 +141,9 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
       "registry_base_url": "https://docker.io",
       "identifier": "mcp/filesystem",
       "version": "1.0.2",
+      "transport": {
+        "type": "stdio"
+      },
       "runtime_arguments": [
         {
           "type": "named",
@@ -201,7 +213,7 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
   },
   "remotes": [
     {
-      "transport_type": "sse",
+      "type": "sse",
       "url": "http://mcp-fs.anonymous.modelcontextprotocol.io/sse"
     }
   ],
@@ -241,6 +253,9 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
       "identifier": "weather-mcp-server",
       "version": "0.5.0",
       "runtime_hint": "uvx",
+      "transport": {
+        "type": "stdio"
+      },
       "environment_variables": [
         {
           "name": "WEATHER_API_KEY",
@@ -294,6 +309,9 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
       "identifier": "Knapcode.SampleMcpServer",
       "version": "0.5.0",
       "runtime_hint": "dnx",
+      "transport": {
+        "type": "stdio"
+      },
       "environment_variables": [
         {
           "name": "WEATHER_CHOICES",
@@ -340,6 +358,9 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
       "registry_base_url": "https://docker.io",
       "identifier": "example/database-manager-mcp",
       "version": "3.1.0",
+      "transport": {
+        "type": "stdio"
+      },
       "runtime_arguments": [
         {
           "type": "named",
@@ -451,6 +472,9 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
       "identifier": "@example/hybrid-mcp-server",
       "version": "1.5.0",
       "runtime_hint": "npx",
+      "transport": {
+        "type": "stdio"
+      },
       "package_arguments": [
         {
           "type": "named",
@@ -468,7 +492,7 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
   ],
   "remotes": [
     {
-      "transport_type": "sse",
+      "type": "sse",
       "url": "https://mcp.anonymous.modelcontextprotocol.io/sse",
       "headers": [
         {
@@ -490,7 +514,7 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
       ]
     },
     {
-      "transport_type": "streamable-http",
+      "type": "streamable-http",
       "url": "https://mcp.anonymous.modelcontextprotocol.io/http"
     }
   ],
@@ -532,7 +556,10 @@ The `dnx` tool ships with the .NET 10 SDK, starting with Preview 6.
       "registry_base_url": "https://github.com",
       "identifier": "https://github.com/modelcontextprotocol/text-editor-mcpb/releases/download/v1.0.2/text-editor.mcpb",
       "version": "1.0.2",
-      "file_sha256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce"
+      "file_sha256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce",
+      "transport": {
+        "type": "stdio"
+      }
     }
   ],
   "_meta": {
@@ -574,6 +601,9 @@ This example shows an MCPB (MCP Bundle) package that:
       "registry_base_url": "https://registry.npmjs.org",
       "identifier": "@legacy/old-weather-server",
       "version": "0.9.5",
+      "transport": {
+        "type": "stdio"
+      },
       "environment_variables": [
         {
           "name": "WEATHER_API_KEY",
--- a/docs/server-json/server.schema.json
+++ b/docs/server-json/server.schema.json
@@ -114,6 +114,7 @@
           "examples": [
             "npx",
             "uvx",
+            "docker",
             "dnx"
           ]
         },
--- a/docs/server-json/server.schema.json
+++ b/docs/server-json/server.schema.json
@@ -117,15 +117,19 @@
             "dnx"
           ]
         },
-        "transport_type": {
-          "type": "string",
-          "enum": [
-            "streamable-http",
-            "sse",
-            "stdio"
+        "transport": {
+          "anyOf": [
+            {
+              "$ref": "#/$defs/StdioTransport"
+            },
+            {
+              "$ref": "#/$defs/StreamableHttpTransport"
+            },
+            {
+              "$ref": "#/$defs/SseTransport"
+            }
           ],
-          "description": "Transport protocol type for the package",
-          "example": "stdio"
+          "description": "Transport protocol configuration for the package"
         },
         "runtime_arguments": {
           "type": "array",
@@ -323,26 +327,70 @@
         }
       ]
     },
-    "Remote": {
+    "StdioTransport": {
       "type": "object",
       "required": [
-        "transport_type",
+        "type"
+      ],
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": [
+            "stdio"
+          ],
+          "description": "Transport type",
+          "example": "stdio"
+        }
+      }
+    },
+    "StreamableHttpTransport": {
+      "type": "object",
+      "required": [
+        "type",
         "url"
       ],
       "properties": {
-        "transport_type": {
+        "type": {
           "type": "string",
           "enum": [
-            "streamable-http",
+            "streamable-http"
+          ],
+          "description": "Transport type",
+          "example": "streamable-http"
+        },
+        "url": {
+          "type": "string",
+          "description": "URL template for the streamable-http transport. Variables in {curly_braces} reference argument value_hints, argument names, or environment variable names. After variable substitution, this should produce a valid URI.",
+          "example": "https://api.example.com/mcp"
+        },
+        "headers": {
+          "type": "array",
+          "description": "HTTP headers to include",
+          "items": {
+            "$ref": "#/$defs/KeyValueInput"
+          }
+        }
+      }
+    },
+    "SseTransport": {
+      "type": "object",
+      "required": [
+        "type",
+        "url"
+      ],
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": [
             "sse"
           ],
-          "description": "Transport protocol type",
+          "description": "Transport type",
           "example": "sse"
         },
         "url": {
           "type": "string",
           "format": "uri",
-          "description": "Remote server URL",
+          "description": "Server-Sent Events endpoint URL",
           "example": "https://mcp-fs.example.com/sse"
         },
         "headers": {
@@ -378,7 +426,14 @@
             "remotes": {
               "type": "array",
               "items": {
-                "$ref": "#/$defs/Remote"
+                "anyOf": [
+                  {
+                    "$ref": "#/$defs/StreamableHttpTransport"
+                  },
+                  {
+                    "$ref": "#/$defs/SseTransport"
+                  }
+                ]
               }
             },
             "_meta": {
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -30,10 +30,51 @@ func HasNoSpaces(s string) bool {
 	return !strings.Contains(s, " ")
 }
 
-// IsValidURL checks if a URL is in valid format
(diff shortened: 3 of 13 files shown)


docs/server-registry-api/openapi.yaml
```diff
@@ -325,6 +325,11 @@ components:
           description: A mapping of environment variables to be set when running the package.
           items:
             $ref: '#/components/schemas/KeyValueInput'
+        transport_type:
+          type: string
+          enum: [ streamable-http, sse, stdio ]
```
connor4312: This isn't enough for a client to know how to consume the package -- http/sse packages will need to expose their URL too, perhaps through templated input that can reference their args/environment variables?

connor4312:
Continuing from https://github.com/modelcontextprotocol/registry/pull/340#issuecomment-3245742819

> I think it's not necessary to be advertised via remotes, usually it's on host-address:8080 (depending on the server there can be a flag/env var to set the port it listens on) and the host-address bit depends on your runtime. From there on if you say you support/want to talk over sse there should be an /sse endpoint and similarly for streamable there's /message I believe (someone correct me if I got it wrong, haven't checked it in detail)

It is definitely necessary. Ports are not standardized and of course only one process can listen on a port. Nor are particular subpaths, such as `/sse` or `/message`, standardized in the spec. For a client to consume an http/sse package there must be a URL or algorithm to derive the URL specified in the package's metadata on the registry.

We should also think about how we can ensure HTTP MCP servers running locally can be connected to securely. A common attack vector is escalation of privledge from services or users running on the same machine or network. Best practice for this kind of thing is to include a randomized nonce (be it a header, query string, or URL fragment -- a header is easiest for MCP) that the client and server share. So I suggest a means to include that as well when thinking about how the URL gets specified.

docs/server-registry-api/openapi.yaml
rdimitrov: I think that matches my assumption too. So far I've seen SSE/streamable-http servers typically allow you to set the `port` as an arg/env var (usually defaults to 8080) while the `host` usually comes from the runtime. _And of course the rest of the path portion should follow the spec (e.g. /mcp, /sse, etc.)._ 

I found this to be enough for clients to construct the URL, but I don't mind if you think we should have something more explicit, i.e. an additional templated property 👍 

**edit: The page refreshed right after I posted this and now I saw your next comment**

connor4312: Assuming the default port and path might work for a toy client but it's not something that works for complex scenarios where a client manages multiple MCP servers concurrently for example.

Maybe for the template we do something like this:

```yaml
# in the package -- allow including extra data within the transport type
        transport_type:
          anyOf:
            - $ref: '#/components/schemas/StdioTransport'
            - $ref: '#/components/schemas/HttpTransport'
# stdio definition is pretty easy/static
    StdioTransport:
      type: object
      properties:
        type:
          type: string
          enum: [stdio]
      required:
        - type
# http transport type is more involved. While this is similar to http packages,
# I don't think there's a reason to do input with headers for example, so there's no reuse
    HttpTransport:
      type: object
      properties:
        type:
          type: string
          enum: [http]
        url:
          type: string
          example: http://{host}:{port}/my-server/mcp
          description: |
            Fully qualified URL for the streamable HTTP when the server is started. Clients should retry connecting to this URL as it may take some time for the server to be ready.

            Identifiers wrapped in `{curly_braces}` should be looked up and replaced with the corresponding value of the input, based on the `value_hint` in PositionalArguments or the `name` in NamedArguments or KeyValueInput. The input's `default` value should be used if the input was not explicitly configured.
        headers:
          type: array
          items:
            - type: object
              required:
                - name
                - value
              properties:
                name:
                  type: string
                  description: Name of the header.
                  example: SOME_VARIABLE
                value:
                  type: string
                  description: Value of the header. Identifiers may be wrapped in `{curly_braces}` to reference existing variables, following the same rules at HttpTransport.url.
      required:
        - type
        - url
```

This would let it work to specify a URL or defaults based on the server's spawn arguments. The curly_braces format is reused from templated arguments, although the reference mechanic is new. This also works to do nonces (as I suggest below) if we add some tag like `is_randomized` on an argument.

SSE would basically be the same, but as SSE was removed from the spec back in June I'm not convinced we need to support it in new work going forward :)

rdimitrov: Yeah, I see your point now, I agree it's better to make this deterministic for clients 👍 I'll proceed with incorporating your feedback into this PR so we can keep the discussion here 👍 I'll probably address the URL part now and file an issue for the nonce part so we can do it in a follow up.

Thanks for your detailed replies, it is much appreciated! 🙏

rdimitrov: @connor4312 - hey, whenever you get a chance could you take another look? I've updated the PR description too to better match the changes.

docs/server-registry-api/openapi.yaml
```diff
@@ -530,3 +520,36 @@ components:
...
+    StreamableHttpTransport:
+      type: object
+      properties:
+        type:
+          type: string
+          enum: [streamable-http]
+          example: "streamable-http"
+        url:
+          type: string
+          format: uri
```
connor4312: This technically might not be a uri format itself. After interpolation it should produce a URI, but the template string might not be a URI before that. E.g. it could just be `{uri}` if I have a server that takes a named argument `--listen-on {uri}`

connor4312:
API shape overall lgtm, will let others weigh in too :)

docs/server-registry-api/openapi.yaml
rdimitrov: Thanks! I'll update that on both schemas since the server.json has it too 👍

tadasant:
Did I miss a discussion somewhere about removing SSE? I'm not super strongly opposed, but I would guess something like 20-40% of remote servers in use in the wild today are still SSE. So it's a pretty significant move to not allow them in the registry.

I totally agree with removing them for local (not a common use case, has security issues), but would have thought to keep them in remotes

docs/server-json/server.schema.json
```diff
@@ -118,14 +118,15 @@
           ]
         },
         "transport_type": {
```
tadasant: This key feels out of place IMO -- can we just flatten it? i.e. instead of creating `remote.transport_type.type` and `remote.transport_type.url`, etc, we just have `remote.transport_type`, `remote.url`, etc.

docs/server-json/examples.md
```diff
@@ -132,6 +141,9 @@ This will essentially instruct the MCP client to execute `dnx Knapcode.SampleMcp
       "registry_base_url": "https://docker.io",
       "identifier": "mcp/filesystem",
       "version": "1.0.2",
+      "transport_type": {
```
tadasant: Seems like this would be better named:

```suggestion
      "transport": {
```

docs/server-json/server.schema.json
tadasant: i.e. positioning as "`remotes` is an array of `Transport`, which is one of StreamableHttpTransport (or SSETransport, if we choose to re-add that; and perhaps soon WebSocketTransport pending that ongoing SEP)"

docs/server-json/server.schema.json
```diff
@@ -323,27 +324,41 @@
...
-          "example": "sse"
+          "description": "Transport type",
+          "example": "streamable-http"
         },
         "url": {
           "type": "string",
-          "format": "uri",
-          "description": "Remote server URL",
-          "example": "https://mcp-fs.example.com/sse"
+          "description": "URL template for the streamable-http transport. Variables in {curly_braces} reference argument value_hints, argument names, or environment variable names. After variable substitution, this should produce a valid URI.",
```
tadasant, line 364: It seems a little odd to me to reach across other fields (value_hints, argument names, env vars) to fill out this url template. Should we consider keeping this consistent and just allowing for variable definitions adjacent to and/or inside `url` via `KeyValueInput`?

tadasant (commented):
Thanks for updating! Added some questions / suggestions.

Thanks for catching the bit about needing a URL for local streamable @connor4312 and @domdomegg, it's a good point. I would also not be opposed to just nixing HTTP-based local servers for now to cull down on our complexity here; I haven't seen a practical use case for that mode. Easy to just make the enum only allow stdio for now and add it later if folks present use cases.

connor4312:
> HTTP-based local servers for now to cull down on our complexity here

I think this is something that will be wanted eventually. HTTP servers can do extra things like OAuth (and form-based elicitations if that arrives in MCP) and we have an open feature request on VS Code to easily support spawnable MCP servers that use the HTTP transport.

docs/server-json/examples.md
rdimitrov: Yeah, I prefer `transport` too, I'll be happy to rename it 👍

docs/server-json/server.schema.json
rdimitrov: > This key feels out of place IMO -- can we just flatten it? i.e. instead of creating remote.transport_type.type and remote.transport_type.url, etc, we just have remote.transport_type, remote.url, etc.

I tried this initially, but it turned out to be messier in practice. Having a dedicated transport object for each option ended up being clearer as schema validation got tricky, i.e. there wasn’t really an easy way to enforce which fields should be present for a given transport type.

rdimitrov:
> Did I miss a discussion somewhere about removing SSE? I'm not super strongly opposed, but I would guess something like 20-40% of remote servers in use in the wild today are still SSE. So it's a pretty significant move to not allow them in the registry.
> 
> I totally agree with removing them for local (not a common use case, has security issues), but would have thought to keep them in remotes

You're right, we didn't 👍 I did it based on @connor4312's feedback as I agree it feels right to start the registry by being compliant with the spec regarding this. That said I can easily revert it if we agree this might affect adoption and we want to support the deprecated SSE (at least for now).

> HTTP-based local servers for now to cull down on our complexity here

Personally I think we should have this. Especially if we want to allow for enterprise adoption besides the desktop experience of running MCP servers, i.e. it would be better to run a streamable server instead of stdio.

docs/server-json/examples.md
rdimitrov: Actually would it be okay if I address that in a follow-up? I think it’ll be easier to review separately and I don’t want to make this PR too heavy. I can jump on it right after this one’s merged so it shouldn’t cause any extra friction

tadasant:
> I think this is something that will be wanted eventually. HTTP servers can do extra things like OAuth (and form-based elicitations if that arrives in MCP) and we have an open feature request on VS Code to easily support spawnable MCP servers that use the HTTP transport.

> Personally I think we should have this. Especially if we want to allow for enterprise adoption besides the desktop experience of running MCP servers, i.e. it would be better to run a streamable server instead of stdio.

Got it, fair enough, I'll retract my thought of potentially culling them then; let's keep it.

---

> You're right, we didn't 👍 I did it based on @connor4312's feedback as I agree it feels right to start the registry by being compliant with the spec regarding this. That said I can easily revert it if we agree this might affect adoption and we want to support the deprecated SSE (at least for now).

We haven't yet been wielding the registry as a tool for pushing adoption of better practices (for example, we had a very brief idea like "maybe we should _require_ all servers to have a docker image to standardize around that" way back that we shuttered), so I'm inclined to stick to that for now and optimize for registry adoption rather than try to thread a needle with this go live of easy-enough-to-adopt-but-still-pushing-better-practices. Probably should have added "meet the industry where it's currently at" as a design principle. Otherwise it's a slippery slope with making harder decisions like not allowing API Key headers and forcing all auth to be OAuth based. So I think we should keep SSE, at least in Remote, unless folks have strong objections.

docs/server-json/server.schema.json
tadasant: I guess if we rename `transport_type` to just `transport`, this feels a lot better. Still a little odd that the only key in `remote` is `transport`, but maybe we'll come up with something in the future that needs to live on a Remote alongside `transport` that will make us happy we added the layer :)

domdomegg:
A few thoughts:
- I think I'd prefer `transport` to `transport_type`, if we're going to have type and url underneath it
- As much as I am for the death of SSE servers, I think we should prioritize supporting a wide array of servers which will have to include sse initially, and we should use other mechanisms to encourage the move to streamable http
- Reaching over other fields does feel a bit weird, but is maybe okay. I think this is likely to be rare anyways.
- A case I worry about is servers that try to spin up on a random port, no idea how to handle these. I think a reasonable answer to these is 'don't' for now... (appreciate this goes against my 'support wide range of things' above but I think this is much rarer than SSE)

docs/server-json/server.schema.json
domdomegg: > I tried this initially, but it turned out to be messier in practice. Having a dedicated transport object for each option ended up being clearer as schema validation got tricky, i.e. there wasn’t really an easy way to enforce which fields should be present for a given transport type.

I'm not sure I follow this - are you able to expand more on this difficulty? I think I agree with Tadas that remotes should just be an array of transports I think.

connor4312:
>  think a reasonable answer to these is 'don't' for now... 

Yea, this is kind of nasty. In VS Code we have a '`serverReadyAction`' in debug land that is essentially a regex on the process' stdout, but using it is painful and there has not been ideas to improve it in the last decade :) I think it's reasonable for HTTP servers to have some optional argument for a port, a way to reference that in the URL (as in this PR) and then users can change that port if it conflicts with something on their device. Smart clients might even see a `port` parameter and automatically default to an unused port on the machine without user intervention.

rdimitrov:
Thank you all for reviewing and commenting this change, really appreciate it! 💯 

In the interest of wrapping this up soon I'll summarise the remaining action items we agreed on that I have to address:
- Bring back the SSE transport type as it can affect adoption in a negative way
- Rename `transport_type` to `transport`. I'd like to do that in a follow up PR thought so it's easier to review, is this okay?

Let me know if I missed something 👍 

In the meantime I'll make sure to address these (bring back SSE here + the rename on top of it)

domdomegg:
I think it's those two (bring back `sse`, rename `transport_type` to `transport`) plus [using transports as remotes directly](https://github.com/modelcontextprotocol/registry/pull/345#discussion_r2319523040)?

Happy to do it all in one PR - as long as you keep the commits separate think it's easier to reason about as one unit.

docs/server-json/server.schema.json
rdimitrov: I'm sorry, I see what you mean now 🤦 

For some reason I read that as "let's keep up the old approach of having these properties under the remote object" instead of "replace the remote object with a list of transports". 

Yes, your suggestion makes total sense! I'll implement it 🙏

rdimitrov:
> I think it's those two (bring back `sse`, rename `transport_type` to `transport`) plus [using transports as remotes directly](https://github.com/modelcontextprotocol/registry/pull/345#discussion_r2319523040)?
> 
> Happy to do it all in one PR - as long as you keep the commits separate think it's easier to reason about as one unit.

Perfect 🙏 On it then 👍

rdimitrov:
Alright, I've rebased the changes along with implementing the feedback items from above so it should be ready to review 👍

internal/validators/utils.go
```diff
@@ -43,8 +84,67 @@ func IsValidURL(rawURL string) bool {
 		return false
 	}
 
-	if u.Host == "" || u.Hostname() == "localhost" {
+	if u.Host == "" {
```
domdomegg, line 87: Was this change intentional? Just flagging as might have been a merge conflict accident with #355

domdomegg: Ah just seen it's moved to IsValidRemoteURL below, ignore me 😅

internal/validators/validators.go
```diff
@@ -130,11 +136,83 @@ func validateArgumentValueFields(name, value, defaultValue string) error {
...
+	
+	return variables
+}
+
+// validatePackageTransport validates a package's transport with templating support
+func validatePackageTransport(transport *model.Transport, availableVariables []string) error {
+	// Validate transport type is supported
+	switch transport.Type {
+	case model.TransportTypeStdio:
+		// No additional validation needed for stdio - URL should be empty
```
domdomegg, line 176: do we want to validate the the url _is_ empty?

internal/validators/validators.go
```diff
@@ -130,11 +136,83 @@ func validateArgumentValueFields(name, value, defaultValue string) error {
...
+			return fmt.Errorf("%w: %s", ErrInvalidRemoteURL, transport.URL)
+		}
+		return nil
+	default:
+		return fmt.Errorf("unsupported transport type: %s", transport.Type)
+	}
+}
+
+// validateTransport validates a remote transport (no templating allowed)
+func validateTransport(obj *model.Transport) error {
```
domdomegg, line 200: maybe rename `validateRemoteTransport` or something?

domdomegg:
@claude can you raise two separate PRs for the review comments above? if you get stuck instead raise an issue assigned to `@rdimitrov`

domdomegg:
➡️ followups in https://github.com/modelcontextprotocol/registry/pull/365

rdimitrov:
@domdomegg - thanks for addressing the comments 🙏 Funny enough I must have lost the most important commit about applying the changes on the openapi schema while I was rebasing the changes at the end 😄

BobDickinson:
Should these changes (particularly changing the Remote schema type to Transport and adding transport to the Package) have been made to the "generic" OpenAPI spec also?

https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/api/openapi.yaml

I tend to use that one for generating my TypeScript client types, but noticed it doesn't have the package transport attribute.

Happy to submit a PR if so (or defer to someone with more context).

domdomegg:
Yep @BobDickinson!

---

As a side note maybe we should add tests to keep these all in sync? Seems like it should be possible to check that the Official MCP Registry API is some superset of the generic MCP registry API spec?

➡️ https://github.com/modelcontextprotocol/registry/issues/567

PR #347, opened 2025-09-02
Remove Azure/AKS deployment configuration from deploy

domdomegg:
## Summary
- Removed Azure deployment section and references from README
- Updated project structure to reflect current GCP prod/staging configs  
- Removed Azure Load Balancer annotation from ingress controller
- Updated provider documentation to show GCP and local options only
- Added monitoring.go to project structure documentation

This cleans up outdated Azure/AKS deployment configuration that is no longer used, leaving only the current GCP-based deployment structure.


rdimitrov (commented):
lgtm 👍

PR #348, opened 2025-09-02
Validate server registry urls are allowlisted for that registry type

domdomegg:
## Summary
Closes a validation gap that allowed servers to be published with unsupported registry types and base URLs.

## Changes
- **Registry Type Validation**: Only allow `npm`, `pypi`, `oci`, `nuget`, `mcpb`
- **Registry Base URL Validation**: Enforce strict allowlist of official registry URLs
- **Required Fields**: Registry type and base URL required when package identifier provided
- **Comprehensive Tests**: Clear test structure showing valid vs invalid cases side-by-side

## Validation Rules
- `npm` → `https://registry.npmjs.org`
- `pypi` → `https://pypi.org`
- `oci` → `https://docker.io` 
- `nuget` → `https://api.nuget.org`
- `mcpb` → `https://github.com` or `https://gitlab.com`

All other combinations are rejected with clear error messages.

### DIFF
--- a/internal/validators/constants.go
+++ b/internal/validators/constants.go
@@ -13,6 +13,11 @@ var (
 	// Remote validation errors
 	ErrInvalidRemoteURL = errors.New("invalid remote URL")
 
+	// Registry validation errors
+	ErrUnsupportedRegistryType      = errors.New("unsupported registry type")
+	ErrUnsupportedRegistryBaseURL   = errors.New("unsupported registry base URL")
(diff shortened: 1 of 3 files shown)


domdomegg:
@claude please review

rdimitrov (commented):
lgtm 👍 

_Note not affecting this PR: I've been meaning to give this a go but I think it would be nice to have the validators as methods for the given structs they are validating, so we can have something like `package.Validate()` or in this case `registryType.Validate()`. If we want to ensure we don't use objects representing data of interest without any validation we easily make that an interface_

PR #352, opened 2025-09-03
Add embedded CLI example

KalleBylin:
Add example for MCP servers embedded inside a CLI tool and update example validator count.

## Motivation and Context
Document how to model CLI-embedded MCP servers using existing packages/arguments, per discussion [#274](https://github.com/modelcontextprotocol/registry/discussions/274). No schema changes.

## How Has This Been Tested?
- Ran validators: schemas valid; 11/11 examples pass.
- Scripts: ./tools/validate-schemas.sh, ./tools/validate-examples.sh

## Breaking Changes
None.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
Startup reference: [Agentic integrations - Snyk MCP Server](https://docs.snyk.io/integrations/developer-guardrails-for-agentic-workflows)

### DIFF
--- a/tools/validate-examples/main.go
+++ b/tools/validate-examples/main.go
@@ -24,7 +24,7 @@ const (
 	// IMPORTANT: Only change this count if you have intentionally added or removed examples
 	// from the examples.md file. This check prevents accidental formatting changes from
 	// causing examples to be skipped during validation.
-	expectedExampleCount = 11
+	expectedExampleCount = 12
 )
 
 func main() {


docs/server-json/examples.md
```diff
@@ -553,6 +553,41 @@ This example shows an MCPB (MCP Bundle) package that:
...
+  "status": "active",
+  "version_detail": {
+    "version": ">=1.1298.0"
+  },
+  "packages": [
+    {
+      "registry_type": "npm",
+      "registry_base_url": "https://registry.npmjs.org",
+      "identifier": "snyk",
+      "version": ">=1.1298.0",
```
connor4312: `version` is a static version, not a range

docs/server-json/examples.md
```diff
@@ -553,6 +553,41 @@ This example shows an MCPB (MCP Bundle) package that:
...
+      "runtime_hint": "npx",
```
connor4312: this should not be necessary -- unless there are specifics to this package that would make it not work outside of npx?

docs/server-json/examples.md
```diff
@@ -553,6 +553,41 @@ This example shows an MCPB (MCP Bundle) package that:
...
+Some CLI tools bundle an MCP server, without a standalone MCP package or a public repository. In these cases, reuse the existing `packages` shape by pointing at the host CLI package and supplying the `package_arguments` and `runtime_hint` if needed to start the MCP server.
+
+```json
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
+  "name": "io.snyk/cli-mcp",
+  "description": "MCP server provided by the Snyk CLI",
+  "status": "active",
+  "version_detail": {
+    "version": ">=1.1298.0"
```
connor4312: This should also be a version number, not a range

docs/server-json/examples.md
KalleBylin: Makes sense from the MCP server perspective, but it does get a bit confusing with the CLI tool versioning. The version of the CLI will keep changing but not necessarily the MCP Server. 

I'll leave it static using the CLI version (rather than a version unique to the MCP server which can make installation confusing) and leave it up to the publisher if they want to set up an update automation so the version here matches the latest CLI version

docs/server-json/examples.md
KalleBylin: Good call, agreed. The server can also be invoked directly from the CLI. I’ve removed runtime_hint to keep it runtime-agnostic.

PR #355, opened 2025-09-04
Check for localhost in remote url

Avish34:
Add check for localhost in remote url.
https://github.com/modelcontextprotocol/registry/discussions/274#discussioncomment-14124511

## Motivation and Context

We don't want authors to publish server with remote url set as localhost.

## How Has This Been Tested?

It has been tested using UTs.

## Types of changes

- [ ] Documentation update

## Checklist


Avish34:
@tadasant / @domdomegg I see we have one testcase which says "allow localhost urls" in validation_test.go. Can you please help clarify what's the expectation here?


`{
			name: "localhost URLs allowed with any namespace",
			serverDetail: apiv0.ServerJSON{
				Name: "com.example/test-server",
				Remotes: []model.Remote{
					{URL: "http://localhost:3000/sse"},
				},
			},
			expectError: false,
		},`

internal/validators/validators_test.go
```diff
@@ -179,6 +179,26 @@ func TestValidate(t *testing.T) {
 			},
 			expectedError: validators.ErrInvalidRemoteURL.Error(),
 		},
+		{
```
domdomegg, line 182: could we also add a test case for `localhost:8080`? Not sure if this is covered right now, and seems good to catch just in case.

domdomegg (approved):
happy to merge as-is and add the test as a follow-up, will hold open for a few hours so you can add the test to this one if you want :)

domdomegg:
@Avish34 I think we probably don't want to enable localhost urls, that test is wrong and can be deleted. I think we'll figure out the latest with how we handle local servers in https://github.com/modelcontextprotocol/registry/pull/345#issuecomment-3246605600, for now we can disallow them here.

internal/validators/validators_test.go
Avish34: Good catch! Added the same.

domdomegg (approved):

domdomegg:
will fix the merge conflicts for this in a sec

PR #359, opened 2025-09-04
output new ServerID after publishing

slimslenderslacks:
When using `mcp-publisher publish ...`, output the new server `ID` if publishing is successful.

## Motivation and Context

It's useful to know this during testing. It means that you don't have to go and search for the new server ID afterwards. Once we have the `v0/servers/{name}/{version}`, I think we could consider removing this again.

## How Has This Been Tested?

I've been using this while I test publishing to locally running registries.

## Breaking Changes

none

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/pkg/api/v0/types.go
+++ b/pkg/api/v0/types.go
@@ -46,8 +46,8 @@ type Metadata struct {
 }
 
 func (s *ServerJSON) GetID() string {
-	if s.Meta != nil && s.Meta.IOModelContextProtocolRegistry != nil {
-		return s.Meta.IOModelContextProtocolRegistry.ID
+	if s.Meta != nil && s.Meta.Official != nil {
+		return s.Meta.Official.ID
 	}
 	return ""
 }


cmd/publisher/commands/publish.go
```diff
@@ -65,26 +65,38 @@ func PublishCommand(args []string) error {
...
 	_, _ = fmt.Fprintln(os.Stdout, "✓ Successfully published")
+	if serverID := id(response); serverID != "" {
+		_, _ = fmt.Fprintf(os.Stdout, "✓ Server Id %s", serverID)
+	}
+
 	return nil
 }
 
-func publishToRegistry(registryURL string, serverData []byte, token string) error {
+func id(response apiv0.ServerJSON) string {
```
rdimitrov: nit: Should we rename this to be more explicit, i.e. `GetServerJSONId` or better make this a method of ServerJSON, so we can have something like `response.GetId()`

cmd/publisher/commands/publish.go
```diff
@@ -65,26 +65,38 @@ func PublishCommand(args []string) error {
...
+	return ""
+}
+
+func publishToRegistry(registryURL string, serverData []byte, token string) (apiv0.ServerJSON, error) {
 	// Parse the server JSON data
 	var serverJSON apiv0.ServerJSON
 	err := json.Unmarshal(serverData, &serverJSON)
 	if err != nil {
-		return fmt.Errorf("error parsing server.json file: %w", err)
+		return serverJSON, fmt.Errorf("error parsing server.json file: %w", err)
```
domdomegg: I think we probably shouldn't return the serverJSON if it didn't actually get published - probably better to return nil in these cases and just return the error to avoid confusion/surprise.

cmd/publisher/commands/publish.go
slimslenderslacks: I added a function to that type.

cmd/publisher/commands/publish.go
slimslenderslacks: I made this a little clearer.

cmd/publisher/commands/publish.go
```diff
@@ -96,27 +101,31 @@ func publishToRegistry(registryURL string, serverData []byte, token string) erro
...
 	// Read response
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		return fmt.Errorf("error reading response: %w", err)
+		return nil, fmt.Errorf("error reading response: %w", err)
 	}
 
 	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("server returned status %d: %s", resp.StatusCode, body)
+		return &serverJSON, fmt.Errorf("server returned status %d: %s", resp.StatusCode, body)
```
rdimitrov:
```suggestion
		return nil, fmt.Errorf("server returned status %d: %s", resp.StatusCode, body)
```

cmd/publisher/commands/publish.go
```diff
@@ -96,27 +101,31 @@ func publishToRegistry(registryURL string, serverData []byte, token string) erro
...
 	}
 
-	return nil
+	if err := json.Unmarshal(body, &serverJSON); err != nil {
+		return &serverJSON, err
```
rdimitrov:
```suggestion
		return nil, err
```

rdimitrov (commented):
Left a few suggestions so we can keep the function return behaviour the same 👍

rdimitrov (commented):
LGTM 👍 Thanks, @slimslenderslacks! 🙏

rdimitrov (approved):
lgtm 👍

rdimitrov:
This would also address - https://github.com/modelcontextprotocol/registry/issues/401

rdimitrov:
@slimslenderslacks - hey, Jim, there are some conflicts with the latest main, can you rebase your changes? Thanks! 👍

slimslenderslacks:
@rdimitrov I just rebased.

PR #366, opened 2025-09-06
Introduce Goreleaser

rdimitrov:
## Motivation and Context

The following PR introduces Goreleaser for the releases of the registry server and the publisher CLI.

**Details:**
- Builds registry and publisher binaries for Linux, macOS, and Windows (amd64 + arm64)
- Leverage GitHub's release process by creating releases directly from GitHub's interface
- Added .goreleaser.yaml configuration
- New .github/workflows/release.yml workflow triggered by GitHub releases
- Updated Docker tagging in deploy workflow (:latest and :semver(the tag) for releases, :main-YYYYMMDD-sha for testing/development)
- Consolidated version handling in main.go files
- Added release documentation

## How Has This Been Tested?

Locally (but we have to test it by creating a release, i.e. v0.0.1 once this gets merged)

## Breaking Changes

Yes, the notion of `latest` for the container image is now different

## Types of changes

- [ ] Documentation update

## Checklist


.github/workflows/deploy.yml
```diff
@@ -36,7 +36,6 @@ jobs:
         with:
           images: ghcr.io/${{ github.repository }}
           tags: |
-            type=raw,value=latest,enable={{is_default_branch}}
```
domdomegg, line 39: Do we want to change the release cadence for the registry itself?

I think I'd probably be keen for continuous releases of the registry itself off the main branch (because we can easily revert etc. as needed, and it makes deployments smaller and easier to monitor).

To be clear, I'd still be for manual releases of the publisher tool in the way you've set up (because it's harder to revert as it gets installed on other people's systems, and people get annoyed if there are too many updates).

rdimitrov: Ah, good point. I was imagining we'll have something like a staging environment where all main builds would automatically be enrolled and then we'll have a production environment where all stable releases would be used? 

Another approach is would it work if I add a rolling "main" tag that follows the latest push to main? In my experience "latest" usually refers to the latest stable release and so I'm concerned this might become confusing.

domdomegg: yeah I'd be happy with a `main` tag 👍

rdimitrov: done 👍

docs/releasing.md
```diff
@@ -0,0 +1,27 @@
+# Release Guide
```
domdomegg, line 1: todo: can we move this to guides/contributing

.github/workflows/deploy.yml
```diff
@@ -36,8 +36,8 @@ jobs:
...
             type=sha,prefix=main-{{date 'YYYYMMDD'}}-,enable={{is_default_branch}}
+            type=raw,value=main,enable={{is_default_branch}}
```
domdomegg, line 40: todo can we update deploy/k8s/registry.go with this

rdimitrov:
@domdomegg - I'll file a follow up for the comments 👍

PR #379, opened 2025-09-08
Clean up schema by simplifying version structure

domdomegg:
## Summary

Simplified the schema structure by:
- Replacing `version_detail.version` with direct `version` field  
- Removing unused `release_date` field from seed data
- Updating all API types, models, and tests to match new structure
- Maintaining backwards compatibility in API responses

This change makes the schema more intuitive and reduces nesting depth.

### DIFF
--- a/internal/api/handlers/v0/servers_test.go
+++ b/internal/api/handlers/v0/servers_test.go
@@ -133,7 +133,7 @@ func TestServersListEndpoint(t *testing.T) {
 						Source: "github",
 						ID:     "example/test-matching",
 					},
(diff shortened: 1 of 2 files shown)


domdomegg:
@claude please review

domdomegg:
(for context, I think the history here was that the registry would add the `is_latest` and `release_date` into version detail. But we moved it into the `_meta` to keep server.json immutable, so we should remove these and flatten to version. Also `release_date` is a duplicate of `published_at`, so we can remove it)

PR #383, opened 2025-09-09
Fix server ID consistency across versions

tadasant:
## Summary

This PR addresses the server ID consistency issues and implements the new version-specific endpoints as outlined in the problem statement.

## Current state

The `["_meta"]["io.modelcontextprotocol.registry/official"]["id"]` field is currently unique per-published-server.json. The notion of `server` in the Registry API could more precisely be described as "A specific version of an MCP server implementation".

This is the ID you currently use to query the `/v0/servers/{id}` endpoint.

### No reliable way to associate different versions as the "same MCP server"

Because ID doesn't stay constant across versions, we have no way to reliably recreate a version history for a server. We could _almost_ rely on `name`, but `name` should be mutable. It's common, for example, for MCP server projects to start as a personal project (e.g. `io.github.tadasant/fetch`) and get graduated to a company project (e.g. `com.pulsemcp/pulse-fetch`), and it should not be required that server maintainers create a new server entry to maintain that version history and the associated downstream privileges (e.g. inclusion in subregistries) that might have previously created. I could also see this being especially common for the case that a `io.github.{username}` hobby project becomes a real project with a proper reverse DNS domain.

### The `/v0/servers/{id}` endpoint is not actually that useful

If a consumer is already aware of the unique ID, presumably it has previously already fetched that `server.json`. It'd be fairly uncommon for a consumer of the Registry to have the ID but not a copy of the server.json.

What we _really_ want `/v0/servers/{id}` to do is fetch either the latest version of a particular server implementation, or perhaps its entire version history.

## Solution

Refactor the notion of `["_meta"]["io.modelcontextprotocol.registry/official"]["id"]` to represent an ID for an MCP server implementation that stretches across versions.

**New endpoints implemented:**

- `GET /v0/servers/{id}` - Serves the _latest_ version (updated behavior)
- `GET /v0/servers/{id}?version=1.0.0` - Serves a specific version (new)
- `GET /v0/servers/{id}/versions` - Serves all versions of that server (new)

### Database Schema Changes

- **Primary Key**: `record_id` (format: `{server_id}_{version}`)
- **Server Consistency**: `server_id` column ensures version consistency
- **Dual Schema Support**: Backwards compatible with old schema during migration

### Service Layer Changes

- Updated `GetByID()` to return latest version by default
- Added `GetByIDAndVersion()` for specific version retrieval
- Added `GetVersionsByID()` for complete version history
- Enhanced database filtering with `ServerID` filter support

### API Changes

- Enhanced `/v0/servers/{id}` with version query parameter support
- Added `/v0/servers/{id}/versions` endpoint
- Updated OpenAPI specification and documentation
- Proper error handling (404 for missing servers/versions)

### Testing

- ✅ All existing tests pass
- ✅ Local development environment tested
- ✅ New endpoints verified with seed data
- ✅ Error handling confirmed (404s, validation)

This is a fairly minimal change that stays adherent with reasonable API design practices and addresses all the concerns raised in the problem statement.

### DIFF
--- a/docs/reference/api/generic-registry-api.md
+++ b/docs/reference/api/generic-registry-api.md
@@ -15,7 +15,9 @@ The official registry has some more endpoints and restrictions on top of this. S
 
 ### Core Endpoints
 - **`GET /v0/servers`** - List all servers with pagination
-- **`GET /v0/servers/{id}`** - Get detailed server information by ID  
+- **`GET /v0/servers/{id}`** - Get detailed server information by ID (returns latest version by default)
+- **`GET /v0/servers/{id}?version={version}`** - Get a specific version of a server
+- **`GET /v0/servers/{id}/versions`** - Get all versions of a server
 - **`POST /v0/publish`** - Publish new server (optional, registry-specific authentication)
 
 ### Authentication


domdomegg:
> The MCP server app.getdialer/dialer had two different IDs for its two versions (1.0.0 and 1.0.1), but the ID field was meant to be consistent across all versions of the same server.

Ah, fwiw I thought the ID was an identifier for a particular server version rather than for a particular server.

What's the use case for having the same ID for different versions, over using the name? Is it to enable renames etc.?

Also just a minor heads up that this might cause weirdness with cursor pagination when listing servers, which you might want to also double check.

rdimitrov:
> Ah, fwiw I thought the ID was an identifier for a particular server version rather than for a particular server.

Yeah, I had the same assumption 👍

slimslenderslacks:
I had been confused about this as well. At one point a few months ago, I thought either Tadas or Toby had said that the `id` identified the server and maybe there was going to be a query parameter for the versions. But in the meantime, the `id` identifies one particular version of a server, right?

formulahendry:
> Ah, fwiw I thought the ID was an identifier for a particular server version rather than for a particular server.

+1. Otherwise, the api of `/v0/servers/{id}` would return an array of ServerDetail, which is not expected...
https://github.com/modelcontextprotocol/registry/blob/72f9f80d680f2e88309f4025106d5f44124b359f/docs/reference/api/openapi.yaml#L39

tadasant:
Closing as https://github.com/modelcontextprotocol/registry/issues/396 had a slightly different conclusion -- will start a new branch

PR #412, opened 2025-09-10
Add Docker Hub to list of supported registries

slimslenderslacks:
## Motivation and Context

We just want to highlight that Docker Hub will support this project with updates to the `docker mcp` cli and the [mcp hub](https://hub.docker.com/mcp) catalog.


domdomegg (approved):
Awesome to hear!

PR #413, opened 2025-09-10
Ban 'latest' as a reserved version string

domdomegg:
## Summary
- Adds validation to prevent servers from being published with version string "latest"
- "latest" is now a reserved keyword that cannot be used as a version string

## Implementation
- Added `ErrReservedVersionString` error constant
- Created `validateVersion` function to check for reserved version strings
- Integrated validation into the package field validation flow
- Added comprehensive test coverage

### DIFF
--- a/internal/validators/constants.go
+++ b/internal/validators/constants.go
@@ -10,6 +10,7 @@ var (
 
 	// Package validation errors
 	ErrPackageNameHasSpaces = errors.New("package name cannot contain spaces")
+	ErrReservedVersionString = errors.New("version string 'latest' is reserved and cannot be used")
 
 	// Remote validation errors
 	ErrInvalidRemoteURL = errors.New("invalid remote URL")
--- a/internal/validators/validators.go
+++ b/internal/validators/validators.go
@@ -72,6 +72,11 @@ func validatePackageField(obj *model.Package) error {
 		return ErrPackageNameHasSpaces
 	}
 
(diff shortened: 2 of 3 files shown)


connor4312:
We may also want to validate that the version is a specific version and doesn't look like a semver version range (prefixed with =, ^, ~, etc)

rdimitrov (approved):
Filed an issue to address @connor4312's comment separately 👍

frenchi:
Just an fyi of a suspected bug: this bans `latest` on `serverDetail.Packages.version` but not `serverDetail.Version`. Fixed over in #435 .
 
Repro case:
```go
{
			name: "server with reserved version 'latest'",
			serverDetail: apiv0.ServerJSON{
				Name:        "com.example/test-server",
				Description: "A test server",
				Repository: model.Repository{
					URL:    "https://github.com/owner/repo",
					Source: "github",
				},
				Version: "latest",
				Packages: []model.Package{
					{
						Identifier:      "test-package",
						RegistryType:    "npm",
						RegistryBaseURL: "https://registry.npmjs.org",
						Version:         "1.0.0",
						Transport: model.Transport{
							Type: "stdio",
						},
					},
				},
			},
			expectedError: validators.ErrReservedVersionString.Error(), 
		},
```

domdomegg:
Good catch @frenchi, thank you! Will review that.

Issue #183, opened 2025-07-13
Introduce notion of "marketing_url" / "homepage"


tadasant:
See conversation: https://github.com/modelcontextprotocol/registry/pull/130#issuecomment-2944736838

> The use case that makes sense to me would be a "landing page URL" like https://zapier.com/mcp.
> 
> We should discuss what the right naming would be for this, as `url` is clearly too ambiguous (that is evident even from just this conversation). Some options:
> - `landing_page_url`
> - `marketing_url`
> - `learn_more_url`
> - `more_information`
> - `canonical_url`
> 
> And maybe we would require this URL to match the reverse DNS verification in `name`.

@jonathanhefner noted:

> Another option is `homepage`.  That's what NPM ([example](https://www.npmjs.com/package/@modelcontextprotocol/sdk)) and RubyGems ([example](https://rubygems.org/gems/mcp)) use, at least when displayed.
> 
> > And maybe we would require this URL to match the reverse DNS verification in `name`.
> 
> This might be too onerous a requirement.  I can imagine someone wanting to switch to a new domain after the fact.  (Gotta grab that new `.ai` TLD! :fire:)

tadasant:
Relevant to this SEP that is landing: https://github.com/modelcontextprotocol/modelcontextprotocol/issues/973#issuecomment-3192076617

We should probably align with naming used there (websiteUrl)

tadasant:
That SEP has landed so this is shovel-ready to implement as websiteUrl

KalleBylin:
@tadasant Opened #422. Ready for review 🙏🏼

tadasant:
Thank you @KalleBylin ! https://github.com/modelcontextprotocol/registry/pull/422 is in

PR #422, opened 2025-09-10
feat: Add website_url field for MCP servers

KalleBylin:
Adds optional website_url field to the server.json schema allowing publishers to provide a url to the server's homepage, documentation, or project website. This provides a central link for users to learn more about the server. Particularly useful when the server has custom installation instructions or setup requirements.

Issue: #183 
PR: #130 
Discussion: #274 

## Motivation and Context
This change addresses the need identified in #130, #183 and #274 for a standardized way to direct users to an official landing page similar to `homepage` used in NPM ([example](https://www.npmjs.com/package/@modelcontextprotocol/sdk)). It is also serves as a viable option for MCP Servers that don't follow typical package manager patterns (e.g., embedded in a Desktop app).

## How Has This Been Tested?
- Schema validation: All 13 example server.json files validate successfully against the updated schema
- Go validator tests: Added additional test cases covering URL format validation and namespace matching
- Run all checks: Updated example files to work with anonymous publishing and verified `make check` passes

## Breaking Changes
None. The website_url field is optional and backward compatible with existing server.json files.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
Security considerations:
- Website URL validation enforces that the domain matches the publisher's reverse-DNS namespace
- HTTP/HTTPS protocols are required (no file:// or other schemes)
- Validation is consistent with existing remote URL validation patterns

### DIFF
--- a/internal/api/router/router.go
+++ b/internal/api/router/router.go
@@ -2,6 +2,8 @@
 package router
 
 import (
+	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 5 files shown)


domdomegg:
Small point of uncertainty is whether we should use snake_case or camelCase. Snake would fit better with the rest of sever.json, camelCase would fit better with MCP spec.

I think I probably lean towards snake case for fitting with server.json (possibly we should have used camelCase here instead generally...) but holding off merging this for a little bit until maybe some others weigh in (@tadasant @rdimitrov maybe?)

internal/validators/validators.go
```diff
@@ -44,6 +49,11 @@ func ValidateServerJSON(serverJSON *apiv0.ServerJSON) error {
 		return err
 	}
 
+	// Validate reverse-DNS namespace matching for website URL
+	if err := validateWebsiteURLNamespaceMatch(*serverJSON); err != nil {
```
domdomegg, line 53: Think it's reasonable to start with this validation, and then we can consider removing it in future if people find it annoying 👍

domdomegg (approved):

rdimitrov:
> Small point of uncertainty is whether we should use snake_case or camelCase. Snake would fit better with the rest of sever.json, camelCase would fit better with MCP spec.
> 
> I think I probably lean towards snake case for fitting with server.json (possibly we should have used camelCase here instead generally...) but holding off merging this for a little bit until maybe some others weigh in (@tadasant @rdimitrov maybe?)

If we want to consider doing it I think there's literally no better time than now to make the switch to camelCase 😃 If not, I agree that it should remain aligned with what the rest of the server.json follows 👍

internal/validators/validators.go
KalleBylin: Yeah, my thought exactly. Feels easier to relax this restriction rather than enforcing it afterwards 😅

tadasant:
Good thing we are in Preview 😅 

I think this is fine to merge as website_url. We are already inconsistent in other places (see `maxLength`). Issue to track separately (and agree we should action ASAP 🙁 ): https://github.com/modelcontextprotocol/registry/issues/428

PR #424, opened 2025-09-10
Check for version and identifier for npm packages

Avish34:
Adding validation for version and identifier for packages.

## Motivation and Context
While debugging a bug, I found out that we need exact version to be there in package to fetch from npm, otherwise we won't be able to parse the response as it returns all versions. Even if we do, we still need to go and figure out which version have the mcpname associated with while iterating over all versions. So, we should make mandatory to have identifier and version present in the given package object if we want to validate package association with registries.

## How Has This Been Tested?

UTs.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/validators/registries/npm.go
+++ b/internal/validators/registries/npm.go
@@ -22,6 +22,18 @@ func ValidateNPM(ctx context.Context, pkg model.Package, serverName string) erro
 		pkg.RegistryBaseURL = model.RegistryURLNPM
 	}
 
+	if pkg.Identifier == "" {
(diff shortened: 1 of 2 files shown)


Avish34:
@domdomegg / @tadasant please review!

domdomegg (approved):
Thank you! I think we should probably have similar validation on some of the other package types - any chance you'd be up for also looking at those? (e.g. PyPi, Nuget, ...)

Avish34:
@domdomegg - Happy to take it up!!

Issue #415, opened 2025-09-10
Validate that the version is a specific version and doesn't look like a semver version range


rdimitrov:
We may also want to validate that the version is a specific version and doesn't look like a semver version range (prefixed with =, ^, ~, etc)

_Originally posted by @connor4312 in https://github.com/modelcontextprotocol/registry/issues/413#issuecomment-3275742954_

domdomegg:
I think I'd be supportive of this if we check nobody has done this in the existing registry data (if they have, we should explore and understand why people have done this / if it's valid etc.).

Easy to remove the restriction later if it does block valid workflows.

jonathanhefner:
I am curious why we would not simply treat these as non-semantic version numbers?  The [BNF for valid semantic versions](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) is already limited to specific version numbers; range qualifiers such as `^` are defined by package managers (and can vary between package managers).  And our documentation already [defines how non-semantic versions are handled](https://github.com/modelcontextprotocol/registry/blob/773f09441ad8d9ce43c34957a7fb21d02e4bed5d/docs/reference/faq.md?plain=1#L83):

> The registry attempts to parse versions as semantic versions for proper ordering. Non-semantic versions are allowed but will be ordered by publication timestamp.

If using ranges is a common mistake, we could perhaps add a warning / confirmation in the CLI.  Is it a common mistake though?

domdomegg:
I think it makes it likely that clients implementing things like npm packages accidentally install a version range, rather than a specific version. Which makes it harder for subregistries to provide guarantees about specific packages. I think it's very unlikely people really intend to submit a semver version range a lot of the time as an intentional version string - and we could lift this restriction if there's very good reason people are doing this.

jonathanhefner:
> I think it makes it likely that clients implementing things like npm packages accidentally install a version range, rather than a specific version. Which makes it harder for subregistries to provide guarantees about specific packages.

My understanding was that this restriction was primarily for the server version (used by the registry itself), similar to the recent restriction on using `latest`.

With regards to restricting actual package version numbers, I think we should delegate that responsibility to the package managers themselves.  For example, npm [already enforces restrictions](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#version).

ossamalafhel:
Hi team,

[...]

PR #435, opened 2025-09-11
feat(validators): Validate that Server versions are not ranges

frenchi:
Validates that server versions are not ranges.

**NB:** This also introduces `validateVersion` as a check on the top-level `ValidateServerJSON`, as previously we were only checking version ranges of packages via `validatePackageField` (i.e. `latest` would have still been a valid top level server version), despite #413's wording suggesting otherwise.

## Motivation and Context
Solves https://github.com/modelcontextprotocol/registry/issues/415

## How Has This Been Tested?

Tests added in `internal/validators/validators_test.go`

`make check` is clean.

## Breaking Changes
As per @domdomegg's comment, we should
> check nobody has done this in the existing registry data (if they have, we should explore and understand why people have done this / if it's valid etc.).

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/data/seed.json
+++ b/data/seed.json
@@ -21,7 +21,9 @@
         "registry_type": "npm",
         "identifier": "@21st-dev/magic",
         "registry_base_url": "https://registry.npmjs.org",
-        "transport": {"type": "stdio"}
+        "transport": {
+          "type": "stdio"
+        }
       }
     ],
     "_meta": {
@@ -80,7 +82,9 @@
         "registry_type": "pypi",
         "identifier": "adfinmcp",
         "registry_base_url": "https://pypi.org",
-        "transport": {"type": "stdio"}
+        "transport": {
+          "type": "stdio"
+        }
       }
     ],
     "_meta": {
@@ -114,7 +118,9 @@
         "registry_type": "npm",
         "identifier": "agentql-mcp",
         "registry_base_url": "https://registry.npmjs.org",
-        "transport": {"type": "stdio"}
+        "transport": {
+          "type": "stdio"
+        }
       }
     ],
     "_meta": {
@@ -138,7 +144,7 @@
     "version": "0.0.1-seed",
     "packages": [
       {
-        "version": "",
+        "version": "0.0.16",
         "package_arguments": [
           {
             "description": "Positional argument",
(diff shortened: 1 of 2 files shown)


internal/validators/validators.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"regexp"
```
rdimitrov, line 8: We have to try this but I think the semver library we already use (from the core x package) may work instead of the regex approach (there's a IsValid method that I think fails on ranges).

internal/validators/validators.go
```diff
@@ -135,16 +167,53 @@ func validatePackageField(obj *model.Package) error {
...
+	if orRangeRe.MatchString(trimmed) {
+		return true
+	}
+	if dottedVersionLikeRe.MatchString(trimmed) {
+		// wildcard in a dotted version (x/X/*) implies range-like intent
+		return strings.Contains(trimmed, "x") || strings.Contains(trimmed, "X") || strings.Contains(trimmed, "*")
+	}
+	return false
+}
+
```
frenchi, line 216:
```suggestion
// looksLikeVersionRange detects common semver range syntaxes and wildcard patterns.
// Examples that should return true:
// - "^1.2.3",
// - "~1.2.3",
// - ">=1.0.0",
// - "1.x",
// - "1.2.*",
// - "1 - 2",
// - "1.2 || 1.3"
func looksLikeVersionRange(version string) bool {
	s := strings.TrimSpace(version)
	if s == "" {
		return false
	}

	if isComparatorRange(s) || isHyphenRange(s) || isOrRange(s) || isWildcardDotted(s) {
		return true
	}

	return false
}

func isComparatorRange(s string) bool {
	for _, cmp := range []string{"^", "~", ">=", "<=", ">", "<", "="} {
		if strings.HasPrefix(s, cmp) {
			rhs := strings.TrimSpace(strings.TrimPrefix(s, cmp))
			return semver.IsValid(v(rhs))
		}
	}
	return false
}

func isHyphenRange(s string) bool {
	// exact “a - b” (one hyphen with spaces)
	const sep = " - "
	if !strings.Contains(s, sep) {
		return false
	}
	parts := strings.Split(s, sep)
	if len(parts) != 2 {
		return false
	}
	return semver.IsValid(v(parts[0])) && semver.IsValid(v(parts[1]))
}

func isOrRange(s string) bool {
	if !strings.Contains(s, "||") {
		return false
	}
	parts := strings.Split(s, "||")
	semverSides := 0
	for _, p := range parts {
		if semver.IsValid(v(strings.TrimSpace(p))) {
			semverSides++
			if semverSides >= 2 {
				return true
			}
		}
	}
	return false
}

func isWildcardDotted(s string) bool {
	// quick dotted-shape check, then look for wildcard segment
	if !strings.Contains(s, ".") {
		return false
	}
	for _, seg := range strings.Split(s, ".") {
		seg = strings.TrimSpace(seg)
		if seg == "x" || seg == "X" || seg == "*" {
			return true
		}
	}
	return false
}

// v adds a "v" prefix if not present
func v(version string) string {
	if !strings.HasPrefix(version, "v") {
		return "v" + version
	}
	return version
}

```

internal/validators/validators.go
frenchi: Thanks for the reply @rdimitrov 

So, I intially started by looking at our [IsSemanticVersion(version string) bool](https://github.com/modelcontextprotocol/registry/blob/773f09441ad8d9ce43c34957a7fb21d02e4bed5d/internal/service/versioning.go#L13) & semver's `IsValid(v string) bool`.

`isValid` requires a `v` prefix, so to take that approach we'd effectively be duplicating `ensureVPrefix` from `versioning.go` over here (done below as `func v`). Importing the existing `IsSemanticVersion` in to `validators.go` would cause a cyclic dependency via `registry_service.go`. This would be solvable by a bigger refactor, but I figured that was out of scope for what should be a small PR.

Also, given the requirement that version is `string[255]` from issue #158 (i.e. we only SHOULD, not MUST, match semver), I tried to hack around a more general string case, and regex was the cleanest thing I landed on.

An alternative approach with `IsValid` would look something like this: https://github.com/modelcontextprotocol/registry/pull/435/files#r2340817609

Happy to make that change if that's preferable.

rdimitrov: I was thinking along the lines of something like this - https://go.dev/play/p/ZhBV3r5L-pe 
_I copy-pasted a few of the tests and asked Claude to extract the cases and it seems to work overall._

docs/reference/server-json/server.schema.json
```diff
@@ -105,7 +105,8 @@
           "type": "string",
           "description": "Package version",
           "example": "1.0.2",
-          "minLength": 1
+          "minLength": 1,
+          "remarks": "Must be a specific version. Version ranges are rejected (e.g., '^1.2.3', '~1.2.3', '>=1.2.3', '1.x', '1.*')."
```
domdomegg, line 109: Are remarks a thing in JSON schema? Should this be added to the description instead?

internal/validators/validators_test.go
```diff
@@ -17,6 +17,160 @@ func TestValidate(t *testing.T) {
 		serverDetail  apiv0.ServerJSON
 		expectedError string
 	}{
```
domdomegg, line 19: do we have a test case for non semver versions? (answer may be yes already! but just want to check)

rdimitrov:
> Think it's reasonable to do with regex - I think we can't just use the semver package here because we also want to support non-semver-versions, and just bad semver-version-ranges.

Got it, it all makes sense then 👍 Just curious but was there a specific use case for deciding to support things other than semver?

domdomegg:
> Just curious but was there a specific use case for deciding to support things other than semver?

I think mainly to enable flexibility here because there are a lot of ecosystems/packages that don't regularly use semver, that we didn't want to constrain. I think there should be more discussion of this somewhere when we were discussing adding versioning validation. (@claude maybe you can find this issue / collate discussions here?). We could revisit this, although I think it's orthogonal for now to disallowing semver version ranges.

PR #436, opened 2025-09-11
Chore: Dependency pinning & Depdendabot introduction

frenchi:
This PR does three things all related to depdency/vulnerability management and housekeeping.

Happy to split them up if we only want a subset of the changes.

1. Pins the Github Action workflow hashes under `.github/workflows` to full-length commit SHAs.
2. Adds a base level `.github/dependabot.yml` config to keep these (and other go/Docker) deps up to date.
3. Introduces the `make pin-deps` helper which uses https://github.com/santrancisco/pmw to manually manage pinned workflows.

## Motivation and Context
- Github's [security guide](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions) recommends using the immutable SHA of an action when consuming third party actions.
- Depdendabot makes ongoing dep house keeping manageable.
- `pmw` introduces a depdency - but it's convenience outweighs the low risk, imo.

## How Has This Been Tested?
N/A

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+  - package-ecosystem: github-actions
+    directory: /
+    groups:
+      actions:
+        patterns:
+          - "*"
+    schedule:
+      interval: daily
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+        interval: "daily"


domdomegg:
This looks great! Similar to @rdimitrov I think I'd prefer just pin the actions manually for now, and avoid adding pmw.

Equally happy with dependabot or renovate tbh. If we use renovate, then I'd have a preference for it running in a GitHub action rather than installed as a special priviledged app, if possible? This also enables more configuration/infra as code.

frenchi:
Thanks for the feedback @rdimitrov & @domdomegg. Removed pmw.

> PS. btw I'm one of the maintainers of https://github.com/stacklok/frizbee but I still feel Renovate is a better all-around tool if we want to automate most of the dependency management in one place.


Ah! Frizbee is great, thanks for the tool. I thought your name was familiar :) We might have met at an OpenSSF event.

If we use renovate, then I'd have a preference for it running in a GitHub action rather than installed as a special priviledged app, if possible?

Agreed that Renovate offers a superset of features. I have a very small preference for keeping deps minimal & github is already trusted.

Running [renovate as an github action](https://github.com/renovatebot/github-action) is possible, but requires still self-hosted renovate as a backend. The setup guide for [self-hosted Renovate](https://github.com/mend/renovate-ce-ee/blob/main/docs/setup-for-github.md) looks doable at a glance, but a bit more involved than this PR (and also still requires a privileged app, just one that we would create).

I'll leave this one as dependabot for now, unless anyone has a strong preference for renovate.

_Edit: Looks like community edition is limited at 10 repos, so this could be a blocker in future, if we wanted a consistent way to do manage deps across the org.Edit: Looks like community edition is limited at 10 repos, so this could be a blocker in future, if we wanted a consistent way to manage deps across the org._

rdimitrov:
> Ah! Frizbee is great, thanks for the tool. I thought your name was familiar :) We might have met at an OpenSSF event.

Same! :smiley: We should definitely meet next time, I would be happy to chat :smiley:  

Regarding Renovate - that's a shame, I was hoping it's similarly easy to install as Dependabot. 

_edit: Apologies for the double notifications but instead of clicking `quote reply` I clicked `edit` on the comment below and accidentally had to revert it_

PR #439, opened 2025-09-11
Add GitHub Container Registry (GHCR) support for MCP servers

### LINKED ISSUE #393: Add ghrc as a docker registry

I haven't tried this first hand, this is just based off of reviewing the docs. I see that https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/official-registry-requirements.md states this in the support registries section:
> Docker/OCI: https://docker.io only

I take this to mean that docker.io (Docker Hub) is the only supported registry for container images. From my past experience, I think this will be a problem because Docker Hub imposes pretty severe rate limiting on images. Once you hit that, you have to start paying for users to be able to download your images.

Longer term, I think the list of supported registries should grow substantially, but in the short term, I think you would give folks a lot more flexibility by adding GitHub's `ghcr.io` as a supported OCI registry.

Alternatively, we could lift all restrictions on the registry, but I assume you want to keep it somewhat locked down to know providers.

We could also come out of the gate with a longer list of support registries including GCR, ACR, Quay.io, Gitlab's container registry, etc, but simply adding ghcr.io seemed like the best bang for the buck.

Gaurav-Gosain:
This commit implements support for publishing MCP servers to GitHub Container
Registry (ghcr.io) in addition to the existing Docker Hub support, addressing
Docker Hub rate limiting concerns as requested in #393.

Key changes:
- Add RegistryURLGHCR constant for https://ghcr.io
- Refactor OCI validation to support multiple container registries with extensible RegistryConfig system
- Implement registry-agnostic authentication system that handles:
  - Docker Hub token-based authentication via auth.docker.io
  - GHCR token-based authentication via ghcr.io/token service
- Update Accept headers to support OCI image indexes for multi-arch images
- Extract helper functions to reduce cyclomatic complexity and improve code maintainability
- Add comprehensive test coverage for both Docker Hub and GHCR scenarios
- Include real image testing with github/github-mcp-server and nkapila6/mcp-local-rag
- Update publishing documentation with GHCR examples and authentication details

The implementation maintains backward compatibility with existing Docker Hub
usage while providing a future-proof architecture for adding additional
container registries. All authentication is now handled through proper token
services for both registries.

Fixes #393

## Motivation and Context
Docker Hub rate limiting has been causing issues for the MCP registry when validating OCI images. Adding GHCR support provides an alternative container registry that many MCP server authors are already using, reducing the load on Docker Hub and providing more options for the community.

## How Has This Been Tested?
- All existing unit tests continue to pass
- Added comprehensive test coverage for GHCR scenarios including:
  - Non-existent images (proper error handling)
  - Real images without MCP annotations (github/github-mcp-server)  
  - Real images with proper MCP annotations (nkapila6/mcp-local-rag)
  - Registry URL validation for both supported registries
- Tested authentication flow manually using curl commands
- Verified multi-arch image support with OCI index manifests

## Breaking Changes
None. This is a purely additive change that maintains full backward compatibility with existing Docker Hub configurations.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
The implementation uses a RegistryConfig struct to abstract registry-specific details like API endpoints, authentication URLs, and token services. This design makes it straightforward to add support for additional container registries in the future (such as AWS ECR, Azure Container Registry, etc.) by simply adding new cases to the getRegistryConfig function.

The authentication system properly handles the different token services used by each registry:
- Docker Hub: auth.docker.io with service=registry.docker.io
- GHCR: ghcr.io/token with service=ghcr.io

Both registries now support the full OCI specification including multi-arch images via OCI image indexes.


internal/validators/registries/oci.go
```diff
@@ -14,13 +14,44 @@ import (
...
+		return &RegistryConfig{
+			APIBaseURL: dockerIoAPIBaseURL,
+			AuthURL:    "https://auth.docker.io/token",
+			Service:    "registry.docker.io",
+			Scope:      fmt.Sprintf("repository:%s/%s:pull", namespace, repo),
+		}
+	case model.RegistryURLGHCR:
+		return &RegistryConfig{
+			APIBaseURL: ghcrAPIBaseURL,
+			AuthURL:    "https://ghcr.io/token",
```
rdimitrov: Shouldn't this be based on ghcrAPIBaseURL, i.e. `fmt.Sprintf("%s/%s", ghcrAPIBaseURL, "token")`?

Gaurav-Gosain: Good catch! You're absolutely right. Fixed in 721faff.

internal/validators/registries/oci.go
```diff
@@ -59,75 +89,111 @@ func ValidateOCI(ctx context.Context, pkg model.Package, serverName string) erro
...
+	}
+
+	return &manifest, nil
+}
+
+// getConfigDigestFromManifest extracts the config digest from an OCI manifest
+func getConfigDigestFromManifest(ctx context.Context, client *http.Client, registryConfig *RegistryConfig, namespace, repo string, manifest *OCIManifest) (string, error) {
+	// Handle rate limited case (empty manifest)
+	if len(manifest.Manifests) == 0 && manifest.Config.Digest == "" {
+		return "", nil // Rate limited case, skip validation
```
rdimitrov: Could we have the rate limiting case handled more explicitly? My concern is that a broken or empty manifest reply from a registry would skip validation whereas my understanding is this should be the case only if we are being rate limited? (@domdomegg - which is probably also not ideal but still)

Gaurav-Gosain: That makes sense, I've implemented a proper fix using an explicit `ErrRateLimited` error type. Now rate limiting is detected only via HTTP 429 status codes, while genuinely empty/broken manifests cause validation to fail with clear error messages. Fixed in 9a9783f.

Gaurav-Gosain:
> @Gaurav-Gosain - Thank you very much for addressing this! 💯
> 
> I like that it was implemented in a way that makes it easy to extend in the future. I left a small comment, but overall it looks great 👍

Thank you for the positive feedback 🙏!

The implementation now properly separates concerns and ensures that only explicit rate limiting (HTTP 429) results in skipped validation, while broken or empty responses properly fail validation.

rdimitrov (approved):
It looks good from my side. I'll wait for the others to have a look at it too before we can merge it 👍

docs/guides/publishing/publish-server.md
```diff
@@ -238,18 +238,36 @@ LABEL io.modelcontextprotocol.server.name="io.github.username/server-name"
...
+### Example server.json (GitHub Container Registry)
 ```json
 {
   "name": "io.github.username/server-name", 
   "packages": [
     {
       "registry_type": "oci",
+      "registry_base_url": "https://ghcr.io",
       "identifier": "yourusername/your-mcp-server",
       "version": "1.0.0"
```
SamYuan1990, line 272: is version here means image tag?

SamYuan1990: btw, any changes to github action approach? can github action approach been automatic?

Gaurav-Gosain: Hi @SamYuan1990, yes `version` here refers to the image `tag`. The existing Docker Hub implementation also referred to it as `version` so wanted to keep it consistent and not introduce any breaking changes.

I have not looked at the GitHub actions tbh, do you have any suggestions?

SamYuan1990: I suppose Github action doc can be updated, but fully support with automation seems other story.
https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/github-actions.md

I suppose maybe we can ask LLM to see if we can just use some github action default env values. as $repo or something? to build the json file.

nkapila6: Don't understand your question. 

Is it for how to make the yaml file to publish on ghcr? Have a look at the repo readme, they clearly outline what's required.

For an example, look here: https://github.com/nkapila6/mcp-local-rag/blob/main/.github%2Fworkflows%2Fpublish-image.yml#L31

SamYuan1990: ref https://docs.github.com/en/actions/reference/workflows-and-actions/variables
follow show me the code
the json file seems to be:

---
```json
{
  "name": "io.github.$GITHUB_ACTION_REPOSITORY", 
  "packages": [
    {
      "registry_type": "oci",
      "registry_base_url": "https://ghcr.io",
      "identifier": "$GITHUB_ACTION_REPOSITORY",
      "version": "container_version"
}
```

**Note** for container version your can figure out from $GITHUB_REF or hardcode.

---

I hope my understand is correct as $GITHUB_ACTION_REPOSITORY in Github action is repo name, and identifier is your container package name which same as your repo name?
so... when we using github action to build and upload a container into ghcr.... we can one step further as how to generate this json file from github action env. (maybe a shell script?)

rdimitrov:
@toby @tadasant @domdomegg - hey, when you have time can you have a look at this so we can move forward with it 👍

tadasant (approved):
Thank you this looks good to me!

Flagging likely conflicts with https://github.com/modelcontextprotocol/registry/pull/485 - but this PR has been around much longer so let's start here.

pab1it0:
@Gaurav-Gosain @tadasant @rdimitrov Deploy failed:  https://github.com/modelcontextprotocol/registry/actions/runs/17807752019/job/50624702508

rdimitrov:
@pab1it0 - Thank you! This has already been flagged so hopefully it should be addressed soon 👍

tadasant:
FYI folks can follow along on the incident here: https://discord.com/channels/1358869848138059966/1369487942862504016/1417633144290414602

PR #457, opened 2025-09-12
chore: remove outdated version tags from GitHub Actions

domdomegg:
## Summary
- Removes version comment tags (e.g. `#v4`, `#v5`) from all GitHub Actions workflow files
- These tags become incorrect when Dependabot updates the action SHAs
- The SHA pins are sufficient for version control

## Context
This addresses @domdomegg's comment in #453 about these tags becoming outdated when using Dependabot. The SHAs already provide version pinning, making the comment tags redundant and potentially misleading.

## Changes
- Removed version tags from all `uses:` statements in `.github/workflows/*.yml`
- No functional changes to workflows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,15 +16,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout code
-      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v4
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
 
     - name: Set up Go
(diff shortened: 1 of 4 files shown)


domdomegg:
Dependabot doesn't keep these up to date sadly - e.g. it bumped actions to v5 but left the comment as v4. Agree I'd prefer to keep them if possible and they can stay synced, but more confusing if they're misleading and unsynced.

Ideally I'd want to do something like docker tags where you can do thing:v5@sha256:abcdef, but I don't think there's an equivalent syntax for github actions

rdimitrov:
I think you're right (ref. https://github.com/dependabot/dependabot-core/issues/8011). I have noticed in the past times when it didn't worked as expected but I was assuming it works at least 50% of the time 😄 Anyway, given it's not to be trusted happy to proceed 👍

Issue #396, opened 2025-09-09
Server ID should not be a primary key / should stay consistent across versions of the same server


tadasant:
See https://discord.com/channels/1358869848138059966/1369487942862504016/1414790523239268412

> I think we have a bug in how we're handling the ["io.modelcontextprotocol.registry/official"]["id"] field -- The MCP server app.getdialer/dialer has two different ID's for its two versions, but the intent of that ID field was to keep ID's consistent across versions. Will draw up a ticket unless I missed something re: intent of IDs

> I think the issue is that we're using that field as a primary key, but the PK should be the server ID + the version (or just a separate generated ID)

> Starting pushing on this with CC but there's a few layers to it, plus we should set up a migration for the existing data (I think as long as we fix it soon we can rely on name to be consistent and just rewrite the ID's to match the earliest submitted server.json's entry). I'm a little jammed up the next ~24h so if someone wants to tackle it please feel free

Draft: https://github.com/modelcontextprotocol/registry/pull/383

Example of a problem:

```json
{
      "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
      "name": "app.getdialer/dialer",
      "description": "An MCP server that provides your you make outbound phone calls using your own phone number",
      "status": "active",
      "repository": {
        "url": "",
        "source": ""
      },
      "version": "1.0.0",
      "remotes": [
        {
          "type": "streamable-http",
          "url": "https://getdialer.app/mcp"
        },
        {
          "type": "sse",
          "url": "https://getdialer.app/sse"
        }
      ],
      "_meta": {
        "io.modelcontextprotocol.registry/official": {
          "id": "04f25b17-331d-48d7-b08b-37baece208a7",
          "published_at": "2025-09-08T23:47:09.452137971Z",
          "updated_at": "2025-09-09T00:16:49.162620072Z",
          "is_latest": false
        }
      }
    },
    {
      "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
      "name": "app.getdialer/dialer",
      "description": "An MCP server that provides lets you make outbound phone calls using your own phone number",
      "status": "active",
      "repository": {
        "url": "",
        "source": ""
      },
      "version": "1.0.1",
      "remotes": [
        {
          "type": "streamable-http",
          "url": "https://getdialer.app/mcp"
        },
        {
          "type": "sse",
          "url": "https://getdialer.app/sse"
        }
      ],
      "_meta": {
[...]

Issue #428, opened 2025-09-11
Refactor all server.json snake_case fields to camelCase


tadasant:
To align with the MCP spec: https://github.com/modelcontextprotocol/registry/pull/422#issuecomment-3276371588

domdomegg:
Probably list of things to do:
- update the spec. probably needs a version bump given breaking change. also update the default $schema value and upload to static repo
- update code implementation
- update docs
- migration to update all historical data in database
- cut new release of publisher cli
- announce this change to people. maybe it would be ideal if we could provide a command to automigrate the old format to the new one to avoid thrash? or do this automatically in the publisher cli on publish for a short while (e.g. 7 days) with a big noisy warning? this is definitely optional though, only if it's easy and neat.

pree-dew:
@domdomegg Can I pick this issue?

domdomegg:
@pree-dew yeah sure go for it! give me a poke when/if you need me to cut a new release or do any other actions :)

tadasant:
@pree-dew have you started on this? I know you've been into some other things so no worries if not

We want to land this breaking change and https://github.com/modelcontextprotocol/registry/pull/465 at the same time, so whoever takes this on should branch off of that

If you're able to get a PR up for this in the next ~24h please go for it, that would be great. If not, I might take a swing at this tomorrow if you don't mind

rdimitrov:
I can help out too in my morning time unless someone been me to it 👍

pree-dew:
@tadasant I will try to come with the PR in next 24h, if in any case I get stuck and it will take time then I am happy if you or @rdimitrov can take over. In any case it is important that work should be shipped at the time that we are expecting it to be, whoever is handling it, so no problem at all :)

rdimitrov:
@pree-dew - sounds great 👍 If you don't mind reply here when you are about to start working on it so we don't accidentally overlap 😃

pree-dew:
@rdimitrov I have started working on this.

pree-dew:
@tadasant Raised a PR for the change https://github.com/modelcontextprotocol/registry/pull/488

From the todo list, migration part of existing record is pending, I am working on it. 

One point that I want to ask is `_meta` fields still has snake_case variables like `is_latest`, `published_at` etc as part of server api response while other fields are in camelCase (after the change). These are internal fields but it looks like inconsistency. Let me know your thoughts on this, currently I have not made changes for `_meta` fields.

tadasant:
[...]

PR #465, opened 2025-09-12
Introduce serverId, rename id to versionId

tadasant:
Wielded Claude Code to do this; across many iterations and have reviewed all the code + largely rewritten the PR commentary here.

This PR implements the server ID consistency changes from [#396](https://github.com/modelcontextprotocol/registry/issues/396). Our goal is to:
- Rename `id` to `versionId` to make it clear that it is the ID for a specific version, not a server
- Introduce `serverId` as a net new concept. This will be 1:1 with `name`. While not strictly necessary (we could just use name as a PK), the convention of having a UUID as an ID on our objects is just a helpful subjective architecture decision. Plus also leaves open the door to reconsidering the "should we allow renames" question.

### 🗄️ Database Schema Changes
- **Renamed**: `id` column → `version_id` (unique identifier for each server version)

### 📋 Metadata Structure Updates
- **JSON Schema**: Updated `_meta.io.modelcontextprotocol.registry/official` to include:
  - `serverId`: Consistent ID across all versions of a server
  - `versionId`: Unique ID for this specific version (renamed from `id`)

### 🔧 API Endpoint Updates
- **New endpoint structure**: Changed from version-ID-based to server-ID-based URLs:
  - `GET /v0/servers/{server_id}` - Get latest version of server
  - `GET /v0/servers/{server_id}?version=X.X.X` - Get specific version  
  - `GET /v0/servers/{server_id}/versions` - List all versions of a server
  - `PUT /v0/servers/{server_id}?version=X.X.X` - Edit specific version
- **Consistent URL patterns**: All server endpoints now use `server_id` as the primary identifier
 
### Database Migration Strategy
The migration (`005_add_server_id_rename_version_id.sql`):
1. Adds new `version_id` column  
2. Generates deterministic UUIDs for servers with the same name
3. Updates JSON metadata to add `serverId` and rename `id` to `versionId`
4. Copies existing `id` values to `version_id` column
5. Drops old `id` column and makes `version_id` the new primary key

## Testing

Besides of course automated tests passing, I:
1. Set up my local db with the pre-005-migration state, imported all the production data
2. Ran the migration
3. Spot checked some diffs of the data that resulted --

```json
// before
{
  "name": "io.github.ruvnet/ruv-swarm",
  "_meta": {
      "id": "0a9cda7f-2b73-4839-844b-6582f335de42",
      "is_latest": false,
      "updated_at": "2025-09-10T17:21:34.92137012Z",
      "published_at": "2025-09-10T17:15:32.69690433Z"
    }
  },
  "version": "1.0.18",
  "repository": {
    "url": "",
    "source": ""
  },
  "description": "Neural network swarm orchestration with WebAssembly acceleration and MCP integration"
}

// after
{
  "name": "io.github.ruvnet/ruv-swarm",
  "_meta": {
      "serverId": "e8538952-96af-4a33-9e31-7fbed012521e",
      "is_latest": false,
      "versionId": "0a9cda7f-2b73-4839-844b-6582f335de42",
      "updated_at": "2025-09-10T17:21:34.92137012Z",
      "published_at": "2025-09-10T17:15:32.69690433Z"
    }
  },
  "version": "1.0.18",
  "repository": {
    "url": "",
    "source": ""
  },
  "description": "Neural network swarm orchestration with WebAssembly acceleration and MCP integration"
}

```

Servers that have multiple versions stayed consistent in their `serverId`:

```json
// First version
{
  "name": "app.getdialer/dialer",
  "_meta": {
      "serverId": "80f08bdb-f3e4-4f19-b072-1676ae49b663",
      "is_latest": false,
      "versionId": "04f25b17-331d-48d7-b08b-37baece208a7",
      "updated_at": "2025-09-09T00:16:49.162620072Z",
      "published_at": "2025-09-08T23:47:09.452137971Z"
    }
  },
  "status": "active",
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
  "remotes": [
    {
      "url": "https://getdialer.app/mcp",
      "type": "streamable-http"
    },
    {
      "url": "https://getdialer.app/sse",
      "type": "sse"
    }
  ],
  "version": "1.0.0",
  "repository": {
    "url": "",
    "source": ""
  },
  "description": "An MCP server that provides your you make outbound phone calls using your own phone number"
}

// Second version
{
  "name": "app.getdialer/dialer",
  "_meta": {
      "serverId": "80f08bdb-f3e4-4f19-b072-1676ae49b663",
      "is_latest": true,
      "versionId": "faa26cb9-3aad-415e-aece-f715fff5fe5e",
      "updated_at": "2025-09-09T00:16:49.107842808Z",
      "published_at": "2025-09-09T00:16:49.107842808Z"
    }
  },
  "status": "active",
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
  "remotes": [
    {
      "url": "https://getdialer.app/mcp",
      "type": "streamable-http"
    },
    {
      "url": "https://getdialer.app/sse",
      "type": "sse"
    }
  ],
  "version": "1.0.1",
  "repository": {
    "url": "",
    "source": ""
  },
  "description": "An MCP server that provides lets you make outbound phone calls using your own phone number"
}
```

And I also tried publishing a test server anew; looked good:

```bash
⏺ Bash(curl -s "http://localhost:8080/v0/servers/102dc4e4-6ca8-40e6-aa11-cd45763d3352" | jq '.')
  ⎿  {                                                                                             
       "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
       "name": "io.modelcontextprotocol.anonymous/test-claude-migration-server",
       "description": "A test server to verify the new migration and API endpoints work correctly",
       "status": "active",
       "repository": {
         "url": "https://github.com/example/test-claude-migration-server",
         "source": "github",
         "id": "example/test-claude-migration-server"
       },
       "version": "1.0.0",
       "_meta": {
           "serverId": "102dc4e4-6ca8-40e6-aa11-cd45763d3352",
           "versionId": "48c9cd55-7bc4-4ac2-896e-59e81eb653e4",
           "published_at": "2025-09-12T18:33:23.264203-07:00",
           "updated_at": "2025-09-12T18:33:23.264203-07:00",
           "is_latest": true
         }
       }
     }

⏺ Bash(curl -s "http://localhost:8080/v0/servers/102dc4e4-6ca8-40e6-aa11-cd45763d3352/versions" | jq '.')
  ⎿  {                                                                                                 
       "servers": [
         {
           "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
           "name": "io.modelcontextprotocol.anonymous/test-claude-migration-server",
           "description": "A test server to verify the new migration and API endpoints work correctly",
           "status": "active",
           "repository": {
             "url": "https://github.com/example/test-claude-migration-server",
             "source": "github",
             "id": "example/test-claude-migration-server"
           },
           "version": "1.0.0",
           "_meta": {
               "serverId": "102dc4e4-6ca8-40e6-aa11-cd45763d3352",
               "versionId": "48c9cd55-7bc4-4ac2-896e-59e81eb653e4",
               "published_at": "2025-09-12T18:33:23.264203-07:00",
               "updated_at": "2025-09-12T18:33:23.264203-07:00",
               "is_latest": true
             }
           }
         }
       ],
       "metadata": {
         "count": 1
       }
     }
```

Co-Authored-By: Claude <noreply@anthropic.com>


slimslenderslacks:
@tadasant the publisher command is writing out the "Server Id" after a successful publish, but it's the versionId.

https://github.com/modelcontextprotocol/registry/blob/main/cmd/publisher/commands/publish.go#L75

Maybe a good idea to write them both now.

tadasant:
> @tadasant the publisher command is writing out the "Server Id" after a successful publish, but it's the versionId.
> 
> https://github.com/modelcontextprotocol/registry/blob/main/cmd/publisher/commands/publish.go#L75
> 
> Maybe a good idea to write them both now.

I think we actually just want to print out `serverId` and the `version`, and not bother with the `versionId`: the `versionId` can't be used for anything API-wise (the API is such that you want the `serverId` plus the version like `0.1.5`; the versionID UUID is just an internal detail a consumer would never touch).

That said, was worth updating the naming and adding `version` in there: https://github.com/modelcontextprotocol/registry/pull/465/commits/2264c5b1b7832ca4eb8727291174a89c40bf1c54

Thanks for the nudge to polish that up!

tadasant:
@claude review this whole PR please

tadasant:
Claude's one note of feedback addressed in: https://github.com/modelcontextprotocol/registry/pull/465/commits/06a50ca7abb440c38baee4d0e4f17ea863b05b81

rdimitrov:
@tadasant - Nice! As I shared I'm supportive of having 2 IDs and it got me thinking that once we introduce both server and version IDs at the moment there will be no way to make use of the version ones from the API perspective. I’ll give it some more thought this evening to see what options I can come up with and reply again here. In the meantime if you (and others, of course) have a chance it would be great if you could think about it too so we can make a decision soon-ish 👍

tadasant:
> @tadasant - Nice! As I shared I'm supportive of having 2 IDs and it got me thinking that once we introduce both server and version IDs at the moment there will be no way to make use of the version ones from the API perspective. I’ll give it some more thought this evening to see what options I can come up with and reply again here. In the meantime if you (and others, of course) have a chance it would be great if you could think about it too so we can make a decision soon-ish 👍

Just so I understand your concern, you're worried about the fact that `versionId` doesn't have an API-facing use case?

Personally, I think that is OK. If we really wanted, we could add something like:

`GET /v0/servers/{server_id}/versions/{version_id}`

Or

`GET /v0/versions/{version_id}`

But personally I think this is unnecessary surface area. I can't think of a use case where you'd want to fetch a version_id: because server.json is immutable, if you've fetched it once, you already have it (and should cache it). 

We could always add one of those endpoints later if it turns out there are meaningful use cases for it.

rdimitrov:
> Just so I understand your concern, you're worried about the fact that `versionId` doesn't have an API-facing use case?

Exactly, but I guess my reasoning is not just for fetching, but also admin/maintain operations on a particular version (edit/deletions). The use cases I can think of are:
* There's a report that there's a malicious version so it would be awkward (and not a best practice) to edit the DB directly instead of having an API to remove it
* A sub-registry may decide to cache only the latest version entry and keep a simpler cache of just the identifiers along with their IDs of older versions so upon interest they'll fetch the details of a particular version. Although I acknowledge this is against the intended way of using it there are no guardrails around someone doing this yet.

---

I had a chat with Claude based on your suggestions and the thoughts I had, and it came up with this which I think I favour (added comments after each part):

> ```bash
> # Server-centric (primary interface)
> GET    /v0/servers                           # Browse/search all servers
> GET    /v0/servers/{server_id}               # Latest version (most common)
> GET    /v0/servers/{server_id}/versions      # All versions
> GET    /v0/servers/{server_id}/versions/{version} # Specific version
> ```

* This is like your approach with the difference in `/v0/servers/{server_id}/versions/{version}` vs `/v0/servers/{server_id}?version=X.X.X` which I think I find cleaner (less url params)?

> ```
> # Version-direct (admin/power users)  
> GET    /v0/versions/{version_id}             # Direct version access
> PUT    /v0/versions/{version_id}             # Direct version edit
> DELETE /v0/versions/{version_id}             # Direct version removal
> ```

* This is based on your suggestions which I also like 👍 

I left the following two suggestions from Claude for the sake of presenting the full API layout but not to be addressed with this PR, if at all. 

> ```
> # Publishing
> POST /v0/servers                             # Publish any version (new server or new version)
> ```

* I think I agree as the first time I saw the API I was interested why there's a separate `publish` endpoint instead of POST-ing to `/servers`? This is not part of this problem per see so happy to discuss/address in another issue 👍 
 
> ```
> # Registry data
> GET /v0/servers/{server_id}/stats         # Download counts, etc.
> ```

* This could be a nice vendor-neutral set of signals that sub-registries can leverage to enhance their own curation and I know there are already discussions around this. Again suitable for another issue.

tadasant:
> > Just so I understand your concern, you're worried about the fact that `versionId` doesn't have an API-facing use case?
> 
> Exactly, but I guess my reasoning is not just for fetching, but also admin/maintain operations on a particular version (edit/deletions).

> The use cases I can think of are:
> > 
> > * There's a report that there's a malicious version so it would be awkward (and not a best practice) to edit the DB directly instead of having an API to remove it

I agree we need this functionality. This PR currently maintains the ability to do this via:

`- PUT /v0/servers/{server_id}?version=X.X.X - Edit specific server version`

> * A sub-registry may decide to cache only the latest version entry and keep a simpler cache of just the identifiers along with their IDs of older versions so upon interest they'll fetch the details of a particular version.

But this is still possible right? With:

`- **GET /v0/servers/{server_id}?version=X.X.X** - Get specific version of server`

Are you saying you agree with the functionalities implemented in this PR, you're just proposing we change `GET /v0/servers/{server_id}?version=X.X.X` and `PUT /v0/servers/{server_id}?version=X.X.X` to instead be `GET    /v0/servers/{server_id}/versions/{version}` and  `PUT    /v0/versions/{version_id}             # Direct version edit`?

My reasoning for preferring the query param on the GET is that consumers need not be exposed to the `version_id` implementation detail. It's unnecessary overhead to have them grok ("what is a version_id?") when to the MCP ecosystem it's simpler to speak in terms of the actual `version`.  And then I think `GET /v0/servers/{server_id}/versions/{version}` would be unintuitive because we're mixing a UUID with a non-UUID  so close to each other (`/v0/servers/{UUID}/versions/{string}`). The query param is best of both worlds.

And then my reasoning for preferring the query param on PUT is simply consistency between the GET and PUT API's. Though if you feel strongly we should have `version_id`-based admin PUT endpoints, I don't feel particularly strongly the other way.

rdimitrov:
The query parameter is actually a pretty valid point, I missed that! Thanks! 🙏  

To try and summarise my thinking about `semver-like` vs `UUIDs`. I tend to lean toward having a UUIDs-first design by default since they provide a clear, canonical reference without ambiguity. 

In practice if I were to implement a curated mirror of the official registry where not every version is exposed to my end users (because it didn't pass my filter for example) I would probably expect to work with UUIDs that map to a `server.json` blob. In that context the “version” is more or less a human-readable label while the UUID remains the true source of identity. 

I've been contributing to the project for only some time and I think what causes my confusion is the docs say that the intended usage of the registry is to act as a source for other sub-registries and not be consumed by client apps directly and I see UUID-first API better suited for ETL systems and semver-based API better suited for clients/end users.

In any case, I’m convinced by what you’re suggesting and I don't want to slow it so I’m good with moving forward as you’ve proposed. Let's address the persistent `server_id` reference now and continue the discussion/meet/have follow up PRs, if needed on the rest 👍 

Last but not least I think this is the perfect time to have these discussions and align ourselves better, so really appreciate your comments! 🙏 🍻

tadasant:
> The query parameter is actually a pretty valid point, I missed that! Thanks! 🙏
> 
> To try and summarise my thinking about `semver-like` vs `UUIDs`. I tend to lean toward having a UUIDs-first design by default since they provide a clear, canonical reference without ambiguity.
> 
> In practice if I were to implement a curated mirror of the official registry where not every version is exposed to my end users (because it didn't pass my filter for example) I would probably expect to work with UUIDs that map to a `server.json` blob. In that context the “version” is more or less a human-readable label while the UUID remains the true source of identity.
> 
> I've been contributing to the project for only some time and I think what causes my confusion is the docs say that the intended usage of the registry is to act as a source for other sub-registries and not be consumed by client apps directly and I see UUID-first API better suited for ETL systems and semver-based API better suited for clients/end users.
>
> In any case, I’m convinced by what you’re suggesting and I don't want to slow it so I’m good with moving forward as you’ve proposed. Let's address the persistent `server_id` reference now and continue the discussion/meet/have follow up PRs, if needed on the rest 👍
> 
> Last but not least I think this is the perfect time to have these discussions and align ourselves better, so really appreciate your comments! 🙏 🍻

That's fair - I think part of my inconsistency here is driven by the belief that while the core Registry functionality - the `GET /v0/servers` endpoint - is for sub-registry consumers, I think the `GET /v0/servers/{server_id}?version={version}` endpoint specifically is mostly for publishers (not sub-registries doing ETL) who are, for example, trying to inspect what they've recently published to the registry. Versus, it's not clear to me in what practical scenario a sub-registry would need to run a GET on a specific version.

FWIW: my perspective here would definitely be very different if server.json were mutable; so a lot of this thinking hinges on the fact that I expect these ETL processes to only touch most server.json entries once per-sub-registry in their lifetime. After you fetch and save it once, if you are following our guidance that you should mirror/cache it, you never need to fetch it again.

@rdimitrov did you intend to mark this as approved? Or still pending your review

rdimitrov (dismissed):
@tadasant - my bad, I missed to click approve after my last reply 👍 

Given it's breaking change though I wonder if we should wait for @domdomegg to have a look in case too he has any comments?

docs/reference/api/openapi.yaml
domdomegg, line 1:
```yaml
  version: "2025-07-09"
```

should we maybe bump this version string, as I think it probably is a breaking change

domdomegg: relatedly we might want to align this with switch from snake_case to camelCase, so there's fewer breaking change moments

domdomegg (approved):
Am happy with this. Have skimmed the code and implementation seems good, although I am somewhat relying on @.rdimitrov's approval here.

As above, I think I'd prefer this went out potentially with a version bump to the general server registry API schema and alongside #428

tadasant:
Agree let's land this and https://github.com/modelcontextprotocol/registry/issues/428 at the same time. Whoever takes that should branch off this branch. Targeting landing both by ~tomorrow EOD

pree-dew:
Migration PR https://github.com/modelcontextprotocol/registry/pull/519

PR #466, opened 2025-09-12
Fix #463: Omit empty repository from JSON output for remote-only servers

### LINKED ISSUE #463: Blank repository get inserted in published server
What's the value of including a blank repository in the published json when my server is remote-only?

```json
{
  ...
  "repository": {
    "url": "",
    "source": ""
  },
  ...
}
```

jalateras:
## Summary

This PR fixes an issue where remote-only servers (servers with no repository) were still including an empty repository object with blank URL and source fields in their JSON output. This adds unnecessary noise to the API responses.

## What Changed

1. **Added omitempty tags to Repository struct fields**: The `URL` and `Source` fields in the Repository struct now have `omitempty` tags to prevent empty string values from being serialized.

2. **Implemented custom MarshalJSON for ServerJSON**: A custom marshaling method was added to properly handle the case where a Repository struct is initialized but contains only empty values. This ensures the entire repository field is omitted from the JSON output when all its fields are empty.

3. **Added comprehensive tests**: Test coverage was added for both the Repository struct marshaling and the ServerJSON marshaling to verify the behavior works correctly.

## Technical Details

The fix uses a two-pronged approach:
- First, the Repository struct fields get `omitempty` tags so individual empty fields are omitted
- Second, a custom MarshalJSON method for ServerJSON checks if all Repository fields are empty and omits the entire repository object in that case

This ensures backward compatibility while fixing the issue for remote-only servers.

## Testing

- Added unit tests for Repository JSON marshaling/unmarshaling
- Added integration tests for ServerJSON to verify empty repositories are omitted
- All existing tests pass

Fixes #463

### DIFF
--- a/pkg/api/v0/marshal.go
+++ b/pkg/api/v0/marshal.go
@@ -1,0 +1,49 @@
+package v0
+
+import (
+	"encoding/json"
+	
+	"github.com/modelcontextprotocol/registry/pkg/model"
+)
+
+// MarshalJSON implements custom JSON marshaling for ServerJSON to properly handle empty repositories
+func (s ServerJSON) MarshalJSON() ([]byte, error) {
+	// Check if repository is empty (all fields are zero values)
+	isEmptyRepo := s.Repository.URL == "" && s.Repository.Source == "" && 
+		s.Repository.ID == "" && s.Repository.Subfolder == ""
+	
+	// Create an alias type to avoid infinite recursion
+	type Alias ServerJSON
+	
+	if isEmptyRepo {
+		// Create a version without the repository field
+		type ServerJSONNoRepo struct {
+			Schema        string              `json:"$schema,omitempty"`
+			Name          string              `json:"name"`
+			Description   string              `json:"description"`
+			Status        model.Status        `json:"status,omitempty"`
+			Version       string              `json:"version"`
+			WebsiteURL    string              `json:"website_url,omitempty"`
+			Packages      []model.Package     `json:"packages,omitempty"`
+			Remotes       []model.Transport   `json:"remotes,omitempty"`
+			Meta          *ServerMeta         `json:"_meta,omitempty"`
+		}
+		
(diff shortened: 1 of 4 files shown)


pkg/api/v0/marshal.go
```diff
@@ -0,0 +1,49 @@
+package v0
+
+import (
+	"encoding/json"
+	
+	"github.com/modelcontextprotocol/registry/pkg/model"
+)
+
+// MarshalJSON implements custom JSON marshaling for ServerJSON to properly handle empty repositories
+func (s ServerJSON) MarshalJSON() ([]byte, error) {
```
rdimitrov, line 10: Instead of custom marshalling could we handle this in the database retrieval methods? After unmarshalling `ServerJSON` in `./internal/database/postgres.go` and `./internal/database/memory.go` have something like:

```go
// Clean empty repository to make omitempty work
if serverJSON.Repository.URL == "" && serverJSON.Repository.Source == "" &&
   serverJSON.Repository.ID == "" && serverJSON.Repository.Subfolder == "" {
    serverJSON.Repository = model.Repository{}
}
```

Your current approach works fine too though! My only reasoning is this would be simpler to maintain and leverage the existing: 

```go
Repository    model.Repository    `json:"repository,omitempty"`
``` 

rather than a custom marshalling logic. What do you think?

rdimitrov (commented):
@jalateras - Thanks for finding and addressing this issue! 💯 The solution works well, added a small suggestion to make it even simpler 👍

BobDickinson:
If I'm reading this correctly, it looks like it will address https://github.com/modelcontextprotocol/registry/issues/613 (`mcp-publisher` generating these empty fields).  

FWIW, the empty fields are a schema violation and there are ~90 such servers currently published.

rdimitrov:
I will close this one as it got addressed by https://github.com/modelcontextprotocol/registry/pull/717

PR #476, opened 2025-09-14
fix: validate server names to allow only single slash (#471)

ossamalafhel:
## Summary
  Added validation to reject server names with multiple slashes, ensuring consistency between JSON schema pattern and API validation.

  - Count slashes in parseServerName() and reject if > 1
  - Add ErrMultipleSlashesInServerName error constant
  - Add unit tests in validators_test.go
  - Add integration test in publish_test.go

  ## Motivation and Context
  Fixes #471. The JSON schema regex pattern `^[^/]+/[^/]+$` only allows a single slash in server names, but the API validation was not enforcing this constraint. This inconsistency
   allowed invalid server names like `com.example/server/path` to be published.

  ## How Has This Been Tested?
  - Unit tests added to verify slash validation logic
  - Integration tests added to verify the publish endpoint rejects multi-slash names
  - Tested various edge cases: trailing slashes, consecutive slashes, URL-like paths
  - All existing tests pass

  ## Breaking Changes
  None. This change only adds validation to reject previously invalid server names that shouldn't have been accepted.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed
 
## Additional context
  - The fix ensures server names follow the intended format: `namespace/name` (e.g., `com.example/my-server`)
  - This validation aligns the API behavior with the existing JSON schema regex pattern
  - Test coverage includes edge cases like `com.example//server`, `com.example/server/`, and `a/b/c/d/e/f`
  - The error message clearly indicates the issue: "server name cannot contain multiple slashes"

PR #475, opened 2025-09-14
fix: use authenticated user endpoint for fetching organizations (#396)

ossamalafhel:
Changed GitHub API endpoint from /users/{username}/orgs to /user/orgs to retrieve all organizations for the authenticated user.

  Problem

  The /users/{username}/orgs endpoint only returns organizations where the user's membership is public, causing 403 errors for users with private organization membership when
  trying to publish.

  Solution

  Using /user/orgs returns all organizations (public and private) for the authenticated user, allowing them to publish to any organization they belong to.

  Impact

  - Users can publish to all their organizations without changing visibility
  - Maintains existing security model (same auth token required)
  - Resolves 403 permission errors reported in issue #396
  - Added test to ensure correct endpoint usage

  This PR fixes the organization publishing permissions issue by using the correct GitHub API endpoint that returns all organizations, not just public ones.

  Motivation and Context

  Multiple users reported being unable to publish servers to their organizations despite being members. This was causing frustration and blocking legitimate use cases where
  organization membership is kept private for security or privacy reasons.

  How Has This Been Tested?

  - Added unit test TestGitHubHandler_UsesUserOrgsEndpoint that verifies:
    - The /user/orgs endpoint is called (not /users/{username}/orgs)
    - Private organizations are included in the response
    - Permissions are correctly generated for all organizations
  - All existing auth tests continue to pass
  - Manually verified the endpoint change follows GitHub API documentation

  Breaking Changes

  None. This is a backwards-compatible fix that expands access to include private organizations.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

  Additional context

  The GitHub API documentation confirms that /user/orgs is the correct endpoint for retrieving all organizations of the authenticated user. This change aligns with GitHub's
  recommended practices for OAuth applications.

PR #479, opened 2025-09-15
feat: add regex validation to server name in API types

ossamalafhel:
Body:
  ## Summary
  Following @domdomegg's suggestion in #476, this PR adds regex validation to the server name field in the API types.

  ## Changes
  - Add pattern validation `^[^/]+/[^/]+$` to the `Name` field in `pkg/api/v0/types.go`
  - This ensures server names have exactly one slash, matching the existing backend validation
  - The pattern will be visible in the API documentation (OpenAPI spec)

  ## Context
  This complements the validation fix from #476 by making the validation rules visible at the API specification level, improving the developer experience for API consumers.

  Related to #475
  Builds on #476


internal/validators/validators.go
```diff
@@ -407,6 +407,11 @@ func parseServerName(serverJSON apiv0.ServerJSON) (string, error) {
 	if slashCount > 1 {
 		return "", ErrMultipleSlashesInServerName
 	}
+	
+	// Check for consecutive slashes
+	if strings.Contains(name, "//") {
```
domdomegg: why is this needed - I'd expect it's already caught by the slashCount check just above?

ossamalafhel: Good catch! You're absolutely right. The strings.Contains(name, "//") check is redundant because:
  - "com.example//server" has strings.Count(name, "/") = 2
  - This is already caught by slashCount > 1

  I've removed the redundant check. Thanks for the review!

internal/api/handlers/v0/publish_test.go
```diff
@@ -277,8 +277,8 @@ func TestPublishEndpoint(t *testing.T) {
 				},
 			},
 			setupRegistryService: func(_ service.RegistryService) {},
-			expectedStatus:       http.StatusBadRequest,
-			expectedError:        "server name cannot contain multiple slashes",
+			expectedStatus:       http.StatusUnprocessableEntity,
+			expectedError:        "expected string to match pattern",
```
joelverhagen: I think this error message is a little too generic. The previous one mentions the field in question (server name). It is unclear what field is the problem now.

internal/api/handlers/v0/edit_test.go
```diff
@@ -236,8 +236,8 @@ func TestEditServerEndpoint(t *testing.T) {
 				Version: "1.0.0",
 			},
 			serverID:       testServerID,
-			expectedStatus: http.StatusBadRequest,
-			expectedError:  "Bad Request",
+			expectedStatus: http.StatusUnprocessableEntity,
```
joelverhagen: nitpick: IMO HTTP 400 is better simply because it's more ubiquitous. I have never seen HTTP 422 until this point. I don't feel strongly, but perhaps sticking with 400 would cause a little less surprise for users.

pkg/api/v0/types.go
```diff
@@ -29,7 +29,7 @@ type ServerMeta struct {
 // ServerJSON represents complete server information as defined in the MCP spec, with extension support
 type ServerJSON struct {
 	Schema        string              `json:"$schema,omitempty"`
-	Name          string              `json:"name" minLength:"1" maxLength:"200"`
+	Name          string              `json:"name" minLength:"1" maxLength:"200" pattern:"^[^/]+/[^/]+$" doc:"Server name in format 'reverse-dns-namespace/name' (e.g., 'com.example/server')"`
```
joelverhagen: The regex is a little more strict than this I think:
https://github.com/modelcontextprotocol/registry/blob/63cf08efad6ec056cf7c174221db25572a429385/docs/reference/server-json/server.schema.json#L50C23-L50C55

```suggestion
	Name          string              `json:"name" minLength:"1" maxLength:"200" pattern:"^[a-zA-Z0-9.-]+/[a-zA-Z0-9._-]+$" doc:"Server name in format 'reverse-dns-namespace/name' (e.g., 'com.example/server')"`
```

joelverhagen: Consider lifting the pattern out of the annotations into the validator code itself to allow us to make the error message more specific.

internal/api/handlers/v0/edit_test.go
ossamalafhel: @joelverhagen You're right on both points:

  1. Error message: The error is indeed less helpful now. Huma returns a generic message, though the full error response should include the field location:
  {
    "title": "Unprocessable Entity",
    "status": 422,
    "detail": "validation failed",
    "errors": [{
      "message": "expected string to match pattern ^[^/]+/[^/]+$",
      "location": "body.name",
      "value": "com.example/server/invalid"
    }]
  }
  1. So while the message is generic, location: "body.name" indicates which field failed.
  2. HTTP 400 vs 422: I agree 400 is more ubiquitous and familiar. We actually tried hard to make it work with 400 - we spent considerable effort trying to intercept Huma's
  validation to return 400 instead, but it's not possible. Huma validates the schema (including regex patterns) before our handler runs, and it always returns 422 for validation
  failures.

  We have two options:
  - Keep the regex pattern (better API documentation) but accept 422
  - Remove the regex pattern and validate only in handler code to return 400 with clearer messages

  Given your feedback about 400 being more familiar and the error messages being clearer, maybe option 2 is better for developer experience?

  @domdomegg What do you think - would you be okay with removing the regex pattern to maintain the familiar 400 status and clearer error messages?

pkg/api/v0/types.go
ossamalafhel: Great catch! You're absolutely right - I used a more permissive regex than what's in the schema. The schema's regex is more appropriate as it enforces valid characters for both the namespace and name parts.

I'll update it to use the correct pattern. Thanks for pointing to the schema file!

joelverhagen: @domdomegg - are server names case sensitive? NuGet and PyPI treat package names as case insensitive at the API level but npm and Docker do not (last I checked). Personally, I recommend enforcing lowercase everywhere and being case sensitive. It is has caused headaches in NuGet and the roughly "domain name + path" pattern of server names is not unlike URLs which have a lowercase domain name (ubiquitous) and kebab case path portion (not ubiquitous but common in say GH repo names).

ossamalafhel: That's a great suggestion! By moving the pattern validation to our validator code, we can:
  - Return HTTP 400 (more familiar) instead of 422
  - Provide clearer error messages like "server name must match pattern [a-zA-Z0-9.-]+/[a-zA-Z0-9._-]+: only alphanumeric characters, dots, hyphens and underscores are allowed"
  - Still document the expected format in the doc tag

  This would give us the best of both worlds - clear documentation and better error messages. I'll make this change.

internal/validators/validators.go
```diff
@@ -413,6 +413,19 @@ func parseServerName(serverJSON apiv0.ServerJSON) (string, error) {
 		return "", fmt.Errorf("server name must be in format 'dns-namespace/name' with non-empty namespace and name parts")
 	}
 
+	// Validate namespace and name format according to schema
+	// Pattern: ^[a-zA-Z0-9.-]+/[a-zA-Z0-9._-]+$
+	namespacePattern := regexp.MustCompile(`^[a-zA-Z0-9.-]+$`)
+	namePattern := regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
```
joelverhagen: @domdomegg - are server names case sensitive? NuGet and PyPI treat package names as case insensitive at the API level but npm and Docker do not (last I checked). Personally, I recommend enforcing lowercase everywhere and being case sensitive. It is has caused headaches in NuGet and the roughly "domain name + path" pattern of server names is not unlike URLs which have a lowercase domain name (ubiquitous) and kebab case path portion (not ubiquitous but common in say GH repo names).

ossamalafhel: @joelverhagen Good point! Currently the regex allows uppercase letters in both namespace and name parts.

  I agree that enforcing lowercase would be beneficial for consistency, especially since:
  - The namespace part follows reverse DNS convention where domains are lowercase
  - It avoids confusion between `Com.Example/server` vs `com.example/server`
  - It aligns with web conventions

  Should we update the regex to enforce lowercase in this PR or create a separate issue for it? If we do it here, we'd need to:
  1. Update the regex patterns to `^[a-z0-9.-]+/[a-z0-9._-]+$`
  2. Update the error messages and documentation
  3. Add tests for uppercase rejection

  What do you think @domdomegg?

Avish34: nit: Can we move the regex to **utils.go**? We have kept regex for the URLs there. It kind of make sense to have all regex used in validation at one place.

internal/validators/validators.go
```diff
@@ -413,6 +412,19 @@ func parseServerName(serverJSON apiv0.ServerJSON) (string, error) {
...
+	
+	if !namespacePattern.MatchString(parts[0]) {
```
Avish34: We can have constant error strings, you can check constants.go under validators folder.

internal/validators/validators.go
ossamalafhel: @Avish34 Thanks for the review! Good suggestions:

  1. Moved regex patterns to `utils.go` alongside other validation regexes
  2. Added constant error messages in `constants.go`:
     - `ErrInvalidNamespaceCharacters`
     - `ErrInvalidNameCharacters`

  This keeps all validation patterns centralized and error messages consistent. Changes implemented in the latest commit!

internal/validators/validators.go
ossamalafhel: @Avish34 Great catch! I've refactored the code following your suggestions:

  - Regex patterns now live in `utils.go` as `serverNamespacePattern` and `serverNamePattern`
  - Error messages are now constants in `constants.go` for better maintainability
  - This aligns with the existing pattern for other validators

internal/api/handlers/v0/edit.go
```diff
@@ -68,6 +69,10 @@ func RegisterEditEndpoints(api huma.API, registry service.RegistryService, cfg *
 
 		// Prevent renaming servers
 		if currentServer.Name != input.Body.Name {
+			// Validate the new name format before rejecting the rename
```
domdomegg, line 72: This seems unnecessary, if we're going to reject the request anyway?

internal/validators/validators.go
```diff
@@ -402,11 +401,27 @@ func parseServerName(serverJSON apiv0.ServerJSON) (string, error) {
 		return "", fmt.Errorf("server name must be in format 'dns-namespace/name' (e.g., 'com.example.api/server')")
 	}
 
+	// Check for multiple slashes - reject if found
+	slashCount := strings.Count(name, "/")
```
domdomegg, line 405: I think this would already be caught by the parts split and patterns below? Ideally I want to keep this simple as it's hard to follow when there are quite so many rules - eliminating overlapping/duplicate ones is valuable. Possibly can all be merged into one fairly simple regex.

rdimitrov:
Hey, @ossamalafhel, sorry for the long delay here.

We are moving forward with another effort that revisits the whole validation process and supersedes this so we’re going to close the PR.

Thanks a lot for the contribution and for your patience and hope you’ll contribute again in the future 🙌  Cheers!

PR #484, opened 2025-09-15
Extra validation for registries

Avish34:
Package identifier and version are required for some of the registries else we won't be able to pull the meta data to validate the packages.

## How Has This Been Tested?

Unit test.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist


Avish34:
@domdomegg please review!

internal/validators/registries/pypi.go
```diff
@@ -11,6 +12,10 @@ import (
 	"github.com/modelcontextprotocol/registry/pkg/model"
 )
 
+var (
```
domdomegg, line 15: I think PyPi should also specify version

Avish34: @domdomegg We are not using version in pypi url to fetch the data about package. So, I don't think we should add validation for it.

domdomegg: hmm that seems like potentially a mistake in the validation logic - we should be checking it for the version submitted. can we fix this too?

Avish34: sure @domdomegg, I will push the fix in the same PR.

Avish34: @domdomegg added the required changes, please review.

Avish34: @domdomegg  shall we merge this if the changes looks good?

PR #490, opened 2025-09-16
Fix #481: Add server name scoping support in DNS TXT records

### LINKED ISSUE #481: Allow server name scoping in TXT record for auth

Currently, a TXT record on a DNS name provides wildcard (broad) access to all subdomains over the TXT record (not just immediate subdomains -- all levels, which is a little surprising compared to wildcard SSL certs which only provide one level of subdomain). It also provides all values for the path portion of the server name with a similar wildcard.

Here are the claims in a DNS TXT auth token.

Image

This is for the following TXT record on my personal `mcp.joelverhagen.com` domain:
```
v=MCPv1; k=ed25519; p=OHjrTGdvR2dFk1g5uTVNJ4/RxpDLYjVJTtTQlcwW0Jg=
```

This currently asserts that the **domain record controller** deserves the same permissions as the **MCP server author**. In other words, it is not possible to have a DNS record admin provide subsets of names to teams in the path portion of the server name. This is possible with subdomains, like (`com.microsoft.foo/*` and `com.microsoft.bar/*`) but not with server names.

At Microsoft we are considering `com.microsoft` for all DNS parts of the server name and qualifiers only within the path. It would be great to allow name prefixes to be defines in the TXT record.

Add an optional `n=<pattern>` key value pair to the TXT record which internally defaults to `*` in the authentication service when it is evaluating TXT records.

For example, `com.joelverhagen.mcp/foo-*` could be expressed as:

```
v=MCPv1; k=ed25519; p=OHjrTGdvR2dFk1g5uTVNJ4/RxpDLYjVJTtTQlcwW0Jg=; n=com.joelverhagen.mcp/foo-*
```

This is only 75 characters, well below the 255 character maximum.

This extension would allow the DNS controller to issue key pairs to teams and isolate them to specific name prefixes.

We could opt to make `n=` only about the path portion if we want to be more succinct but this would prevent the DNS controller from restricting access to subdomains.

It is very cool you have multiple TXT records and the first with a matching key words. This allows zero downtime key rotation as well. Nice job!

The code that would change is here I think:
https://github.com/modelcontextprotocol/registry/blob/63cf08efad6ec056cf7c174221db25572a429385/internal/api/handlers/v0/auth/dns.go#L124-L145

- Keep namespacing and permission isolation at the subdomain level not with prefixes in the path name portion.
- Simply issue N key pairs to the various teams using the same DNS name, conceding that they have shared permissions. This at least allows granular revocation.

**Additional context**

N/A

jalateras:
## Summary

This PR implements optional server name scoping in DNS TXT records for authentication, allowing fine-grained permission control through an optional `n=<pattern>` parameter.

## Problem

Currently, DNS TXT records provide wildcard access to ALL subdomains and path portions, making it impossible for DNS controllers to delegate limited namespaces to different teams. This was particularly problematic for enterprise scenarios like Microsoft's, where they need to scope permissions to specific prefixes (e.g., `com.microsoft/team-foo-*`).

## Solution

Added support for an optional `n=<pattern>` parameter in DNS TXT records that allows specifying exact permission patterns. When not specified, the system maintains backward compatibility by defaulting to wildcard permissions.

### Example usage:
```
v=MCPv1; k=ed25519; p=OHjrTGdvR2dFk1g5uTVNJ4/RxpDLYjVJTtTQlcwW0Jg=; n=com.example/team-foo-*
```

## Changes

- **Added `DNSAuthRecord` struct**: New data structure to hold both public key and optional name pattern
- **Updated TXT parsing**: Modified regex to capture optional `n=<pattern>` parameter
- **Enhanced permission building**: `buildPermissions()` now uses specific patterns when provided
- **Comprehensive test coverage**: Added extensive tests for all scoping scenarios

## Testing

- ✅ All existing tests pass (backward compatibility maintained)
- ✅ Added new test suite `TestDNSAuthHandler_NamePatternScoping` with 8 test cases
- ✅ Verified wildcard behavior when `n=` is not specified
- ✅ Tested multiple keys with different patterns
- ✅ Tested complex patterns with multiple path segments

## Backward Compatibility

This change is fully backward compatible:
- Existing TXT records without `n=` parameter continue to work exactly as before
- Default behavior (wildcard permissions) is preserved when `n=` is not specified
- No changes required for existing deployments

Fixes #481

### DIFF
--- a/internal/api/handlers/v0/auth/dns.go
+++ b/internal/api/handlers/v0/auth/dns.go
@@ -40,6 +40,12 @@ func (r *DefaultDNSResolver) LookupTXT(ctx context.Context, name string) ([]stri
 	return (&net.Resolver{}).LookupTXT(ctx, name)
 }
 
+// DNSAuthRecord represents a DNS TXT authentication record with optional name pattern
+type DNSAuthRecord struct {
+	PublicKey   ed25519.PublicKey
+	NamePattern string // Defaults to "*" for wildcard access
+}
+
 // DNSAuthHandler handles DNS-based authentication
 type DNSAuthHandler struct {
 	config     *config.Config
@@ -120,29 +126,34 @@ func (h *DNSAuthHandler) ExchangeToken(ctx context.Context, domain, timestamp, s
 		return nil, fmt.Errorf("failed to lookup DNS TXT records: %w", err)
 	}
 
-	// Parse public keys from TXT records
-	publicKeys := h.parsePublicKeysFromTXT(txtRecords)
+	// Parse auth records from TXT records
+	authRecords := h.parseAuthRecordsFromTXT(txtRecords)
 
-	if len(publicKeys) == 0 {
+	if len(authRecords) == 0 {
 		return nil, fmt.Errorf("no valid MCP public keys found in DNS TXT records")
 	}
 
-	// Verify signature with any of the public keys
+	// Verify signature and collect all valid records
 	messageBytes := []byte(timestamp)
-	signatureValid := false
-	for _, publicKey := range publicKeys {
-		if ed25519.Verify(publicKey, messageBytes, signature) {
-			signatureValid = true
-			break
+	var validRecords []DNSAuthRecord
+	for _, record := range authRecords {
+		if ed25519.Verify(record.PublicKey, messageBytes, signature) {
+			validRecords = append(validRecords, record)
 		}
 	}
 
-	if !signatureValid {
+	if len(validRecords) == 0 {
 		return nil, fmt.Errorf("signature verification failed")
 	}
 
-	// Build permissions for domain and subdomains
-	permissions := h.buildPermissions(domain)
+	// Build permissions from all valid records
+	var permissions []auth.Permission
+	for _, record := range validRecords {
+		permissions = append(permissions, h.buildPermissions(domain, record.NamePattern)...)
+	}
+	if len(permissions) == 0 {
+		return nil, fmt.Errorf("no valid permissions found for the given name pattern")
+	}
 
 	// Create JWT claims
 	jwtClaims := auth.JWTClaims{
@@ -160,14 +171,16 @@ func (h *DNSAuthHandler) ExchangeToken(ctx context.Context, domain, timestamp, s
 	return tokenResponse, nil
 }
 
-// parsePublicKeysFromTXT parses Ed25519 public keys from DNS TXT records
-func (h *DNSAuthHandler) parsePublicKeysFromTXT(txtRecords []string) []ed25519.PublicKey {
-	var publicKeys []ed25519.PublicKey
-	mcpPattern := regexp.MustCompile(`v=MCPv1;\s*k=ed25519;\s*p=([A-Za-z0-9+/=]+)`)
+// parseAuthRecordsFromTXT parses DNS authentication records from TXT records
+// Supports optional n=<pattern> parameter for name scoping
+func (h *DNSAuthHandler) parseAuthRecordsFromTXT(txtRecords []string) []DNSAuthRecord {
+	var authRecords []DNSAuthRecord
+	// Updated pattern to capture optional n=<pattern> parameter
+	mcpPattern := regexp.MustCompile(`v=MCPv1;\s*k=ed25519;\s*p=([A-Za-z0-9+/=]+)(?:;\s*n=([^;]+))?`)
 
 	for _, record := range txtRecords {
 		matches := mcpPattern.FindStringSubmatch(record)
-		if len(matches) == 2 {
+		if len(matches) >= 2 {
 			// Decode base64 public key
 			publicKeyBytes, err := base64.StdEncoding.DecodeString(matches[1])
 			if err != nil {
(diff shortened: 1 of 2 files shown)


internal/api/handlers/v0/auth/dns.go
```diff
@@ -120,29 +126,29 @@ func (h *DNSAuthHandler) ExchangeToken(ctx context.Context, domain, timestamp, s
...
+	// Verify signature with any of the auth records
 	messageBytes := []byte(timestamp)
-	signatureValid := false
-	for _, publicKey := range publicKeys {
-		if ed25519.Verify(publicKey, messageBytes, signature) {
-			signatureValid = true
+	var validRecord *DNSAuthRecord
+	for i, record := range authRecords {
+		if ed25519.Verify(record.PublicKey, messageBytes, signature) {
+			validRecord = &authRecords[i]
```
joelverhagen: I believe the order of the TXT records is undefined and this is given "[random?] first wins". What about collecting all of the valid records an expressing the union of the permissions?

internal/api/handlers/v0/auth/dns.go
```diff
@@ -178,29 +186,52 @@ func (h *DNSAuthHandler) parsePublicKeysFromTXT(txtRecords []string) []ed25519.P
...
-			Action:          auth.PermissionActionPublish,
-			ResourcePattern: fmt.Sprintf("%s/*", reverseDomain),
-		},
-		// DNS implies a hierarchy where subdomains are treated as part of the parent domain,
-		// therefore we grant permissions for all subdomains (e.g., com.example.*)
-		// This is in line with other DNS-based authentication methods e.g. ACME DNS-01 challenges
 		{
 			Action:          auth.PermissionActionPublish,
-			ResourcePattern: fmt.Sprintf("%s.*", reverseDomain),
+			ResourcePattern: namePattern,
```
joelverhagen, line 266: This allows arbitrary access, I think, by putting say `n= *` in the record. This would give the `*` name space, which is a severe security issue, if I'm not mistaken.

I believe we must include the `reverseDomain` in the resource pattern otherwise we will allow access to multiple domains.

joelverhagen: Alternatively, we can check that `namePattern` starts with `reverseDomain`.

internal/api/handlers/v0/auth/dns_test.go
```diff
@@ -194,3 +194,172 @@ func TestDNSAuthHandler_ExchangeToken(t *testing.T) {
...
+			txtRecords: []string{
+				fmt.Sprintf("v=MCPv1; k=ed25519; p=%s", publicKeyB64_1),
+				fmt.Sprintf("v=MCPv1; k=ed25519; p=%s; n=com.example/limited-*", publicKeyB64_2),
+			},
+			usePrivateKey:    privateKey1,
+			expectedPatterns: []string{"com.example/*", "com.example.*"},
+			expectedNumPerms: 2,
+			expectError:      false,
+		},
+		{
```
joelverhagen, line 302: We should have some tests where the domain is not present in the name pattern.

internal/api/handlers/v0/auth/dns.go
jalateras: @joelverhagen Great catch! I've updated the implementation to collect ALL valid records and express the union of permissions, exactly as you suggested.

internal/api/handlers/v0/auth/dns.go
jalateras: @joelverhagen You're absolutely right, allowing arbitrary patterns would be a severe vulnerability.

  I've implemented your suggested solution

internal/api/handlers/v0/auth/dns_test.go
jalateras: @joelverhagen test cases added.

jalateras:
squashed and pushed

DNS_SCOPING_TEST_GUIDE.md
```diff
@@ -0,0 +1,294 @@
+# DNS TXT Name Scoping - Testing Guide
```
joelverhagen: I am not sure if this should be in the root of the repo. If we scale this approach out, the root will be filled with a lot of rather specific test guides. I do like the idea of documenting the tests for human and machine alike.
Maybe under `tests` or `docs/tests` or `tests/docs`?

internal/api/handlers/v0/auth/dns_test.go
```diff
@@ -194,3 +194,205 @@ func TestDNSAuthHandler_ExchangeToken(t *testing.T) {
...
+			name:   "complex pattern with multiple path segments",
+			domain: "example.com",
+			txtRecords: []string{
+				fmt.Sprintf("v=MCPv1; k=ed25519; p=%s; n=com.example/services/auth/*", publicKeyB64_1),
+			},
+			usePrivateKey:    privateKey1,
+			expectedPatterns: []string{"com.example/services/auth/*"},
```
joelverhagen: With https://github.com/modelcontextprotocol/registry/pull/476 this name is no longer valid (exactly 1 slash per server name). It seems okay to test this since server name validation is done elsewhere) but IMHO we should reject bad server name patterns as early as possible. No sure what other people thing though.

internal/api/handlers/v0/auth/dns_test.go
```diff
@@ -194,3 +194,205 @@ func TestDNSAuthHandler_ExchangeToken(t *testing.T) {
...
+			name:   "name pattern does not start with reverse domain",
+			domain: "example.com",
+			txtRecords: []string{
+				fmt.Sprintf("v=MCPv1; k=ed25519; p=%s; n=org.example/foo-*", publicKeyB64_1),
+			},
+			usePrivateKey:    privateKey1,
+			expectError:      true,
+			errorContains:    "no valid permissions found",
+		},
+		{
```
joelverhagen, line 365: Could we have tests for a SLD prefixing another? I am looking at the code and, with the meager caffeine fix I have in me right now, I can't see if there's a problem. But I think so.

Suppose the domain is `micro.com` and the `n=com.microsoft/*`, I think this would be allowed by the code, allowing improper permissions.
`string.HasPrefix("com.microsoft/*", "com.micro") = true`

The baseline implementation works because it enforces a `.` or `/` delimiter.

DNS_SCOPING_TEST_GUIDE.md
jalateras: @joelverhagen I will remove

internal/api/handlers/v0/auth/dns_test.go
jalateras: I can modify the test case. It will now expect an error when an invalid name pattern is used. This way, if the validation logic ever changes or is accidentally removed, the test will fail and immediately alert us to the regression.

Thoughts?

internal/api/handlers/v0/auth/dns_test.go
jalateras: 1. Add delimiter checking after domain prefix (must be . or /)
2. Enforce single-slash rule for server names per #476

DNS_SCOPING_TEST_GUIDE.md
jalateras: removed

PR #500, opened 2025-09-17
fix: revert prometheus client_golang update causing deployment failures

tadasant:
## Summary
- Reverts commit 3443121 that bumped `github.com/prometheus/client_golang` from 1.23.0 to 1.23.2
- Fixes production deployment timeouts where pods fail to become ready

## Root Cause
The prometheus client update v1.23.2 appears to cause runtime issues during application startup:
- Both `mcp-registry` and `grafana` deployments timeout waiting to be Ready  
- Pods fail to reach minimum required count for liveness
- 600+ second timeouts suggest initialization hangs during prometheus metrics setup

## Test Plan
- [x] Verify deployment succeeds in staging after revert
- [ ] Monitor production deployment for successful pod startup
- [ ] Investigate prometheus v1.23.2 compatibility issues separately


tadasant:
Closing this PR after realizing staging deployments were succeeding on main branch.

If the prometheus client update was the root cause, both staging and production should have failed with identical symptoms since they use the same code, Docker image, and dependencies.

The fact that staging succeeded but production failed suggests infrastructure differences between environments rather than a code/dependency issue.

Will investigate production cluster resources, configuration differences, and deployment timing issues instead.

PR #520, opened 2025-09-21
Migrate old entries to the new server.json schema

rdimitrov:
Originally opened by @pree-dew 

## Motivation and Context
- Make all entries in current database matching the new schema

## How Has This Been Tested?
- locally

## Breaking Changes
- Yes

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [X] I have added or updated documentation as needed

Before migration:

```
{
      "name": "io.github.kenjihikmatullah/productboard-mcp",
      "description": "Integrate the Productboard API into agentic workflows via MCP",
      "status": "active",
      "repository": {
        "url": "https://github.com/kenjihikmatullah/productboard-mcp",
        "source": "github",
        "id": "939104710"
      },
      "version": "0.0.1-seed",
      "packages": [
        {
          "registry_type": "npm",
          "identifier": "productboard-mcp",
          "version": "1.0.1",
          "transport": {
            "type": "stdio"
          }
        }
      ],
      "_meta": {
        "io.modelcontextprotocol.registry/official": {
          "id": "00613acb-73e2-4f93-8b96-296df17316c8",
          "published_at": "0001-01-01T00:00:00Z",
          "updated_at": "0001-01-01T00:00:00Z",
          "is_latest": true
        }
      }
    }
    
```

After migration:

```
 {
      "name": "io.github.kenjihikmatullah/productboard-mcp",
      "description": "Integrate the Productboard API into agentic workflows via MCP",
      "status": "active",
      "repository": {
        "url": "https://github.com/kenjihikmatullah/productboard-mcp",
        "source": "github",
        "id": "939104710"
      },
      "version": "0.0.1-seed",
      "packages": [
        {
          "registryType": "npm",
          "identifier": "productboard-mcp",
          "version": "1.0.1",
          "transport": {
            "type": "stdio"
          },
        }
      ],
      "_meta": {
        "io.modelcontextprotocol.registry/official": {
          "serverId": "83cfe55c-55c5-4e7e-b9c0-6d7d369f52ed",
          "versionId": "00613acb-73e2-4f93-8b96-296df17316c8",
[...]

PR #488, opened 2025-09-15
change fields related to server json struct to follow camelCase instead of snake_case

pree-dew:
## Motivation and Context
Aligns with issue https://github.com/modelcontextprotocol/registry/issues/428
To bring consistency across naming convention

## How Has This Been Tested?
- Locally test cases are tested

## Breaking Changes
- Yes, they have to update the server fields
- 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #527, opened 2025-09-21
Ensure server.json is immutable 

rdimitrov:
## Motivation and Context

The following PR refactors the code in order to ensure server.json is immutable.  

**Core Changes:**
- Database schema migration: Two-phase migration (007 → 008) to separate immutable server.json from mutable registry metadata
- API structure change: Server responses now use ServerResponse{server, _meta} wrapper pattern instead of flat structure
- Fixed isLatest flag bug: Added UpdateIsLatest() method to properly update previous versions when new versions are published
- Bumped the schema versions
- Added an `edit` command to the mcp-publisher CLI for setting the status

**New Files:**
- internal/database/migrations/007_move_status_to_registry_extensions.sql - Interim migration
- internal/database/migrations/008_separate_server_json_from_registry_metadata.sql - Final table restructure
- internal/api/handlers/v0/status.go - Dedicated PATCH endpoint for status updates

**Modified Components:**
- Database layer: Added UpdateIsLatest() method to PostgreSQL and memory implementations
- API types: Updated pkg/api/v0/types.go with ServerResponse, ResponseMeta, RegistryExtensions
- Service layer: Fixed registry service to use correct method for isLatest updates
- Memory database: Refactored List() function complexity and added missing filters
- Documentation: Updated OpenAPI spec and server.json schema to reflect metadata separation
- Test coverage: Updated mocks to return correct ServerResponse structure
- Migrated the seed.json data with 358 servers to new metadata structure

## How Has This Been Tested?

All tests pass locally and also asked Claude Code to test the API endpoints extensively

## Breaking Changes

Yes, API responses will be different + changes on server.json (no official registry extensions + status field)

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**What's left:**
- [x] Update the publisher cli to allow for setting the status
- [x] Should we bump the schema version?

### DIFF
--- a/tests/integration/main.go
+++ b/tests/integration/main.go
@@ -307,7 +307,7 @@ func getExamples(path string) ([]example, error) {
 	return examples, nil
 }
 
-func compareServerJSON(expected, actual *apiv0.ServerJSON, _ *apiv0.ServerResponse) error {
+func compareServerJSON(expected, actual *apiv0.ServerJSON, actualResponse *apiv0.ServerResponse) error {
 	// Compare core fields (ignore Meta as it contains registry-generated data)
 	if expected.Name != actual.Name {
 		return fmt.Errorf("name mismatch: expected %q, got %q", expected.Name, actual.Name)
@@ -316,7 +316,10 @@ func compareServerJSON(expected, actual *apiv0.ServerJSON, _ *apiv0.ServerRespon
 		return fmt.Errorf("description mismatch: expected %q, got %q", expected.Description, actual.Description)
 	}
 
-	// Status is now in registry metadata, not in server.json - skip status check for now since examples may not have status
+	// Status is in registry metadata - new servers should default to "active"
+	if actualResponse.GetStatus() != "active" {
+		return fmt.Errorf("status mismatch: expected %q, got %q", "active", actualResponse.GetStatus())
+	}
 
 	if expected.Version != actual.Version {
 		return fmt.Errorf("version mismatch: expected %+v, got %+v", expected.Version, actual.Version)


internal/api/handlers/v0/status.go
```diff
@@ -0,0 +1,109 @@
...
+type UpdateServerStatusRequestBody struct {
+	Status string `json:"status" doc:"New status for the server" enum:"active,deprecated,deleted" example:"deprecated"`
+}
+
+// RegisterStatusEndpoints registers the status management endpoints
+func RegisterStatusEndpoints(api huma.API, registry service.RegistryService, cfg *config.Config) {
+	jwtManager := auth.NewJWTManager(cfg)
+
+	// Update server status endpoint
+	huma.Register(api, huma.Operation{
```
domdomegg: I think we might want to think more about how we let people edit their server statuses - and can be a separate non-breaking change if we do want to introduce this. I'd prefer we get the PR in with the structure changes first, then think through editing.

My preference would probably be to only allow:
- publishers upload in status active
- publishers may update active -> deprecated
- any other transitions require an admin (e.g. publishers may not undeprecate a server, or delete a server)

rdimitrov: Makes perfect sense 👍 I'll revert that part part right away 👍

internal/validators/validators.go
```diff
@@ -354,8 +354,8 @@ func ValidatePublishRequest(req apiv0.ServerJSON, cfg *config.Config) error {
 		return err
 	}
 
-	// Validate registry ownership for all packages if validation is enabled and server is not deleted
-	if cfg.EnableRegistryValidation && req.Status != model.StatusDeleted {
+	// Validate registry ownership for all packages if validation is enabled (status is now in registry metadata, so always validate unless explicitly disabled)
```
domdomegg: IIRC we need to disable validation for deleted servers, as I think edits go through this flow. And this makes it impossible to edit a server to be 'deleted' if the packages are broken elsewhere, which is unideal for admins.

tests/integration/main.go
```diff
@@ -304,17 +307,17 @@ func getExamples(path string) ([]example, error) {
...
 		return fmt.Errorf("name mismatch: expected %q, got %q", expected.Name, actual.Name)
 	}
 	if expected.Description != actual.Description {
 		return fmt.Errorf("description mismatch: expected %q, got %q", expected.Description, actual.Description)
 	}
-	if expected.Status != actual.Status {
-		return fmt.Errorf("status mismatch: expected %q, got %q", expected.Status, actual.Status)
-	}
+
+	// Status is now in registry metadata, not in server.json - skip status check for now since examples may not have status
```
domdomegg: we can probably assert status is active

pree-dew:
@domdomegg let me know if you need any help with database transaction implementation in golang, have worked on these things for quite some time :)

tests/integration/main.go
rdimitrov: Addressed 👍

internal/validators/validators.go
rdimitrov: Addressed 👍

internal/api/handlers/v0/status.go
rdimitrov: Addressed 👍

rdimitrov:
> Generally looks good, although I think we should avoid allowing people to edit the status of their servers for now (or should restrict this to just active -> deprecated).

Thanks, I've addressed the feedback and I'll rebase a bit later in the day to solve the conflicts
 
> I also worry this potentially introduces more race conditions in postgres.go, but we can fix these later. I think maybe we need to think more carefully about how we interact with the database - possibly involves learning how people do db transactions in Golang properly.

+1 👍 In other projects I've been using it similar to what you committed earlier but it's worth to assess if we do use it wherever is needed

@pree-dew - please, happy to get feedback for how to approach this better 👍

docs/reference/server-json/CHANGELOG.md
```diff
@@ -116,7 +136,7 @@ All JSON field names standardized to camelCase. **All existing `server.json` fil
 
 #### Updated Schema Reference
 
-🔗 **Current schema**: https://static.modelcontextprotocol.io/schemas/2025-09-16/server.schema.json
+🔗 **Current schema**: https://static.modelcontextprotocol.io/schemas/2025-09-22/server.schema.json
```
domdomegg, line 139: I think this change might be accidental?

tools/admin/takedown.sh
```diff
@@ -1,17 +0,0 @@
-#!/bin/bash
```
domdomegg, line 1: I think we do want something like this, although not sure what the best way to do it is. Maybe edit should provide a ServerResponse and we apply that? (which is only accessible to admins)

domdomegg: Would be okay to do this as a follow-up

domdomegg:
Just linking things up, here's the Discord thread some of this was discussed in: https://discord.com/channels/1358869848138059966/1417915364289020045

<details><summary>Archive copy</summary>
<p>

tadasant (PulseMCP) — 9/17/25, 5:49 PM
@Pree @Rado (Stacklok) I just had a quick chat with @Adam Jones (Anthropic) and we realized we have been thinking about server.json mutability (and migrations) differently. Might be a larger discussion, let’s have it before we make a final call on the breaking change we’re doing with server IDs; sorry for the thrashing
tadasant (PulseMCP) — 9/17/25, 5:49 PM
I’m about to be on a flight so responses will be delayed
Pree — 9/17/25, 5:55 PM
Sure @tadasant (PulseMCP) Happy to discuss the approach of how to treat server.json going forward.
tadasant (PulseMCP) — 9/17/25, 6:07 PM
At a high level, the question is, which of these long term paths do we choose:
Consider server.json immutable, and so we might at any time be storing and serving a range of different schema versions
Make an effort to perform migrations every time we make changes, especially breaking changes, so that we are always serving only the latest schema version

I feel pretty strongly on (1) and will lay out my reasoning - but am open to being convinced I'm missing something here --

Some points why I like (1):
Immutability allows for ETL processes to only pull in the latest publications (e.g. what happened in the last 24h). In 5y, when there are millions of server.json entries, a schema change could result in 5m mutations and require ETL processes to re-dump the entire database
If you see a server.json file in the wild, you can know exactly where to find its corresponding reference in a registry (it will not have been mutated since that static file was created)

Some points why I don't like (2):
Simple field renames like the one we're doing now are migrateable but we’ll have changes in the future we can’t migrate automatically. I suspect that means we'd end up doing this kind of multi-version support anyway
Migrations are hard to write and test; would rather we not rely on them regularly
Allowing mutability changes the meaning of a publish -- for example, if website_url doesn't exist in v1 of the schema, that is different than publishing a v2 of the schema where you intentionally omit a website_url
It creates more burden and complexity for us; if we mess up a migration, lots of things break. Simpler to maintain an immutable ledger

Might have more thoughts later just wanted to kick it off 
Pree — 9/17/25, 6:32 PM
Some initial thoughts 

I think I align more with the idea of making it immutable. 
Reasons:
What users have published is the source of truth for them at first place, an automatic change in that data is going to create confusion and will make it difficult for users on building anything on top of it.
If any user has not used multiple schemes then there is not need for them to update the consumer side.
Doing db migrations accurately is tricky specially when the data is huge not only from the point of view of correctness but also from the point of view of handling it operationally without having a downtime. It also raise risk of inconsistencies in data when anything goes wrong which is a big problem to handle.
Handling multiple schemas on user side using software instead of handling at database layer allows us to have flexibility to handle cases like optional fields, default values and it’s less risky and easy to fix.
 
Jonathan Hefner — 9/17/25, 6:34 PM
I agree with your reasoning.  I think the following point is particularly salient:

Simple field renames like the one we're doing now are migrateable but we’ll have changes in the future we can’t migrate automatically. I suspect that means we'd end up doing this kind of multi-version support anyway

If we don't support multiple schema versions, then we can never add a new required field unless it has a reasonable default value.
Avish (Microsoft) — 9/17/25, 7:09 PM
My two cents - I feel if we need long-term backward compatibility and expect clients to rely on older schemas, we need to serve different schema versions.

The only downside I see is the data can increase rapidly if there are frequent changes in schema. To handle it, maybe we can do following. Might not be the best way, just thinking out loud.
Limit the number of schema versions stored per entry (e.g., keep only the last N versions).
Periodically clean up or archive older versions.
Notify clients when a schema version is deprecated and schedule its removal.
slimslenderslacks — 9/17/25, 7:35 PM
Having them immutable makes a lot of sense (digest-ible). I had been meaning to ask a related question about the current status property in the Server definition. This is one property that seems to be about an existing Server but maybe not part of the Server definition.  It was the only party that seemed undigestible.
In OCI we have the concept of [referrers](https://oras.land/docs/concepts/reftypes/), so that publishers can make additional statements about things without needing to change the thing. Oci references have a whole bunch of other issues 🙂 but I just mention it as an example. 
tadasant (PulseMCP) — 9/17/25, 7:42 PM
Great point on status. Maybe that's something we should pull outside of server.json and into one of those io.modelcontextprotocol/official meta fields
slimslenderslacks — 9/17/25, 7:47 PM
I think that would be better.
I think Runtimes will benefit from the ServerDetails part being content-addressable.
Pree — 9/18/25, 5:57 AM
@tadasant (PulseMCP) @Rado (Stacklok) @Adam Jones (Anthropic) If we are aligning with the idea of making it immutable then should we go ahead with this or there is anything pending to be discussed here?
tadasant (PulseMCP) — 9/18/25, 11:15 AM
Sorry let’s hold until Adam has a chance to respond since we started with definitively opposite opinions here. Won’t be able to deploy the change anyway (need permissions ⁠build(deps): bump anchore/sbom-…⁠)
tadasant (PulseMCP) — 9/18/25, 4:13 PM
Realizing this doesn't really solve the problem because _meta is still technically inside server.json

So either we lift that metadata outside server.json... or server.json is considered immutable except for its _meta fields... the latter of which doesn't seem right

Possible solution: lift this _meta into a ServerDetailInRegistry envelope object that has a server key that is the ServerDetail (server.json), and then stores the _meta info at that higher level. So ServerDetail (server.json) remains immutable (and matches exactly what you would save in your code repository -- a property we don't currently have with the way we're appending _meta[io.modelcontextprotocol/official])

Anyway, I think we should come to consensus on the immutability aspect first, and then can solve this status problem next. I think there is a path to solving it regardless of which direction we go on the first concern. I'm still in favor of immutability for server.json 
Rado (Stacklok) — 9/18/25, 9:39 PM
Just a thought but what do you think about - 

{
  "server": {
    "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
    "name": "io.modelcontextprotocol.anonymous/brave-search",
    "description": "MCP server for Brave Search API integration",
    "websiteUrl": "https://anonymous.modelcontextprotocol.io/examples",
    "repository": {
      "url": "https://github.com/modelcontextprotocol/servers",
      "source": "github"
    },
    "version": "1.0.2",
    "packages": [
      {
        "registryType": "npm",
        "registryBaseUrl": "https://registry.npmjs.org",
        "identifier": "@modelcontextprotocol/server-brave-search",
        "version": "1.0.2",
        "transport": {
          "type": "stdio"
        },
        "environmentVariables": [
          {
            "name": "BRAVE_API_KEY",
            "description": "Brave Search API Key",
            "isRequired": true,
            "isSecret": true
          }
        ]
      }
    ],
    "_meta": {
      "io.modelcontextprotocol.registry/publisher-provided": {
        "brave": {
          "tool": "npm-publisher",
          "version": "1.0.1",
          "build_info": {
            "timestamp": "2023-12-01T10:30:00Z"
          }
        }
      }
    }
  },
    "serverId": "028de86b-0027-4eb9-875b-c40b9b089215",
    "versionId": "0319ba85-54bd-44eb-97a7-58ee04d05b6e",
    "status": "active",
    "publishedAt": "2025-09-17T18:13:49.724857292Z",
    "updatedAt": "2025-09-17T18:13:49.724857292Z",
    "isLatest": true
  }
}
That would be the reply from the registry leaving the registry-specific metadata outside of server.json ensuring its immutability (it's actually similar to what we had when we did the first take of the registry extensions)
tadasant (PulseMCP) — 9/18/25, 10:08 PM
I think that's probably the direction I would go for that concern. Maybe still keeping "io.modelcontextprotocol.registry/official": inside another _meta so the outer shape is MCP-conformant too (and subregistries can add more custom extensions while maintaining MCP conformance), like:

{
  "server": {
    ...
    "_meta": {
      "io.modelcontextprotocol.registry/publisher-provided": {
        "brave": {
          ...
        }
      }
    }
  },
  "_meta": {
      "serverId": "028de86b-0027-4eb9-875b-c40b9b089215",
      ...
    }
  }
}
 
Conceptually, it kind of makes sense that RegistryMCPServerEntry is an envelope of ServerDetail and some other data
Rado (Stacklok) — 9/18/25, 11:01 PM
Good point, I don't mind the additional _meta wrapper if it's needed 👍 I guess we have to also reconsider if there are other properties besides status that would make sense to be moved out of server.json?
slimslenderslacks — 9/19/25, 12:32 AM
In my earlier comment, I thought that ServerDetails was already a sibling to _meta 🙂  But this makes sense to me.  It's clear which section can contain metadata about the Server entry (and might change). What about cache control? So far, it looks like the metadata is not going to change very much.
Adam Jones (Anthropic) — 9/19/25, 2:01 AM
Sorry for taking a while to respond! My rough thoughts are:
Immutability of the data is very important. My main concern here is that a server URL or package version never changes, that would be very weird.
I am much less attached to immutability of the format. I think the biggest pain point here will be the ETL process, although I think breaking changes will be rare enough that it won't be too painful.
I want to make sure it's easy for downstream consumers and clients to consume this data. As someone looking at implementing a subregistry/client myself, I kinda grimace at the thought that I will basically need to convert all these into one format, and would much rather the registry did this for me already.
I acknowledge the problem of adding breaking changes that are not convertable, e.g. required missing fields. I think I don't have a good answer to this other than I think it's unlikely / we'd just have to make it not required / backfill the data / insert a placeholder value - same as migrating a database elsewhere.
Strongly agree re: status being lifted out of the main server.json and put into _meta
Don't mind too much about _meta being adjacent / inside server.json. Appreciate it's a little weird being mutable inside server.json, but also it keeps the structure simpler and flatter (and is similar to what other registries e.g. npm do - example: see _npmOperationalInternal in https://registry.npmjs.org/airtable-mcp-server/latest). I think it is important that _meta is mutable somewhere for subregistries.

Overall while we're in preview and have told people that we might drop all the data anyway, on balance I'd rather do the migration. Especially as I actually expect things to cool down after we do this so we might end up with basically just one format in the end which will be nice for clients.
An alternative I might be happy with is adding a returnAsVersion= parameter which accepts a particular version and converts all the schemas into this format.
I guess there are also two questions to ask here:
do we want to do one migration to sort the above problems out, while we're in preview?
beyond this one, do we want to generally handle non-additive changes via schema migrations or versioning?

I think I am strongly yes for 1, and mild yes for 2.
Adam Jones (Anthropic) — 9/19/25, 2:06 AM
I think this is probably the direction I would go if we wanted more immutability of the server object. I think this looks somewhat similar to what we had before at some point 😅 (although I think this is actually a slight improvement on that - having publisher-provided inside server makes the publishing process easier as then you just have one server.json from the publisher)
Rado (Stacklok) — 9/19/25, 6:37 AM
I think we all agree on the immutability so I would suggest we go with keeping:
registry-owned data outside of server.json
publisher-data inside it

MCP servers are now part of the software supply chain and as such it's expected that the server.json would start to get digest-ed and this is one of the prerequisites for it. 
Rado (Stacklok) — 9/19/25, 6:46 AM
Would it help if I file an issue where I'll try to recap our thoughts so far? I'm happy to do the work too, so we can unblock some of the changes around IDs and so forth.
Pree — 9/19/25, 7:49 AM
@Adam Jones (Anthropic) I agree with this point you mentioned
I want to make sure it's easy for downstream consumers and clients to consume this data. As someone looking at implementing a subregistry/client myself, I kinda grimace at the thought that I will basically need to convert all these into one format, and would much rather the registry did this for me already.

As a consumer who has published servers with different versions or will keep publishing, will expect it to be done by registry itself and not doing it for multiple schemas on their end. All consumers will end up writing own custom version of this implementation if registry will not support it. 
I guess from discussions it looks like we align with immutability but I think its important that we  also take care of making it easier to be consumed. 
tadasant (PulseMCP) — 9/19/25, 2:20 PM
Overall while we're in preview and have told people that we might drop all the data anyway, on balance I'd rather do the migration. 

do we want to do one migration to sort the above problems out, while we're in preview?

I don't mind this - very happy to decide to do the migration in Preview and entirely drop support for the prior schema version. Meant for this discussion to be about the long-term approach, but I'm happy for us to migrate it this go-around while data is still small

beyond this one, do we want to generally handle non-additive changes via schema migrations or versioning?

I'm think I'm still strongly on "versioning" here. I don't disagree with any of your bullet points in your message, but I don't feel they outweigh the other concerns discussed

I think this is probably the direction I would go if we wanted more immutability of the server object. I think this looks somewhat similar to what we had before at some point 😅 (although I think this is actually a slight improvement on that - having publisher-provided inside server makes the publishing process easier as then you just have one server.json from the publisher)

Yeah, I think we got caught up with adding _meta (which I think was the right call) that we didn't thoroughly consider its placement (though to be fair, it's not immediately intuitive that we should split /official from /publisher-provided like it seems we might now)

Would it help if I file an issue where I'll try to recap our thoughts so far? I'm happy to do the work too, so we can unblock some of the changes around IDs and so forth.

Yeah I think cataloging these thoughts, maybe in a Discussion as it's not a single actionable Issue, would be helpful. Or even in the docs somewhere near the design principles.
tadasant (PulseMCP) — 9/19/25, 2:29 PM
Real quick on this:

I want to make sure it's easy for downstream consumers and clients to consume this data. As someone looking at implementing a subregistry/client myself, I kinda grimace at the thought that I will basically need to convert all these into one format, and would much rather the registry did this for me already.

I'm actually not experiencing this pain as much with PulseMCP as I'm implementing. PulseMCP already has a bespoke notion of "what is an MCP server" that aligns with the things we care about. Which is less about e.g. installation instructions and versioning, and more about marketing data and discovery metadata. Because of that, PulseMCP internally will never use any particular version of the server.json schema as its ultimate source of truth. 

I would posit that other sub-registries would treat this similarly? For example, Anthropic might implement a sub-registry UI design that requires a logo for each MCP server. Many, many server.json's are not going to provide that logo. But Anthropic will generate something good-enough as those with missing logos get ingested into the sub-registry. It would be odd to not model this internally as "Every McpServer is required to have exactly one logo", and the server.json's you are ingesting are just an input into your final model, not the actual data you are rendering per-server on your final UI. And your user-facing UI or APIs are not beholden to the change cadence of the upstream MCP Registry.

Maybe there is more than one way to think about this and mine is just one; just flagging that I think in many architectures it will not actually be meaningfully painful for sub-registries to support many different schema versions due to this expected downstream re-transformation of server.json
tadasant (PulseMCP) — 9/19/25, 3:17 PM
@Pree I think we are aligned enough now that we can go ahead and re-add that migration file for this PR. One thing about that migration script: we should also migrate the schema version field

(still blocked on deploying though due to that ongoing deploy/permissions issue, but we can go ahead and get all this merged and create follow up issues for the points raised throughout this discussion)
Pree — 9/19/25, 3:29 PM
@tadasant (PulseMCP) Noted, I will include change of schema version field to newest one.
Adam Jones (Anthropic) — 9/19/25, 6:16 PM
Okay yeah I think I'm happy with:
we are aligned the data should never change
given we are in preview right now, we are okay with breaking changes and migrating the data
once we have GAd, our default position is that we should not migrate old server.json data. instead clients should handle the different versions, which they can see via the $schema key
we should move the _meta keys around as per Tadas's suggestion above.
Rado (Stacklok) — 9/19/25, 8:41 PM
@Adam Jones (Anthropic) @tadasant (PulseMCP) @Pree - hey, to make sure we are synced on the order of these changes and what's needed:
The PR about moving from snake_case to camelCase would be updated to do the migration (it will not change the layout though, just the property names)
TODO: Open a PR changing the layout to what was agreed here (leave out all fields that are not owned by the publisher and can change over time)?
The PR about introducing the version ID and changing the behaviour of the current ID would be rebased on top

Anything I missed and/or can help to move this forward? I'll be flying on Monday and I'll have some time throughout the day, but I'll be relatively unavailable for taking some work after that until I travel back at the end of the week.
tadasant (PulseMCP) — 9/19/25, 8:43 PM
I think that's all correct, though I worry a bit about stacking so much work on a single merge

I know we want to limit the number of breaking changes to a minimal number of merges, but maybe in this case we should push to get the 2 PRs that are almost ready here in, and then we can work the TODO separately? By the time that's ready I suspect we might have another breaking change or two to fold in anyway
Rado (Stacklok) — 9/19/25, 8:50 PM
doing the TODO after all of this sounds good too 👍
Adam Jones (Anthropic) — 9/19/25, 9:27 PM
I kinda would like to stack them together, although appreciate the concerns about doing this in a single merge.

I think traffic is low enough that we could just arrange to merge them in one after the other in quick succession, and that'd maybe be okay?
slimslenderslacks — 9/19/25, 9:39 PM
@Adam Jones (Anthropic) I think we talked about this in a GitHub discussion, but oci package versions can be both tags and digests today, right? 
Adam Jones (Anthropic) — 9/19/25, 9:42 PM
@slimslenderslacks not quite sure where this q is going, but I think that might be a separate versions discussion better for a different thread?

But yes, I believe so.
slimslenderslacks — 9/19/25, 9:42 PM
I've been getting some questions from the field about banning "latest".  To which I always say, well ya ... you should not use latest.  But if you have a latest tag, you'll be able to publish the digest anyway. And then you'll have an immutable server entry that points at immutable content.
Adam Jones (Anthropic) — 9/19/25, 9:43 PM
yeah. it's possible we should enforce this to be a digest only. because it's true for any tag really
slimslenderslacks — 9/19/25, 9:44 PM
Ya, maybe I'll start another thread. I'm just wondering about immutability all the way through to the repository package.
Adam Jones (Anthropic) — 9/19/25, 9:45 PM
also on this, who is owning these 3 things? is it:

@Pree https://github.com/modelcontextprotocol/registry/pull/488
new PR stacked on this?
@tadasant (PulseMCP) https://github.com/modelcontextprotocol/registry/pull/465
slimslenderslacks — 9/19/25, 9:45 PM
But right now, it's enough to just help users understand that they'll be able to create serers that point at digests.
Adam Jones (Anthropic) — 9/19/25, 9:45 PM
oh and I think the version id and current id thing maybe is actually below the first PR, maybe keep it that way for the sake of merge conflicts
Rado (Stacklok) — 9/19/25, 9:51 PM
I'm happy to take the Todo item if we can get the rest done before Monday, otherwise it might take me a bit more time 👍
tadasant (PulseMCP) — 9/19/25, 10:04 PM
I think the only thing missing from @Pree 's PR is the migration file and then we would merge it into the (3) PR

How about I just merge @Pree 's PR into https://github.com/modelcontextprotocol/registry/pull/465 now, Pree can open a PR with the migration separately (branched off #465, merging into #465), and that unblocks @Rado (Stacklok) on branching off of my PR to start that last leg? 
GitHub
[Introduce serverId, rename id to versionId by tadasant · Pull Requ...](https://github.com/modelcontextprotocol/registry/pull/465)
Wielded Claude Code to do this; across many iterations and have reviewed all the code + largely rewritten the PR commentary here.
This PR implements the server ID consistency changes from #396. Our...
Wielded Claude Code to do this; across many iterations and have reviewed all the code + largely rewritten the PR commentary here.
This PR implements the server ID consistency changes from #396. Our...
Pree — 9/19/25, 10:06 PM
Sounds good @tadasant (PulseMCP) I will create a new PR for migration, will be ready by tomorrow.
tadasant (PulseMCP) — 9/19/25, 10:21 PM
Did that - @Rado (Stacklok)  you can go ahead and branch off https://github.com/modelcontextprotocol/registry/pull/465 whenever

I guess you're gonna do a migration too so just watch out merge conflicting with Pree on the migration file name, but otherwise should be straightforward to get this stacked I think
GitHub
[Introduce serverId, rename id to versionId by tadasant · Pull Requ...](https://github.com/modelcontextprotocol/registry/pull/465)
Wielded Claude Code to do this; across many iterations and have reviewed all the code + largely rewritten the PR commentary here.
This PR implements the server ID consistency changes from #396. Our...
Introduce serverId, rename id to versionId by tadasant · Pull Requ...
Rado (Stacklok) — 9/19/25, 10:53 PM
@tadasant (PulseMCP) - sounds great, I'll probably start working on this tomorrow or the day after which should probably give you enough time to complete #465 👍
Adam Jones (Anthropic) — 9/20/25, 1:35 AM
awesome. very happy for you folk to merge it once you've got it all lined up.

feel free to poke me on Discord more if needed.
Pree — Yesterday at 10:48 AM
@tadasant (PulseMCP) @Rado (Stacklok) @Adam Jones (Anthropic) Migration change is pushed 
Rado (Stacklok) — Yesterday at 12:59 PM
Thank you 👍 I'll go ahead and merge the 2 changes then
Rado (Stacklok) — Yesterday at 1:08 PM
Ah, I assumed the migration PR was opened against main so when I merged the PR from Tadas it got closed automatically.
@Adam Jones (Anthropic) @tadasant (PulseMCP) - if you're around - https://github.com/modelcontextprotocol/registry/pull/520
tadasant (PulseMCP) — Yesterday at 2:05 PM
@Rado (Stacklok) maybe I missed something but I thought we were gonna hold off on merging until we had

TODO: Open a PR changing the layout to what was agreed here (leave out all fields that are not owned by the publisher and can change over time)?

lined up?
If so, I think we can't really roll back b/c we already did the data migration unfortunately
Unless someone sees another option, I think we're going to end up doing two separate breaking changes here. 2025-09-16 schema/API version for the camelcase / versionid + another schema/API version for the _meta changes
Rado (Stacklok) — Yesterday at 2:19 PM
Ah, I just re-read the thread and I see I must have gotten it wrong.
Apologies
I'll make sure to file the remaining PR fast enough so we are not for too long in this state 👍
For some reason I assumed we want to have the changes one after another but while also merging them.
tadasant (PulseMCP) — Yesterday at 2:23 PM
All good I think while we are just releasing straight from main without queueing up releases on branches this kind of thing will happen (and probably fine for Preview) 
Pree — Yesterday at 2:24 PM
@Rado (Stacklok) Let me know if you need any help with testing the change/migration, I have the setup ready because of last PR.
Rado (Stacklok) — Yesterday at 2:26 PM
yeah, one of the items I had in mind to chat about was the release process, I think it's fine for now but once we start getting closer to GA it would be good to start doing something like "latest main goes to staging, release of choice goes to prod"
tadasant (PulseMCP) — Yesterday at 2:27 PM
oh do migrations not happen automatically on main merge? 
Pree — Yesterday at 2:30 PM
I think migration run on deployment, can see it went through
Image
tadasant (PulseMCP) — Yesterday at 2:36 PM
yeah data looks like it did. what did you mean by this?

Let me know if you need any help with testing the change/migration, I have the setup ready because of last PR.
oh you are just saying for the next migration, you have a good local setup
(not the one we already merged)
that makes sense
Adam Jones (Anthropic) — Yesterday at 7:49 PM
Just to check we're aligned: we've completed 1 (move to camelCase) and 3 (introduce versionId and serverId). But we haven't done 2 yet (restructure to pull non publisher-provided _meta out of server.json), although we still plan to do this very soon?

If we are going to do it, I think I'd be keen to get 2 in soonish, and then the docs + tooling updated. 
Rado (Stacklok) — Yesterday at 8:12 PM
yes 👍 I'm working on 2 so I should be able to file a PR soon
btw moving the status outside of server.json implied that we have to add a /status endpoint for setting the status (will be available for both publishers and admins)
Rado (Stacklok) — Yesterday at 11:53 PM
Filed this - https://github.com/modelcontextprotocol/registry/pull/527 for the server immutability
GitHub
Ensure server.json is immutable  by rdimitrov · Pull Request #527 ...
Motivation and Context

The following PR refactors the code in order to ensure server.json is immutable.
Core Changes:

Database schema migration: Two-phase migration (007 → 008) to separate immuta...
Ensure server.json is immutable  by rdimitrov · Pull Request #527 ...
also a note for the next time we chat/meet but I would be happy to pick your brains around better e2e testing as in this PR I noticed a few times where the existing tests didn't flagged changes that would have caused user facing issues

</p>
</details>

rdimitrov:
Status update: I'll rebase this on top of #537 so I don't have to re-do the work for the in-memory database 👍

rdimitrov:
This PR should be superseded by https://github.com/modelcontextprotocol/registry/pull/562

PR #528, opened 2025-09-21
Fix race conditions in concurrent server publishing

domdomegg:
## Summary

Fixes race conditions in concurrent server publishing that could result in data integrity violations:
- Multiple versions marked as `isLatest=true` for the same server
- Duplicate (serverId, version) combinations
- Same server name assigned different serverIDs

## Solution

Uses two complementary approaches for correctness:

### 1. PostgreSQL Advisory Locks
- Serializes concurrent publishes per server name using `pg_advisory_xact_lock`
- Uses FNV-1a hash to convert server name to lock ID
- Transaction-scoped locks auto-release on commit/rollback
- **Required after #465**: The serverId introduction means concurrent publishes must atomically coordinate whether to reuse an existing serverId (for additional versions) or generate a new one (for first version). Without locks, race conditions would assign different serverIDs to versions of the same server.

### 2. Unique Constraints (Migration 007)
Two new indexes provide data integrity guarantees:
- **idx_unique_server_version**: Ensures (serverId, version) is unique across all versions
- **idx_unique_latest_version**: Partial index ensuring only one version per serverId has `isLatest=true`

These constraints serve as defense-in-depth safety nets. Even if lock logic had a bug, the database would reject integrity violations.

Also drops redundant `idx_servers_version_id` since `version_id` is already the primary key.

## Why WithPublishLock validation is needed after #465

The `ctx.Err()` check at the start of `WithPublishLock` is critical because:
1. After #465 introduced serverId, the publish flow must atomically check existing versions and assign/reuse serverId
2. If the context is cancelled during lock acquisition, we could deadlock
3. Checking ctx validity before acquiring locks prevents these deadlocks

## Testing

Added `TestPublishConcurrentVersionsNoRace` with 100 concurrent goroutines publishing different versions of the same server. All succeed with:
- Exactly one version marked as latest
- All versions sharing the same serverId
- No duplicate (serverId, version) combinations

NB: this only covers the in memory db, not the postgres db. It'd be nice to write tests for this one as well, but our current setup is a bit annoying for this. Separately looking into ways to easily run all our tests with real postgres and ideally scrap memory.go completely.

### DIFF
--- a/complete.md
+++ b/complete.md
@@ -1,0 +1,2945 @@
+# Complete Documentation
+
+This file contains all documentation from the docs/ directory compiled into a single markdown file.
+
+---
+
+## File: docs/README.md
+
+# MCP Registry Documentation
+
+The MCP registry provides MCP clients with a list of MCP servers, like an app store for MCP servers.
+
+## I want to...
+
+- **📤 Publish my MCP server** → [Publishing Guide](guides/publishing/publish-server.md)
+- **📥 Consume registry data** → [API Usage Guide](guides/consuming/use-rest-api.md)
+- **🔌 Understand the registry's purpose** → [Ecosystem vision](explanations/ecosystem-vision.md)
+- **📋 Look up specific information** → [server.json spec](reference/server-json/generic-server-json.md) | [API spec](reference/api/generic-registry-api.md) | [CLI reference](reference/cli/commands.md)
+
+## Documentation Index
+
+- 🛠️ [How-To Guides: Task-focused instructions](./guides/)
+- 💡 [Explanations: Understanding-oriented content](./explanations/)
+- 📖 [Reference: Technical specifications and lookup material](./reference/)
+
+---
+
+## File: docs/explanations/README.md
+
+# Explanations
+
+Understanding-oriented documentation that clarifies and illuminates the MCP Registry.
+
+- [Design principles](design-principles.md)
+- [Ecosystem vision](ecosystem-vision.md)
+- [Roadmap](roadmap.md)
+- [Tech architecture](tech-architecture.md)
+- [Namespacing](namespacing.md)
+- [Versioning](versioning.md)
+
+---
+
+## File: docs/explanations/design-principles.md
+
+# MCP Registry Design Principles
+
+These are the core constraints that guide the design of the MCP Registry. They are not exhaustive, but they are the most important principles that we will use to evaluate design decisions.
+
+## 1. Single Source of Truth
+
+The registry serves as the authoritative metadata repository for publicly-available MCP servers, both locally-run and remote, open source and closed source. Server creators publish once, and all consumers (MCP clients, aggregators, etc.) reference the same canonical data.
+
+## 2. Minimal Operational Burden
+
+- Design for low maintenance and operational overhead
+- Delegate complexity to existing services where possible (GitHub for auth, npm/PyPI/NuGet for packages)
+- Avoid features that require constant human intervention or moderation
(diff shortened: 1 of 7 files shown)


domdomegg:
Okay, will merge this. I'm not super happy with this implementation: while I think it's an improvement it feels really ugly and still can have cases where we set isLatest on _none_ of the servers if we crash half way through publishing.

Ideally we'd do this in a transaction but this is quite hard to do in go because you can't have generics on interfaces.

I found https://threedots.tech/post/database-transactions-in-go/ which has some ideas, but none feel 'great'.

rdimitrov:
> I found https://threedots.tech/post/database-transactions-in-go/ which has some ideas, but none feel 'great'.

@domdomegg - After reading this I see in the other project I had in mind we were actually using `Transactions in the logic layer` which the author advises to avoid.

Come to think of it for the use case above I guess it would be better to omit setting `isLatest` upon publishing time and rather return it based on the `publishedAt` property?

domdomegg:
> > I found https://threedots.tech/post/database-transactions-in-go/ which has some ideas, but none feel 'great'.
> 
> @domdomegg - After reading this I see in the other project I had in mind we were actually using `Transactions in the logic layer` which the author advises to avoid.

I think `Transactions in the logic layer` might not be _too_ bad, and I'd at least prefer this to the status quo. I'm not convinced the `UpdateFn` works great here in practice because we might be updating many versions of a server.

> Come to think of it for the use case above I guess it would be better to omit setting `isLatest` upon publishing time and rather return it based on the `publishedAt` property?

I think this could work, although I worry it makes lookups potentially quite expensive. i.e. for each server we need to fetch all of its versions every time to figure out if it is `isLatest`. And given we expect publishing to be much rarer than scraping the data from the registry, probably better to make queries cheap and writes expensive.

domdomegg:
Another option I've been thinking about, which does make things a bit more expensive (but not much), is renaming the current table `server_versions` and then having a separate `servers(name, serverId, latestVersion)` table. Then we can join on this hopefully fairly cheaply to figure out if something isLatest. And it forces there to only be one latestVersion AND we can apply a unique constraint on (name) in this table to ensure each server name only corresponds to one ID to avoid race conditions there.

rdimitrov:
> I think this could work, although I worry it makes lookups potentially quite expensive. i.e. for each server we need to fetch all of its versions every time to figure out if it is isLatest.

I was actually thinking to change the query and get 1 entry sorted by descending on publishedAt which should be considered an atomic operation right? Something like - `...ORDER BY publishedAt DESC LIMIT 1`

Also this should work well after the PR about making server.json immutable gets merged since there the publishedAt would be a separate column and we won't have to index through the jsonb entry

domdomegg:
> > I think this could work, although I worry it makes lookups potentially quite expensive. i.e. for each server we need to fetch all of its versions every time to figure out if it is isLatest.
> 
> I was actually thinking to change the query and get 1 entry sorted by descending on publishedAt which should be considered an atomic operation right? Something like - `...ORDER BY publishedAt DESC LIMIT 1`
> 
> Also this should work well after the PR about making server.json immutable gets merged since there the publishedAt would be a separate column and we won't have to index through the jsonb entry

I think this would conflict with the versioning policy supporting semantic versions. We'd need to push this version comparing / ordering somewhere regardless, and while pushing it to the db would be more performant we're still fundamentally having to do a bunch of work for every server returned.

rdimitrov:
Ah, I missed to consider we are preserving the semver-like comparison, I see what you mean now 👍 Hmm..

PR #529, opened 2025-09-22
fix: make seed imports idempotent by ignoring duplicate version_id

iFurySt:
Fix seed imports to be idempotent by handling duplicate version_id conflicts

## Motivation and Context

During seed data imports, duplicate version_id entries could cause database constraint violations and prevent the seed process from completing successfully. This fix makes the seed import
process idempotent by gracefully handling duplicate version_id conflicts using PostgreSQL's `ON CONFLICT DO NOTHING` clause.

## How Has This Been Tested?

- Tested seed import process with existing data in the database
- Verified that duplicate version_id entries are silently ignored without errors
- Confirmed that the seed process can be run multiple times without failures

## Breaking Changes

No breaking changes. This is a backwards-compatible bug fix that improves the robustness of the seed import process.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The fix modifies the `CreateServer` method in `internal/database/postgres.go` to use `ON CONFLICT (version_id) DO NOTHING` when inserting servers. This PostgreSQL-specific syntax ensures that if
a server with the same version_id already exists, the insert operation is silently ignored rather than throwing a constraint violation error.

### DIFF
--- a/internal/database/postgres.go
+++ b/internal/database/postgres.go
@@ -356,6 +356,7 @@ func (db *PostgreSQL) CreateServer(ctx context.Context, server *apiv0.ServerJSON
 	query := `
 		INSERT INTO servers (version_id, value)
 		VALUES ($1, $2)
+		ON CONFLICT (version_id) DO NOTHING
 	`
 
 	_, err = db.pool.Exec(ctx, query, versionID, valueJSON)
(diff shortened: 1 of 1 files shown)


iFurySt:
Xnip2025-09-22_22-39-15

The issue is that every time we run make dev-local, the app runs the seed import again, which tries to insert the same servers and causes duplicate key errors.

You’re right, it was my carelessness, I ended up pushing a bandaid solution. I think the proper fix should be to detect whether the server already exists by version id from the seed file in `importerService.ImportFromPath` rather than at the `INSERT`.

rdimitrov:
Closing this as obsolete as we've moved away from server IDs

PR #530, opened 2025-09-22
Fix atomic latest version update to prevent missing isLatest flags

domdomegg:
## Summary
Fixes a race condition where a server could have no version marked as `isLatest` if the database operations failed independently.

## Changes
- **CreateServer signature**: Now accepts `oldLatestVersionID` parameter to atomically unmark previous latest version
- **PostgreSQL implementation**: Uses a transaction to ensure UPDATE (unmark old latest) and INSERT (create new version) happen atomically
- **Service layer**: Maintains `WithPublishLock` for concurrent operation serialization
- **Two-level protection**: Advisory lock prevents races, transaction ensures atomicity

## Test plan
- [x] Existing race condition tests pass
- [x] Integration tests pass
- [x] Unit tests pass


domdomegg:
Also see https://github.com/modelcontextprotocol/registry/pull/528#issuecomment-3318791248

domdomegg:
(also if not clear: very happy for this to be superseded by a neater implementation! I think the current situation we have is getting pretty messy again. that said, I think we should prioritise on getting the API stable with something like #527 if we are going to do it, to avoid more breaking changes in future. then work on tidying up the implementation)

rdimitrov:
Sharing here so we don't forget: once https://github.com/modelcontextprotocol/registry/pull/537 is merged and we rebase we have to enable the concurrency tests from 1 to 100

pree-dew:
@domdomegg Small suggestion would be here on these points:

1. Rollback should happen when commit fails not always. 
3. Handling rollback failure to identify the right issue, in code we are not handling rollback error which will make debugging hard.
4. Use timeout per transaction so that connection should not stays open.

domdomegg:
Thanks for the suggestions @pree-dew! Some thoughts:

1. **Rollback only on commit failure**: I don't think such a check is necessary with pgx. From [the pgx docs](https://pkg.go.dev/github.com/jackc/pgx#section-readme): "Rollback is safe to call even if the tx is already closed, so if the tx commits successfully, this is a no-op"

2. **Rollback error handling**: If we rollback, that'll be because we've hit some other underlying error. This underlying error that caused the initial failure will already be returned and handled by the caller I think.

3. **Transaction timeout**: Good idea! Have made sure we're passing the context with timeout from `registry_service.go` into the transaction, so the connection won't stay open beyond the timeout.

pree-dew:
@domdomegg May be I should have added more to the scenarios of failure for above points.

1. Rollback error handling is for the cases when the rollback itself fails and we won't be aware about it why it failed. Underlying transaction failing errors and rollback failing errors might be different in many scenarios.

3. Related to transaction timeout, generally it is advisable that each transaction that is happening in a single flow should have it's own transaction timeout that includes rollbacks also so in cases even if one fails then also it should not impact the others. Classic example if even if user cancels the request in middle of transaction, we want rollback to still work to prevent connection dangling and data corruption.

domdomegg:
Ah okay I think I understand what you're saying now and have pushed changes. This fixes 1, and I _think_ fixes your point 2 although I might not be understanding it fully. We create a new context for the database rollback operation so even if the user cancels mid-transaction then the rollback still happens with its own context (rather than trying to use the cancelled context and failing to rollback).

pree-dew:
@domdomegg Yup, you got it right 👍

internal/service/registry_service.go
```diff
@@ -320,12 +313,12 @@ func (s *registryServiceImpl) EditServer(versionID string, req apiv0.ServerJSON)
 	}
 
 	// Check for duplicate remote URLs using the updated server
-	if err := s.validateNoDuplicateRemoteURLs(ctx, updatedServer); err != nil {
+	if err := s.validateNoDuplicateRemoteURLs(ctx, nil, updatedServer); err != nil {
```
crondinini-ant: The fact we use the same `validateNoDuplicateRemoteURLs` function made me think that the EditServer action could modify the server in such a way that it could lead to the same problem being fixed in the Publish endpoint. Is that at all possible?

domdomegg: Good catch! Have wrapped EditServer with a transaction + lock.

PR #537, opened 2025-09-23
Remove the in-memory database

rdimitrov:
## Motivation and Context

The follow PR removes the in-memory database option from the registry.

We had a chat about this with @domdomegg and although it is supposed to be convenient, it isn't that useful in reality. We want to test and develop the registry in a setup that is closer to the environment it is supposed to be run. One of the main reasons are:
* We don't use it anywhere else besides unit testing and potentially local development. Instead we should use postgresql for reliable results (in-memory was actually giving false positive results)
* We have to keep maintaining it and make sure it mimics the behaviour we have for postgresql which is not yielding a lot of value

## How Has This Been Tested?

Yes, I've updated the test unit make target to run in the postgresql environment and now we can see the race condition issue happening.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test-unit: ## Run unit tests with coverage (requires PostgreSQL)
 	@echo "Waiting for PostgreSQL to be ready..."
 	@sleep 3
 	@echo "Running unit tests..."
-	go test -v -race -coverprofile=coverage.out -covermode=atomic -coverpkg=./internal/... ./internal/...
+	go test -v -race -coverprofile=coverage.out -covermode=atomic ./internal/...
 	go tool cover -html=coverage.out -o coverage.html
 	@echo "Coverage report generated: coverage.html"
 	@echo "Stopping PostgreSQL..."
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	go.opentelemetry.io/otel/sdk v1.38.0
 	go.opentelemetry.io/otel/sdk/metric v1.38.0
 	golang.org/x/mod v0.28.0
-	golang.org/x/oauth2 v0.31.0
 )
 
 require (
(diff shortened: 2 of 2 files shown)


domdomegg:
Sounds good, I think this is a good direction go in! Once you get CI passing happy to approve and merge.

We might also want to remove the instructions for non-Docker development from the README. If Docker is required for the tests already anyways, all the issues we get about local dev being broken don't seem worth it to support a non-Docker version.

rdimitrov:
@domdomegg - I've temporarily disabled the concurrency test case since now it's reproduced every time. Also the CI issue is again another conflict that happens because tests are run in parallel and the db state may not be cleared properly. I'll be in meetings most of the day today so I'll try to solve/fix this whenever I can. In case you're available feel free to take over and/or test fixes for the main race condition on top 👍 I'll try to chime in whenever I can 👍

.env.example
```diff
@@ -5,7 +5,7 @@ MCP_REGISTRY_SERVER_ADDRESS=:8080
 MCP_REGISTRY_VERSION=dev
 
 # Database configuration
-# Supported types: postgresql, memory
+# Supported types: postgresql
 MCP_REGISTRY_DATABASE_TYPE=postgresql
```
domdomegg, line 9: can be done in a follow-up: remove this option entirely if it's the only one we support?

tests/integration/run.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+# No persistent state with ephemeral database
```
domdomegg, line 5: nit: probably don't need this comment

Preeti-Dewani:
@domdomegg I am testing it in local after having a word with @rdimitrov . I am okay if you want to take a lead with this otherwise I will continue

domdomegg:
> @domdomegg I am testing it in local after having a word with @rdimitrov . I am okay if you want to take a lead with this otherwise I will continue

Ah sorry just saw this after merging.

Have tested this out locally and seemed to work well for me. Happy to also still hear your feedback on it :)

Issue #482, opened 2025-09-15
Support more key pair algorithms than `ed25519` and storing keys in HSM


joelverhagen:
**Is your feature request related to a problem? Please describe.**

Warranty: I am not a crypto expert. I know just enough to get myself into trouble.

Currently the only key pair algorithm supported for HTTP and DNS based auth is ed25519.

Ideally, it should be possible to use another algorithm that is supported by cloud secret storage providers which have backing HSM. What this would allow is more secure storage of secrets.

Since the public key is in a TXT record, the value must be relatively short, considering the 255 character limit. Perhaps an HTTPS pointer to a public key could be a workaround for length restrictions.

The publish tool should be able to instead of using a local `ed25519` private key on disk, it should be able to reach out to a cloud signing provider, using cloud specific auth. This of course would not need to be in the OSS publish tool but could be a cloud specific publish tool.

However, auth endpoint in the registry would need to be able to evaluate the public key in the DNS TXT record w.r.t to the signed token provided by the client.

**Describe the solution you'd like**

From docs it looks like but Azure Key Vault and AWS KMS support ECDSA. That seems like a good option to me.

Here is some sample code using an ECC key stored in Azure Key Vault:

```csharp
using System.Security.Cryptography;
using System.Text;
using Azure.Identity;
using Azure.Security.KeyVault.Keys;
using Azure.Security.KeyVault.Keys.Cryptography;

Console.WriteLine("Hello, World!");

var credential = new DefaultAzureCredential();
var kvUri = new Uri("https://my-mcp-kv.vault.azure.net/");
var keyUri = new Uri("https://my-mcp-kv.vault.azure.net/keys/mcp");

var ecCryptoClient = new CryptographyClient(keyUri, credential);

byte[] data = Encoding.UTF8.GetBytes("This is some sample data which we will use to demonstrate sign and verify");
byte[] digest = SHA256.HashData(data);

var keyClient = new KeyClient(kvUri, credential);
var key = keyClient.GetKey("mcp");
Console.WriteLine(Convert.ToBase64String(key.Value.Key.X));
Console.WriteLine(key.Value.Key.CurveName);

SignResult ecSignResult = ecCryptoClient.Sign(SignatureAlgorithm.ES256, digest);
Console.WriteLine($"Signed digest using the algorithm {ecSignResult.Algorithm}, with key {ecSignResult.KeyId}. The resulting signature is {Convert.ToBase64String(ecSignResult.Signature)}");
```

[...]

PR #552, opened 2025-09-25
Share more code between DNS and HTTP flows

joelverhagen:
## Motivation and Context

Currently the `dns` and `http` based auth flows share code well in the publisher tool. However there is a lot of duplication in the auth endpoint. This on it's own is not a problem but it does make enhancements to the proof record (the `v=MCPv1; ...` string) harder to keep in sync.

This PR adds a `common.go` file to share logic between the two.

This supports work I am prototyping for https://github.com/modelcontextprotocol/registry/issues/482.

## How Has This Been Tested?

Added new test cases for the permissions. Ran them before and after the refactoring to ensure no behavior change.
Ran existing UTs.
Ran the integration tests.
Manually exercised the flows locally.

## Breaking Changes

None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/api/handlers/v0/auth/common.go
+++ b/internal/api/handlers/v0/auth/common.go
@@ -1,0 +1,204 @@
+package auth
+
+import (
+	"context"
+	"crypto/ed25519"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/modelcontextprotocol/registry/internal/auth"
+	"github.com/modelcontextprotocol/registry/internal/config"
+)
+
+// CoreTokenExchangeInput represents the common input structure for token exchange
+type CoreTokenExchangeInput struct {
+	Domain          string `json:"domain" doc:"Domain name" example:"example.com" required:"true"`
+	Timestamp       string `json:"timestamp" doc:"RFC3339 timestamp" example:"2023-01-01T00:00:00Z" required:"true"`
+	SignedTimestamp string `json:"signed_timestamp" doc:"Hex-encoded Ed25519 signature of timestamp" example:"abcdef1234567890" required:"true"`
+}
+
+// CoreAuthHandler represents the common handler structure
+type CoreAuthHandler struct {
+	config     *config.Config
+	jwtManager *auth.JWTManager
+}
+
+// NewCoreAuthHandler creates a new core authentication handler
+func NewCoreAuthHandler(cfg *config.Config) *CoreAuthHandler {
+	return &CoreAuthHandler{
+		config:     cfg,
+		jwtManager: auth.NewJWTManager(cfg),
+	}
+}
+
+// ValidateDomainAndTimestamp validates the domain format and timestamp
+func ValidateDomainAndTimestamp(domain, timestamp string) (*time.Time, error) {
+	// Validate domain format
+	if !IsValidDomain(domain) {
+		return nil, fmt.Errorf("invalid domain format")
+	}
+
+	// Parse and validate timestamp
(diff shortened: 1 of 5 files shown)


internal/api/handlers/v0/auth/dns.go
```diff
@@ -20,11 +14,7 @@ import (
 
 // DNSTokenExchangeInput represents the input for DNS-based authentication
 type DNSTokenExchangeInput struct {
-	Body struct {
-		Domain          string `json:"domain" doc:"Domain name" example:"example.com" required:"true"`
-		Timestamp       string `json:"timestamp" doc:"RFC3339 timestamp" example:"2023-01-01T00:00:00Z" required:"true"`
-		SignedTimestamp string `json:"signed_timestamp" doc:"Hex-encoded Ed25519 signature of timestamp" example:"abcdef1234567890" required:"true"`
-	}
+	Body CoreTokenExchangeInput
```
domdomegg, line 17: Maybe rename something like `SignatureTokenExchangeInput`?

domdomegg: Maybe rename something like SignatureTokenExchangeInput?

internal/api/handlers/v0/auth/dns.go
```diff
@@ -87,31 +75,14 @@ func RegisterDNSEndpoint(api huma.API, cfg *config.Config) {
...
 // ExchangeToken exchanges DNS signature for a Registry JWT token
 func (h *DNSAuthHandler) ExchangeToken(ctx context.Context, domain, timestamp, signedTimestamp string) (*auth.TokenResponse, error) {
-	// Validate domain format
-	if !isValidDomain(domain) {
-		return nil, fmt.Errorf("invalid domain format")
-	}
-
-	// Parse and validate timestamp
-	ts, err := time.Parse(time.RFC3339, timestamp)
+	_, err := ValidateDomainAndTimestamp(domain, timestamp)
```
domdomegg: It feels like we might be able to merge more of these into one helper that does the whole validation and exchange for a given domain?

E.g. combine ValidateDomainAndTimestamp + DecodeAndValidateSignature

Also maybe combine ParseMCPKeysFromStrings  + VerifySignatureWithKeys

internal/api/handlers/v0/auth/common.go
```diff
@@ -0,0 +1,164 @@
...
+	permissions := []auth.Permission{
+		// Grant permissions for the exact domain (e.g., com.example/*)
+		{
+			Action:          auth.PermissionActionPublish,
+			ResourcePattern: fmt.Sprintf("%s/*", reverseDomain),
+		},
+	}
+
+	if includeSubdomains {
+		// DNS implies a hierarchy where subdomains are treated as part of the parent domain,
```
domdomegg: This comment should probably go into the dns file?

domdomegg (changes requested):
A few minor suggestions but generally looking good!

internal/api/handlers/v0/auth/common.go
joelverhagen: Good point, fixed!

internal/api/handlers/v0/auth/dns.go
joelverhagen: I combined it all. The different parts are how the keys are fetched and whether subdomain permissions are generated. There is also a switch on error message but I did not make that a parameter since I think there is a very sensible default case.

internal/api/handlers/v0/auth/dns.go
joelverhagen: Fixed.

PR #562, opened 2025-09-28
Changes on the server.json format and the API

rdimitrov:
## Motivation and Context

The following PR will address some of the changes we've discussed on Discord along with superseding the PR around ensuring the immutability of server.json (https://github.com/modelcontextprotocol/registry/pull/527). My main reason behind combining these changes is I find it much easier to implement them together instead of separately.

**The goal is to ensure better stability of both the format but also the registry API that will get us closer to GA.**

The changes that will be introduced are:
* Ensure the server.json is immutable. The main updates for this are moving `status` out from `server.json` and have it under the official registry extensions. Also moving the registry extensions out of `server.json` and wrap them together with the server as part of the returned response.
* Streamline the API to use server name and version instead of UUIDs. This can be revisited in the future but let's start with something simple and compliant with the referenced openapi spec. 
* Changes to the integration tests (noticed a few conflicts between the integration setup and running it locally so added extra protections to avoid this)

**What's left:**
- [x] Update the Go types
- [x] Update the database layer (interface + psql implementation + migration file for existing data to the new layout)
- [x] Update the registry service layer
- [x] Update the API handlers
- [x] Update the publisher CLI
- [x] Reflect these to the reference schemas (server.json + openapi)
- [x] Update the seed import feature
- [x] Update the tests and docs
- [x] Address a few small things (mostly polishing)

## Breaking Changes

Migration Summary: UUID -> Server Name/Version

**API Endpoints (Before/After)**

**Before (UUID-based):**

- GET `/v0/servers/{server_id}`
- PUT `/v0/servers/{server_id}`
- GET `/v0/servers/{server_id}/versions`

**After (Name/Version-based):**

- GET `/v0/servers/{server_name}`                           # Latest version
- GET `/v0/servers/{server_name}/versions/{version}`        # Specific version
- GET `/v0/servers/{server_name}/versions`                  # All versions
- PUT `/v0/servers/{server_name}/versions/{version}`        # Edit specific version

Examples:
- GET `/v0/servers/com.example%2Fmy-server (URL-encoded)`
- PUT `/v0/servers/com.example%2Fmy-server/versions/1.2.0?status=deprecated`

  ---
**Server.json Format**

**Before:**
```json
  {
    "name": "com.example/my-server",
    "version": "1.0.0",
    "description": "My MCP server",
    "status": "active",
    "meta": {
      "io.modelcontextprotocol.registry/publisher-provided": {
        "tags": ["database", "postgresql"],
        "category": "data-access",
        "maturity": "beta"
      },
        "publishedAt": "2025-01-01T00:00:00Z",
        "isLatest": true
      }
    }
  }
```
  **After (MCP Spec Only):**
```json
  {
    "name": "com.example/my-server",
    "version": "1.0.0",
    "description": "My MCP server",
    "repository": {
      "url": "https://github.com/user/repo"
    },
    "packages": [...],
    "_meta": {
      "io.modelcontextprotocol.registry/publisher-provided": {
        "tags": ["database", "postgresql"],
        "category": "data-access",
        "maturity": "beta"
      }
    }
  }

```
Key Changes:
- ❌ Removed status and meta.official (now registry-only metadata)
- ✅ Pure MCP specification format for server.json

  ---
**API Response Format**

**Before (Mixed):**
```json
  {
    "id": "uuid-123",
    "name": "com.example/my-server",
    "version": "1.0.0",
    "status": "active",
    "_meta": {
        "publishedAt": "2025-01-01T00:00:00Z",
        "isLatest": true
      },
      "io.modelcontextprotocol.registry/publisher-provided": {
        "tags": ["database", "postgresql"],
        "category": "data-access",
        "maturity": "beta"
      }
  }
```
**After (Separated):**
```json
  {
    "server": {
      "name": "com.example/my-server",
      "version": "1.0.0",
      "description": "My MCP server",
      "repository": {...},
      "packages": [...],
      "_meta": {
        "io.modelcontextprotocol.registry/publisher-provided": {
          "tags": ["database", "postgresql"],
          "category": "data-access",
          "maturity": "beta"
        }
    },
    "_meta": {
        "status": "active",
        "publishedAt": "2025-01-01T00:00:00Z",
        "updatedAt": "2025-01-01T00:00:00Z",
        "isLatest": true
      }
    }
  }
```
Key Changes:
- ✅ Clean separation: server = MCP spec, _meta = registry metadata
- ✅ No more UUIDs - identify by name + version
- ✅ Status changes via query parameter: ?status=deprecated

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/reference/api/openapi.yaml
+++ b/docs/reference/api/openapi.yaml
@@ -22,15 +22,10 @@ paths:
           description: |
             Pagination cursor for retrieving next set of results.
 
-            **Cursor Format**: `{serverName}:{version}`
-
-            Example: `"com.example/my-server:1.0.0"`
-
             Cursors are opaque strings returned in the `metadata.nextCursor` field of paginated responses. Always use the exact cursor value returned by the API.
           required: false
           schema:
             type: string
-            example: "com.example/my-server:1.0.0"
         - name: limit
           in: query
           description: Maximum number of items to return
@@ -316,11 +311,7 @@ components:
               type: string
               description: |
                 Pagination cursor for retrieving the next page of results.
-
-                **Format**: `{serverName}:{version}` (e.g., "com.example/my-server:1.0.0")
-
                 Use this exact value in the `cursor` query parameter of your next request. If null or empty, there are no more results.
-              example: "com.example/my-server:1.0.0"
             count:
               type: integer
               description: Number of items in current page
--- a/docs/reference/api/generic-registry-api.md
+++ b/docs/reference/api/generic-registry-api.md
@@ -30,11 +30,6 @@ All requests and responses use `application/json`
 ### Pagination
 List endpoints use cursor-based pagination for efficient, stable results.
 
-#### Cursor Format
-Pagination cursors use the format: `{serverName}:{version}`
-
-**Example cursor**: `"com.example/my-server:1.0.0"`
-
 #### Usage
 1. **Initial request**: Omit the `cursor` parameter
 2. **Subsequent requests**: Use the `nextCursor` value from the previous response


tadasant (changes requested):
This is awesome, thank you for taking this on!

My review of the Go code is surface-level, but I feel good about the requirements changes based on the examples in the PR comment.

One requested change: We should bump the schema version to `2025-09-29` in `$id` and everywhere it's used, including making that change for all entries in the migration.

docs/reference/api/openapi.yaml
```diff
@@ -99,6 +94,112 @@ paths:
...
+          description: Server or version not found
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: "Server not found"
+    put:
```
domdomegg: I think maybe we can keep this off the main API spec for now? And just have this be for the official registry.

internal/api/handlers/v0/auth/none.go
```diff
@@ -39,7 +39,7 @@ func RegisterNoneEndpoint(api huma.API, cfg *config.Config) {
 		Method:      http.MethodPost,
 		Path:        "/v0/auth/none",
 		Summary:     "Get anonymous Registry JWT",
-		Description: "Get a short-lived Registry JWT token for publishing to io.modelcontextprotocol.anonymous/* namespace",
+		Description: "Get a short-lived Registry JWT token for publishing and editing servers in the io.modelcontextprotocol.anonymous/* namespace",
```
domdomegg: [Follow-up] (optional) May want to clarify the intention is that this is just for local development and automated tests

internal/api/handlers/v0/edit.go
```diff
@@ -31,14 +33,14 @@ func RegisterEditEndpoints(api huma.API, registry service.RegistryService, cfg *
 	huma.Register(api, huma.Operation{
 		OperationID: "edit-server",
 		Method:      http.MethodPut,
-		Path:        "/v0/servers/{server_id}",
+		Path:        "/v0/servers/{server_name}/versions/{version}",
```
domdomegg: [Follow-up] Not a breaking change, but for consistency with other stuff maybe we use `serverName`

internal/api/handlers/v0/servers.go
```diff
@@ -99,60 +105,103 @@ func RegisterServersEndpoints(api huma.API, registry service.RegistryService) {
 		}, nil
 	})
 
-	// Get server details endpoint
+	// Get server details endpoint (latest version)
 	huma.Register(api, huma.Operation{
```
domdomegg, line 109: (optional) I would be tempted to delete this endpoint completely, and just have people use `/v0/servers/{name}/versions/latest`

tools/admin/takedown.sh
```diff
@@ -3,15 +3,28 @@
 
 REGISTRY_URL="${REGISTRY_URL:-https://registry.modelcontextprotocol.io}"
```
domdomegg, line 4: [Follow-up] I think you might also want to update docs/guides/administration/admin-operations.md

internal/database/postgres.go
```diff
@@ -113,33 +112,30 @@ func (db *PostgreSQL) List(
...
 	}
 
-	// Add cursor pagination using primary key version_id
+	// Add cursor pagination using server_name as cursor
 	if cursor != "" {
-		if _, err := uuid.Parse(cursor); err != nil {
-			return nil, "", fmt.Errorf("invalid cursor format: %w", err)
-		}
-		whereConditions = append(whereConditions, fmt.Sprintf("version_id > $%d", argIndex))
+		whereConditions = append(whereConditions, fmt.Sprintf("server_name > $%d", argIndex))
```
domdomegg: I think paginating on server_name alone is a bug - based on how the pagination logic work we'd need to paginate on some stable key. So maybe the cursor could be name + version or something.

domdomegg (changes requested):
I am generally happy with this! Only blocking change is the bug with pagination: https://github.com/modelcontextprotocol/registry/pull/562/files#r2388526356

The UUIDs -> name + version string change I think makes things a lot easier, although might want to give @tadasant some time to review (at least just on this point) as I think he was a proponent of the IDs in the past?

tadasant:
> The UUIDs -> name + version string change I think makes things a lot easier, although might want to give @tadasant some time to review (at least just on this point) as I think he was a proponent of the IDs in the past?

I was at first but have since been convinced otherwise: https://github.com/modelcontextprotocol/registry/discussions/333#discussioncomment-14422185

docs/reference/api/openapi.yaml
rdimitrov: Addressed 🙏

internal/api/handlers/v0/auth/none.go
rdimitrov: Addressed 🙏

internal/api/handlers/v0/edit.go
rdimitrov: Addressed 🙏

tools/admin/takedown.sh
rdimitrov: Addressed 🙏

internal/database/postgres.go
rdimitrov: Thanks! Fixed it 👍

internal/database/postgres.go
```diff
@@ -113,35 +112,45 @@ func (db *PostgreSQL) List(
...
 		}
 	}
 
-	// Add cursor pagination using primary key version_id
+	// Add cursor pagination using compound serverName:version cursor
 	if cursor != "" {
-		if _, err := uuid.Parse(cursor); err != nil {
-			return nil, "", fmt.Errorf("invalid cursor format: %w", err)
+		// Parse cursor format: "serverName:version"
+		parts := strings.SplitN(cursor, ":", 2)
```
domdomegg, line 139: for me to check: is this always safe? i.e. can we have `:` in server names/versions?

docs/reference/api/openapi.yaml
```diff
@@ -19,10 +19,18 @@ paths:
       parameters:
         - name: cursor
           in: query
-          description: Pagination cursor for retrieving next set of results
+          description: |
+            Pagination cursor for retrieving next set of results.
+
+            **Cursor Format**: `{serverName}:{version}`
```
domdomegg: I might drop the note about cursor format from the API spec, and tell consumers to treat it as an opaque string. Just so we have flexibility around the format in future.

internal/database/postgres.go
rdimitrov: I believe we don't allow this on server names and, assuming we follow proper semver, it is not allowed there too, but you have a good point since it was discussed to not enforce semver so potentially this can become an issue. Do you have any suggestions to replace it with another delimiter? Although as long as we allow the version to be everything this wouldn't work either I guess

docs/reference/api/openapi.yaml
rdimitrov: Good point, addressed it 👍

internal/database/postgres.go
domdomegg: Yep I came back to this later, it's not possible to put a `:` in the server name by the server.json spec so this is fine to use as delimiter 👍

I think in the ideal implementation we:
- split at `:`
  - first part is name
  - rest joined with `:` is version string (and for most cases, this should just be one. but if there's more it's them joined by `:`

This should handle all cases reliably right now

---

In the future if we do want to change this, because we're treating it as opaque string we could properly escape it e.g. make this actually a json stringified thing like `{"n":"io.github.domdomegg/airtable-mcp-server","v":"1.2.3"}` (but don't need to do this rn)

domdomegg (approved):

domdomegg:
@rdimitrov I am happy to merge, if you are?

rdimitrov:
> @rdimitrov I am happy to merge, if you are?

Yep, let's go 🙏

internal/database/postgres.go
rdimitrov: Happy to prioritise a follow up if you think we should get this before the freeze?

PR #563, opened 2025-09-28
docs: Add optional GitHub Actions workflow step to show last registration

sualeh:
Add a step to the GitHub Actions workflow to show the results of the publish operation.

## Motivation and Context
It shows the success of the publish operation with the id of the registration on the MCP Server Registry.

## How Has This Been Tested?
By publish my own MCP Server with this workflow step included.

## Breaking Changes
No.

## Types of changes
- [x] Documentation update

## Checklist
- [N/A] My code follows the repository's style guidelines
- [N/A] New and existing tests pass locally
- [N/A] I have added appropriate error handling

## Additional context
None.

### DIFF
--- a/docs/guides/publishing/github-actions.md
+++ b/docs/guides/publishing/github-actions.md
@@ -70,6 +70,22 @@ jobs:
         run: ./mcp-publisher publish
 ```
 
+Optionally, at the end of your workflow, you can show the latest registration for your server as the GitHub Actions workflow job summary. Replace "<<your server name>>" with the name of the server in the YAML snippet below.
+
+```yaml
+      - name: Show MCP registration
+        shell: bash
+        if: always()
+        run: |
+          # Show last registered version
+          curl -s -X GET "https://registry.modelcontextprotocol.io/v0/servers?search=<<your server name>>" \
+            -H "accept: application/json" | \
+            jq '.servers[0] | {name, version, _meta}' > output.json
+          echo '```json' >> $GITHUB_STEP_SUMMARY
+          cat output.json >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+```
+
 ### Step 2: Configure Secrets
 
 You don't need any secrets for publishing to the MCP Registry using GitHub OIDC.


docs/guides/publishing/github-actions.md
```diff
@@ -70,6 +70,22 @@ jobs:
...
 
+Optionally, at the end of your workflow, you can show the latest registration for your server as the GitHub Actions workflow job summary. Replace "<<your server name>>" with the name of the server in the YAML snippet below.
+
+```yaml
+      - name: Show MCP registration
+        shell: bash
+        if: always()
+        run: |
+          # Show last registered version
+          curl -s -X GET "https://registry.modelcontextprotocol.io/v0/servers?search=<<your server name>>" \
```
rdimitrov: I think it's better to use something like `curl -s -f "https://registry.modelcontextprotocol.io/v0/servers/${SERVER_NAME}/versions/latest"`

docs/guides/publishing/github-actions.md
```diff
@@ -70,6 +70,22 @@ jobs:
         run: ./mcp-publisher publish
 ```
 
+Optionally, at the end of your workflow, you can show the latest registration for your server as the GitHub Actions workflow job summary. Replace "<<your server name>>" with the name of the server in the YAML snippet below.
+
+```yaml
+      - name: Show MCP registration
+        shell: bash
+        if: always()
```
rdimitrov, line 78: It might be better for this to be success() instead?

docs/guides/publishing/github-actions.md
```diff
@@ -70,6 +70,22 @@ jobs:
...
+        run: |
+          # Show last registered version
+          curl -s -X GET "https://registry.modelcontextprotocol.io/v0/servers?search=<<your server name>>" \
+            -H "accept: application/json" | \
+            jq '.servers[0] | {name, version, _meta}' > output.json
+          echo '```json' >> $GITHUB_STEP_SUMMARY
+          cat output.json >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
```
jonathanhefner, line 86: I'm unclear on the purpose of this enhancement.  If it is to display the published metadata, why not just `cat server.json`?

Could you elaborate on your use case?

sualeh: This confirms that that last submitted version of the MCP server was actually registered in the registry, and gets the metadata for that registration.

docs/guides/publishing/github-actions.md
sualeh: The report will always show the last registered version, so a visual check can show if the correct (latest) version actually got registered.

docs/guides/publishing/github-actions.md
jonathanhefner: But it's just being logged, not actually confirmed, right?  A human would have to visually inspect the summary and compare it to the version that's in the `server.json` file.

More importantly though, if registration fails, I would expect the build to fail before it reaches this point (e.g., fail at the "Publish to MCP Registry" step).  Have you observed cases where that wasn't true?

sualeh: The build would fail at the point where the registration failed, and the failure would be reported. This is just the status report that shows at the bottom of the GitHub Actions workflow, and acts as an additional information/ debugging step. It is nothing to do with the functionality of registering the server.json per se. We would want the report to show irrespective of whether the registration (and the build) failed or succeeded.

Issue #538, opened 2025-09-23
How to register a specific MCP server from a repository that has multiple MCP servers?


justinyoo:
**Describe the bug**

- I have a repository, say `github.com/contoso/mcp-servers`.
- It contains multiple MCP servers, say `my-mcp-server`, `your-mcp-server` and `their-mcp-server`.
- If I run `mcp-publisher init` under the `my-mcp-server`, it generates the `server.json` file like:

    ```jsonc
    {
      "$schema": "https://static.modelcontextprotocol.io/schemas/2025-07-09/server.schema.json",
      "name": "io.github.contoso/mcp-servers",
      "description": "An MCP server that provides [describe what your server does]",
      "status": "active",
      "repository": {
        "url": "https://github.com/contoso/mcp-servers",
        "source": "github"
      },
      // removed for brevity
    }
    ```

  But to me, it doesn't seem to represent the specific MCP server, `my-mcp-server` in this case.

How can I register a specific MCP server from a repository that contains multiple MCP servers?

MaxKless:
yes, it looks like there's a 1:1 mapping between github repo and mcp server right now, which isn't an assumption that will always hold.
I would love it if we could decouple the github authentication from the mcp server name.

tadasant:
Repositories that contain multiple MCP servers are very much supported; there is a `subfolder` field on `Repository` ([spec](https://github.com/modelcontextprotocol/registry/blob/ae00f70eaf2202970d74f5d459fd174ab8445a8a/docs/reference/server-json/server.schema.json#L7-L37)).

PRs to improve our examples in the docs are very welcome (and/or improvements to default CLI behavior and server.json init process), if someone would like to make that contribution

MaxKless:
thank you for the response @tadasant - I just misunderstood the docs here. After reading the [`server.json`](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md) docs, I saw right away how to configure it properly. I think the tricky thing is that the first part of the name has to match the github org which leads one to believe that the second part has to match the repo, even though it's arbitrary.

What do you think about this change in the `server.json` init process: If `init` is called and creates a `server.json` in a path that isn't the root of the repo, we default to prefilling the mcp name with the parent folder name. 

```
- mcp-servers
  - cool-server
    - src/...
    - server.json <---
- package.json
- ...root files
```

[...]

PR #566, opened 2025-09-29
fix: detect server subfolder in init

MaxKless:
## Motivation and Context
Discussion from this issue: https://github.com/modelcontextprotocol/registry/issues/538
We want to make it clearer that repo name & server name do not HAVE to be the same, especially in monorepos.

@tadasant this seems like a minor thing that doesn't have to be documented specifically, lmk if you disagree :) 

## How Has This Been Tested?
Tested the init flow locally, running the built `init` command from a subfolder and the repo root.

## Breaking Changes
No

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/cmd/publisher/commands/init.go
+++ b/cmd/publisher/commands/init.go
@@ -191,10 +191,8 @@ func buildGitHubServerName(repoURL, subfolder string) string {
 
 	// If we're in a subdirectory, use the current folder name
 	if subfolder != "" {
-		if cwd, err := os.Getwd(); err == nil {
-			folderName := filepath.Base(cwd)
-			return fmt.Sprintf("io.github.%s/%s", owner, folderName)
-		}
+		folderName := filepath.Base(subfolder)
+		return fmt.Sprintf("io.github.%s/%s", owner, folderName)
 	}
 
 	return fmt.Sprintf("io.github.%s/%s", owner, repo)


domdomegg (dismissed):

domdomegg:
Thanks! Are you able to fix the merge conflicts and we can get this shipped 🚀

MaxKless:
awesome, glad to hear it @domdomegg! I rebased and fixed the merge conflicts :)

cmd/publisher/commands/init.go
```diff
@@ -138,6 +180,26 @@ func detectServerName() string {
...
+		return ""
+	}
+
+	owner := parts[3]
+	repo := strings.TrimSuffix(parts[4], ".git")
+
+	// If we're in a subdirectory, use the current folder name
+	if subfolder != "" {
+		if cwd, err := os.Getwd(); err == nil {
+			folderName := filepath.Base(cwd)
```
rdimitrov: It seems we use `subfolder` as a boolean even though it's a string. Wouldn't it be better to reuse the subfolder variable we already have? 

For example if it's != "" construct the path using it (and avoid reading the base path again), otherwise proceed as if we are in the root repo.

MaxKless: you're completely right, I pushed a fix

MaxKless:
Hey! Any update on this? Looks like I got an approval and fixed the comment from the other review so is it ready to be merged?

PR #576, opened 2025-09-29
Updated sse/streamable uri constraints

BobDickinson:
Added token substitution support to SSE transport URIs to fix failing schema validations in public servers (bug fix)

Constrained token substitution to only package transport uris (where there is something to resolve tokens against)

## Motivation and Context
Quite a number of existing servers failed schema validation because of package SSE transports with tokens in their uris. It appears that the intent was to support variable substitution in both streamable and sse transport uris, but the change enabling that was only applied to streamable (where the `description` was updated and the `format: uri` was removed).

It also appears to be the case that variable substitution is only intended to be supported for transports inside of packages where there is something to resolve the variables against.  The original streamable change would have allowed for tokens in remote transport uris as well (which would be unresolvable).

This change makes StreamableHttpTransport and SseTransport work like they used to (restored `format: uri`, and reverted to simplified description).  And it makes new PackageStreamableHttpTransport and PackageSseTransport derived from those that overrides the uri properties with ones that support variable substitution (description, pattern, and examples with variable substitution).

## How Has This Been Tested?
I ran schema validation with a local typescript script against all servers with current schema version using Ajv before and after and verified results.  If there is a tooling in the repo that would be better, please let me know.

## Breaking Changes
No breaking changes

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/reference/server-json/server.schema.json
+++ b/docs/reference/server-json/server.schema.json
@@ -402,7 +402,7 @@
         "url": {
           "type": "string",
           "pattern": "^https?://[^\\s]+$",
-          "description": "Server-Sent Events endpoint URL",
+          "description": "URL template for the sse transport. Variables in {curly_braces} reference argument valueHints, argument names, or environment variable names. After variable substitution, this should produce a valid URI.",
           "example": "https://mcp-fs.example.com/sse"
         },
         "headers": {
--- a/docs/reference/server-json/server.schema.json
+++ b/docs/reference/server-json/server.schema.json
@@ -371,6 +371,7 @@
         },
         "url": {
           "type": "string",
+          "pattern": "^https?://[^\\s]+$",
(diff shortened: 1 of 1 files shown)


BobDickinson:
I'm not sure about the convention for updating the schema version when it changes (or who makes that call), so I didn't touch that.  These changes do not break any currently published servers, but it is more strict in that the tokens are now only allowed in streamable transport uris inside of a package (there were no instances of these outside of a package in the published data, but the schema technically would have allowed it and now does not).

BobDickinson:
I somehow missed https://github.com/modelcontextprotocol/registry/pull/570 which looks like the proper solution.  Assuming that gets done soon, this PR should be closed (preferred option).  It that PR isn't adopted, or adopted soon-ish, this change at least addresses current servers failing validation that shouldn't (in a fairly simple way).

toby:
Thank you for this (and your other awesome contributions!). I agree that we should close this PR in favor of #570, which we are actively working on.

PR #583, opened 2025-09-30
Ensure we populate sane defaults for NULL values

rdimitrov:
## Motivation and Context

The following PR tries to ensure sane defaults for entries that for one reason or another have NULL values for some of their properties.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/database/migrations/008_separate_official_metadata.sql
+++ b/internal/database/migrations/008_separate_official_metadata.sql
@@ -39,9 +39,9 @@ BEGIN
             UPDATE servers
             SET
                 status = COALESCE(NULLIF(rec.value->>'status', 'null'), 'active'),
-                published_at = (official_meta->>'publishedAt')::TIMESTAMP WITH TIME ZONE,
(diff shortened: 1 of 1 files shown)


rdimitrov:
@claude - please review this PR while making sure it covers the problem described in https://github.com/modelcontextprotocol/registry/pull/578. Also assess in case there are other edge cases we have missed to consider.

tadasant (approved):

rdimitrov:
@claude - please review it again. Things to keep in mind are:
* Only one version per server can be marked as latest. We need to preserve original `is_latest` where it exists. If a version already has is_latest = true from the JSON metadata, keep it as the latest version, regardless of published_at. Otherwise use published_at to identify the latest version and mark all else as `is_latest=false`.
* If present, `status` should be preserved too. Otherwise it should default to "active"

domdomegg:
Just confirming that the other day we also reset the staging database so it is up to date with this migration 👍

PR #587, opened 2025-09-30
fix: prevent NULL status values during migration 008

tadasant:
## Summary
- Fixes migration 008 failing with "column contains null values" error
- Adds COALESCE protection to handle 126 servers without status fields
- Includes test utilities to reproduce and verify fixes with production data

## Problem

Migration 008 was failing in production because:
- 126 out of 781 servers have no `status` field in their JSON
- When extracting `rec.value->>'status'`, these became NULL values
- The `ALTER TABLE servers ALTER COLUMN status SET NOT NULL` constraint then failed

## Solution

1. **Prevent NULLs during migration**: Use COALESCE to provide 'active' default when status is NULL/empty
2. **Enhanced debugging**: Added RAISE NOTICE statements to log problematic values
3. **Safety check**: Double-check that throws exception if any NULLs remain
4. **Test utilities**: Created mirror_data tools to test migrations with real production data

## Testing

Created and used `scripts/mirror_data/` tools to:
- Fetch all 781 servers from production API
- Load them into a test database
- Verify the migration fix works correctly

### Results:
- **Before fix**: 126 NULL status values created during migration
- **After fix**: 0 NULL status values, all servers have valid status
- Migration completes successfully with NOT NULL constraint

## Test plan
- [x] Tested locally with production data using mirror_data tools
- [x] Verified 0 NULL values after migration (vs 126 before)
- [x] Confirmed NOT NULL constraint adds successfully
- [ ] Deploy to staging and verify migration succeeds
- [ ] Monitor production deployment

### DIFF
--- a/internal/database/migrations/008_separate_official_metadata.sql
+++ b/internal/database/migrations/008_separate_official_metadata.sql
@@ -38,7 +38,10 @@ BEGIN
             -- Note: status is at top level in old format, not in official_meta
             UPDATE servers
             SET
-                status = rec.value->>'status',
+                status = COALESCE(
+                    NULLIF(NULLIF(rec.value->>'status', ''), 'null'),
+                    'active'
+                ),
                 published_at = COALESCE((official_meta->>'publishedAt')::TIMESTAMP WITH TIME ZONE, NOW()),
                 updated_at = COALESCE((official_meta->>'updatedAt')::TIMESTAMP WITH TIME ZONE, NOW()),
                 is_latest = COALESCE((official_meta->>'isLatest')::BOOLEAN, true)
@@ -47,7 +50,10 @@ BEGIN
             -- Handle records without official metadata (set defaults)
             UPDATE servers
             SET
-                status = rec.value->>'status',
+                status = COALESCE(
+                    NULLIF(NULLIF(rec.value->>'status', ''), 'null'),
+                    'active'
+                ),
                 published_at = NOW(),
                 updated_at = NOW(),
                 is_latest = true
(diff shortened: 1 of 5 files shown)


tadasant:
## Comparison with PR #586

Both PRs solve the same problem but with slightly different approaches:

### PR #586 (rdimitrov's approach)
```sql
-- Adds BEFORE the migration function:
UPDATE servers SET status = 'active' WHERE status IS NULL;

-- In migration function:
status = COALESCE(NULLIF(NULLIF(rec.value->>'status', ''), 'null'), 'active')
```

### This PR (#587)
```sql
-- No pre-update, only in migration function:
status = COALESCE(NULLIF(NULLIF(rec.value->>'status', ''), 'null'), 'active')

-- Plus enhanced safety checks with logging
```

### Analysis

**Similarities:**
- Both use the same COALESCE logic in the migration function
- Both handle empty strings and 'null' strings
- Both default to 'active' when status is missing

**Differences:**
1. **Pre-update**: #586 adds `UPDATE servers SET status = 'active' WHERE status IS NULL` before the migration function runs
   - This is redundant since the column is just created with DEFAULT 'active'
   - At line 11, status column doesn't have data yet, so this UPDATE does nothing

2. **Safety checks**: This PR adds more comprehensive debugging and validation

### Would #586 work?
**Yes**, PR #586 would work. The core fix (COALESCE in the migration function) is identical in both PRs. The pre-update in #586 is harmless but unnecessary since the column is empty when created.

### Recommendation
Either PR would fix the issue. This PR (#587) has additional benefits:
- More detailed logging for debugging
- Test utilities for future migration testing
- Cleaner approach without redundant UPDATE

Both PRs correctly identified and fixed the root cause: servers without status fields in JSON need a default value during migration.

PR #591, opened 2025-09-30
fix: add server name validation and clean invalid data before migration 009

tadasant:
## Summary

This PR adds two important improvements to ensure data integrity:

1. **Adds server name format validation at the API layer** to match the database constraint from migration 008
2. **Includes migration 008 to clean up invalid data** before applying strict constraints in migration 009

### Server Name Validation
- Enforces validation at the API layer matching the database constraint
- **Namespace pattern**: `[a-zA-Z0-9][a-zA-Z0-9.-]*[a-zA-Z0-9]` 
  - Must start and end with alphanumeric characters
  - Can contain dots and hyphens in the middle
- **Name pattern**: `[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9]`
  - Must start and end with alphanumeric characters  
  - Can contain dots, underscores, and hyphens in the middle
- Provides clear, user-friendly error messages indicating which part is invalid
- Uses DRY approach with composed regex patterns

### Migration 008: Data Cleanup
- Removes servers with invalid names or empty versions (5 servers affected in production)
- Fixes invalid status values by setting them to 'active' (1 server affected)
- Removes duplicate name+version combinations
- Includes comprehensive safety checks to prevent accidental data loss
- Must run before migration 009 which adds strict constraints

## Testing
- All validator tests pass
- Migration includes safety checks that match production data
- CI verification included


tadasant:
## ✅ Added Hard Safety Checks

Updated the migration with **exact count validation** to prevent any accidental data loss:

### Safety Measures Added:
1. **Hard failure if DELETE count ≠ 5** - Migration will abort if it would delete anything other than exactly 5 servers
2. **Hard failure if UPDATE count ≠ 1** - Migration will abort if it would update anything other than exactly 1 server
3. **Pre-operation logging** - Shows exactly what will be modified before doing it
4. **Post-operation verification** - Ensures no invalid data remains after cleanup
5. **Transaction safety** - Everything is wrapped in a transaction, so any failure rolls back all changes

### Why We Can Be Confident:

The migration will **ONLY** proceed if it finds exactly:
- **5 servers to delete:**
  - 1 with invalid name: `io.github.joelverhagen/Knapcode.SampleMcpServer/aot`
  - 4 with empty versions: `io.github.timheuer/sampledotnetmcpserver`, `ai.human4/mcp`, `io.github.CDataSoftware/connectcloud-mcp-server`, `io.github.neilberkman/clippy`
- **1 server to update:**
  - `io.github.ptyagiegnyte/egnyte-mcp-server` (status: beta → active)

If the counts don't match exactly, the migration will fail with a clear error message, preventing any unexpected data modifications.

### Example Output:
```sql
NOTICE: Migration 008 Data Cleanup Plan:
NOTICE:   Total servers in database: 788
NOTICE:   Servers to DELETE: 5 (0.63%)
NOTICE:     - Invalid names: 1
NOTICE:     - Empty versions: 4
NOTICE:   Servers to UPDATE (fix status): 1
```

If any count is different, you'll see:
```sql
ERROR: Safety check failed: Expected to delete exactly 5 servers but would delete 7. Aborting to prevent data loss.
```

This makes the migration completely safe - it will either clean up exactly what we expect, or fail without making any changes.

internal/database/migrations/008_clean_invalid_data.sql
```diff
@@ -0,0 +1,182 @@
...
+  AND value->>'status' NOT IN ('active', 'deprecated', 'deleted');
+
+-- Remove duplicate name+version combinations
+-- Keep the one with the highest version_id (most recently added)
+DELETE FROM servers s1
+WHERE EXISTS (
+  SELECT 1 FROM servers s2
+  WHERE s2.value->>'name' = s1.value->>'name'
+    AND s2.value->>'version' = s1.value->>'version'
+    AND s2.version_id > s1.version_id
```
rdimitrov: Assuming this is a uuid I think this one might not work very well, but it's not worth it to change it, we are talking about conflicting server versions here so not a good way to tell which should stay

tadasant: Good call we should used publishedAt

tadasant: https://github.com/modelcontextprotocol/registry/pull/591/commits/3d5ab240e488fd454a6c5df379508d99894f87bc

PR #598, opened 2025-10-01
Enforce schema validation on the publish and edit API level

rdimitrov:
## Motivation and Context

The following PR:
* Enforces schema validation on the API level (requests without schema or older schema will fail)
* Updates the unit tests and examples to match that

## Breaking Changes

**Note this can be considered breaking if clients have not included it before.**

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/validators/validators.go
+++ b/internal/validators/validators.go
@@ -58,7 +58,7 @@ func ValidateServerJSON(serverJSON *apiv0.ServerJSON) error {
 		return fmt.Errorf("$schema field is required")
 	}
 	if !strings.Contains(serverJSON.Schema, model.CurrentSchemaVersion) {
-		return fmt.Errorf("schema version %s is not supported. Please use schema version %s or later", serverJSON.Schema, model.CurrentSchemaVersion)
+		return fmt.Errorf("schema version %s is not supported. Please use schema version %s", serverJSON.Schema, model.CurrentSchemaVersion)
 	}
 
 	// Validate server name exists and format
--- a/internal/validators/validators.go
+++ b/internal/validators/validators.go
@@ -53,6 +53,14 @@ var (
 )
 
 func ValidateServerJSON(serverJSON *apiv0.ServerJSON) error {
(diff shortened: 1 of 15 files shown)


internal/validators/validators.go
```diff
@@ -53,6 +53,14 @@ var (
 )
 
 func ValidateServerJSON(serverJSON *apiv0.ServerJSON) error {
+	// Validate schema version is provided and supported
+	if serverJSON.Schema == "" {
+		return fmt.Errorf("$schema field is required")
+	}
+	if !strings.Contains(serverJSON.Schema, model.CurrentSchemaVersion) {
+		return fmt.Errorf("schema version %s is not supported. Please use schema version %s or later", serverJSON.Schema, model.CurrentSchemaVersion)
```
KalleBylin: Regarding the wording of the error message: `Please use schema version %s or later`

Might be wrong here, but it seems to me that we are doing an exact version check, which would also return an error if a "later" version is used

rdimitrov: thanks! 👍 I've been debating between a few error messages/implementations so of course ended up using the wrong one 😃

internal/validators/validators.go
```diff
@@ -53,6 +53,14 @@ var (
 )
 
 func ValidateServerJSON(serverJSON *apiv0.ServerJSON) error {
+	// Validate schema version is provided and supported
+	if serverJSON.Schema == "" {
```
domdomegg, line 59: rather than this validation here, would it be better to just add to the serverJSON definition? then it would appear e.g. in the autogenerated API docs for us too

domdomegg (changes requested):
As per comment, and fixing CI lint checks

internal/validators/validators.go
rdimitrov: Thanks! Marked it as required, but also decided to leave this check too. My reasoning was the required check happens on the API level, so I thought it's also good to leave it here in case this is called from somewhere else.

PR #608, opened 2025-10-03
Add optional title field to server.json schema

### LINKED ISSUE #606: Add `"title"` property for a server
As discussed at the MCP summit, adding title as a top level property to server.json

This is a human readable title or display name for an MCP server. MCP subregistries or clients MAY choose to use this, but also might not. Many subregistries will take this as a suggestion but may be heavily moderated.

We should also set some standards around this - in general I think we want to not have the suffix 'MCP server'. E.g. 'GitHub' and NOT 'GitHub MCP Server'. And 'Playwright' and not 'Playwright MCP'

domdomegg:
## Summary

Introduces a new optional `title` field for MCP servers to provide a clean, human-readable display name that clients and subregistries can use for display purposes.

Fixes #606

## Changes

- Added `title` field to `ServerJSON` type (optional, max 100 chars)
- Added validation to reject titles ending with 'MCP Server', 'MCP server', or 'MCP' suffixes
- Updated JSON schema and OpenAPI specification
- Added comprehensive test coverage for title validation
- Updated all documentation examples to demonstrate the new field

## Validation

The title field:
- Is optional (can be omitted)
- Must not be only whitespace if provided
- Must not end with MCP-related suffixes (case-insensitive)
- Can contain 'MCP' in the middle (e.g., "MCP Weather API" is valid)

Examples:
- ✅ Valid: "GitHub", "Weather API", "Slack Integration"
- ❌ Invalid: "GitHub MCP", "Weather MCP Server", "Slack-MCP"


domdomegg:
Possibly we can try to autogenerate this as part of the `mcp-publisher init`, think we can do this as a follow-up if necessary

internal/validators/validators.go
```diff
@@ -146,6 +151,35 @@ func validateWebsiteURL(websiteURL string) error {
...
+	if title == "" {
+		return nil
+	}
+
+	// Check that title is not only whitespace
+	if strings.TrimSpace(title) == "" {
+		return fmt.Errorf("title cannot be only whitespace")
+	}
+
+	// Check for MCP suffix variants (case-insensitive)
```
crondinini-ant: Depending on why the MCP suffix is not allowed, it might make sense to block `mcp` as a word completely in the suffix, ignoring the server? So if it says `mcp srvr` or `mcp sv` or `mcp any_word` we'd also block it*

domdomegg: Yeah good point. It's there at the moment to try to help standardize MCP server naming so it's easier for subregistries to process them.

I worry a little about something like 'MCP Gateway'-esque products if we block MCP entirely. But we could do it and then always loosen it more later.

cc @toby if you have thoughts - I know GitHub and VS Code were thinking a lot about standardizing presentation of MCP servers

rdimitrov: While I agree I don't feel it helps having "*-mcp" everywhere I guess there's an argument that to keep upstream neutral we should delegate that on sub-registries to do the filtering/normalising instead of enforcing this centrally?

domdomegg:
Removed the MCP suffix validation based on the feedback. The title field now only validates that it's not entirely whitespace, allowing subregistries to handle naming conventions as they see fit. This keeps the central registry more neutral and flexible.

jonathanhefner:
Has there been a discussion about length limits for title?  I feel like a ~50 character limit would be reasonably unobtrusive and could help mitigate keyword stuffing such as in https://github.com/modelcontextprotocol/registry/discussions/27#discussioncomment-14035301.

domdomegg:
> Has there been a discussion about length limits for title? I feel like a ~50 character limit would be reasonably unobtrusive and could help mitigate keyword stuffing such as in [#27 (reply in thread)](https://github.com/modelcontextprotocol/registry/discussions/27#discussioncomment-14035301).

There's currently a 100 char length limit, which should mitigate this somewhat. We also discussed on Thursday that we'd expect almost all subregistries to do some filtering/moderation of the titles already which would cut out keyword spam.

Happy to reconsider the max length if we think 50 / some other value is better?

jonathanhefner:
> We also discussed on Thursday that we'd expect almost all subregistries to do some filtering/moderation of the titles already which would cut out keyword spam.

I suppose subregistries could just truncate long titles (with "..."). :thinking:  I expect some server authors will be upset that their titles are being presented differently on different subregistries, but I suppose if their titles are that long, they should expect such handling.

I don't feel too strongly about it one way or the other.  50 characters just seemed like a reasonably generous limit, and, if necessary, it would be easier to increase the limit in the future than the reverse.

rdimitrov:
50 characters would probably cover most cases though I think I’m in favour of keeping the limit at 100 mainly since:
* It gives more content for sub-registries to work with when constructing their normalised version of the name (whereas 50 characters might lead to more similarly-sounding names)
* Depending on the use case some names could include domains or prefixes that naturally extend their length

rdimitrov (approved):

domdomegg:
Okay. I'll merge this now - it should be a very easy one-line change to alter this limit if necessary in future etc. Might be a bit annoying for people with long server names (as opposed to lengthening it later) but I suspect it'll probably be fine.

PR #611, opened 2025-10-03
fix: seed from production API by default to ensure validation passes

### LINKED ISSUE #584: Seed data no longer imports due to mcpName requirement for packages
When you start the application from docker-compose, the seed data is imported. Previously this worked fine and you ended up with the 358 servers in your registry. Now all records fail to import as there is new validation for an attribute called mcpName in the packages. 

e.g.
registry  | 2025/09/30 09:27:44 Failed to create server io.github.punkpeye/fastmcp: registry validation failed for package 0 (fastmcp): NPM package 'fastmcp' is missing required 'mcpName' field. Add this to your package.json: "mcpName": "io.github.punkpeye/fastmcp"

i'm not really sure where this extra field is supposed to exist, i've tried multiple configurations.

If there is a new format for publishing or for seed then hopefully the seed.json will be updated?

domdomegg:
Fixes #584

## Summary

Changes the default seed source from local file to production API with filtering to keep startup fast. This ensures local development environments mirror production behavior and all seed data passes validation by default.

## Changes

- **Default seed source**: `https://registry.modelcontextprotocol.io/v0/servers?search=domdomegg` (4 servers)
  - Fast startup (~1-2 seconds)
  - All data passes validation
  - Production-like environment by default
  
- **Updated `data/seed.json`**: Now contains filtered production data (4 servers) for offline development

- **Clear documentation**: Added comments in `docker-compose.yml` and updated `README.md` and `.env.example` explaining:
  - How to use offline mode: `MCP_REGISTRY_SEED_FROM=data/seed.json`
  - How to fetch all servers: `MCP_REGISTRY_SEED_FROM=https://registry.modelcontextprotocol.io/v0/servers`

- **Validation stays enabled**: `MCP_REGISTRY_ENABLE_REGISTRY_VALIDATION=true` by default

## Why this approach

Per @domdomegg's comment in #584, the goal is to make local environments as similar to production as possible. Using the production API with filtering achieves this while keeping startup fast.


rdimitrov (approved):
Thanks for addressing this so quickly! 🙏 

Would it add value if we have a test as part of the CI that ingests the full list of servers in prod (perhaps via some cached actions artifact so we don't query the API upon every PR/commit)? I haven't spent time thinking what use cases that would cover, so just a thought.

domdomegg:
> Would it add value if we have a test as part of the CI that ingests the full list of servers in prod (perhaps via some cached actions artifact so we don't query the API upon every PR/commit)? I haven't spent time thinking what use cases that would cover, so just a thought.

Maybe? Unsure about trade-off with extra demand on registry / whether it would cover much more than we already have in our tests.

PR #629, opened 2025-10-07
Add tooling to auto-generate server.schema.json from OpenAPI spec

domdomegg:
This PR adds a Go tool that extracts the ServerDetail schema from openapi.yaml and generates server.schema.json, ensuring the two specs stay in sync.

### New Tooling
- **Go tool** at `tools/extract-server-schema/` that auto-discovers all schemas referenced by ServerDetail and extracts them from OpenAPI
- **make targets**: `generate-schema` (generates schema) and `check-schema` (verifies sync)
- `check-schema` is now included in `make validate` and `make check`

### Preserved Validation Constraints
The initial auto-generated schema lost many important validation rules and descriptive guidance from the hand-crafted version. The OpenAPI spec has been updated to preserve:

**Security & Descriptions:**
- Command injection security warning on Arguments
- Detailed field descriptions with usage guidance  
- `placeholder` field for Input schema

**Validation Constraints:**
- `fileSha256` pattern validation (`^[a-f0-9]{64}$`)
- Package version restrictions (minLength: 1, no "latest")
- `transport` as required field in Package
- ServerDetail `name` pattern and length constraints (3-200 chars)
- Description, title, and version length limits

**Documentation:**
- Repository field descriptions (url, source, id, subfolder)
- PositionalArgument `valueHint` explains transport URL substitution
- Updated examples to match hand-crafted version (Weather API theme)

**Intentional Changes:**
- Removed `additionalProperties` constraints for flexibility
- Updated schema title to be more descriptive
- Added `$comment` noting auto-generation

See the commit "Preserve validation constraints in auto-generated server.schema.json" for details.

### Fixed Missing Package.transport
- Added `transport` field to OpenAPI Package schema (was in Go model at `pkg/model/types.go:30` but missing from OpenAPI)
- Split transport types into three separate schemas with proper discriminated unions:
  - `StdioTransport` - Only requires `type: stdio`
  - `StreamableHttpTransport` - Requires `type` and `url`, supports `headers`
  - `SseTransport` - Requires `type` and `url`, supports `headers`
- `Package.transport` uses `anyOf` with all three types
- `ServerDetail.remotes` uses `anyOf` with only remote types (no stdio)

### Other Improvements
- Added `GOTOOLCHAIN=auto` to Makefile for Go 1.25 compatibility
- Added `DRIFT_ANALYSIS.md` documenting all schema drift issues found between specs

## Benefits

[...]

PR #630, opened 2025-10-07
schema: Implemented simulated discriminated union for transports and arguments

BobDickinson:
When using `anyOf` in schema7 and no element matches, schema errors from validations of all child elements are produced, which makes it hard to understand the actual schema error.  This will be increasingly important when we move toward providing a schema validation command in `mcp-publisher` and eventually enforcing schema compliance in `/v0/publish` as I will put forward in a later PR. 

## Motivation and Context
When validating schema and there is any failure in an `anyOf` element, the set of schema validation erros returned makes it difficult to understand the issue.  For example, with the old `anyOf` when validating an SSE transport with no `url` you get these schema validation errors:

1. [error] packages.0.transport.type (schema)
   value must be "stdio"
   Reference: #/definitions/StdioTransport/properties/type/enum

2. [error] packages.0.transport (schema)
   missing required fields: 'url'
   Reference: #/definitions/StreamableHttpTransport/required

3. [error] packages.0.transport.type (schema)
   value must be "streamable-http"
   Reference: #/definitions/StreamableHttpTransport/properties/type/enum

4. [error] packages.0.transport (schema)
   missing required fields: 'url'
   Reference: #/definitions/SseTransport/require

Schema7 lacks discriminated union support, but we can use if/then/else to achieve something similar (as I have done in this PR).  When validating the same server.json as above using the if/then/else schema in this PR you get only the correct schema error:

1. [error] packages.0.transport (schema)
   missing required fields: 'url'
   Reference: #/definitions/SseTransport/required

I tried a few variations of validating the `type` under `anyOf` to make something a little more human readable, but I could not induce the correct behavior (even when "filtering" on `type` inside the `anyOf`, all `anyOf` elements were still evaluated and produced schema validation errors).

## How Has This Been Tested?
I validated using a new `validate` command of `mcp-publisher` (that is part of an upcoming PR), with specific test scenarios triggering `anyOf` failures (that validation used the Go lib in this project, santhosh-tekuri/jsonschema/v5).

I also ran the same test scenario using my TypeScript [MCP registry validator](https://github.com/TeamSparkAI/ToolCatalog/blob/main/packages/mcp-registry-validator/README.md), which uses Ajv (the standard schema validator for Javascript).

Lastly I used my validator to run the new schema against all published servers and compared the results to the output using the previous schema.

Note: I also made changes to the `uri` format for the StreamableHttpTransport and SseTransport (required for currently published SSE servers that use tokens in their `url` to pass schema validation).  This will likely be replaced by a better solution in https://github.com/modelcontextprotocol/registry/pull/570 but I think it's worth fixing this now while that is pending (given that it impacts currently published servers).

I think this is the last of the schema changes I'm going to suggest based on my work evaluating existing published servers and implementing schema validation (aforementioned upcoming PR).

## Breaking Changes
This schema is functionally identical the the version it replaces with respect to the transport and argument changes.  The only difference is in the schema validation errors produced (currently no code in the project performs schema validation, so I don't consider that a breaking change).

This schema will also now catch a StreamableHttpTransport `uri` that doesn't start with an http scheme - but there are no such servers currently published (and if there were, they would be in error).  And it will allow tokens in SseTransport `uri` properties, allowing some existing published servers to pass that would have failed (but again, I don't think anyone is actually doing schema validation yet).  I suppose if you really wanted it, you could call the failure to allow a non-http streamable uri a breaking change.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [?] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/reference/server-json/server.schema.json
+++ b/docs/reference/server-json/server.schema.json
@@ -87,17 +87,21 @@
           ]
         },
         "transport": {
-          "anyOf": [
-            {
-              "$ref": "#/definitions/StdioTransport"
-            },
-            {
-              "$ref": "#/definitions/StreamableHttpTransport"
-            },
-            {
-              "$ref": "#/definitions/SseTransport"
+          "type": "object",
+          "properties": {
+            "type": {
+              "type": "string",
+              "enum": ["stdio", "streamable-http", "sse"]
             }
-          ],
+          },
+          "required": ["type"],
+          "if": {"properties": {"type": {"const": "stdio"}}},
+          "then": {"$ref": "#/definitions/StdioTransport"},
+          "else": {
+            "if": {"properties": {"type": {"const": "streamable-http"}}},
+            "then": {"$ref": "#/definitions/StreamableHttpTransport"},
+            "else": {"$ref": "#/definitions/SseTransport"}
+          },
           "description": "Transport protocol configuration for the package"
         },
         "runtimeArguments": {
@@ -292,14 +296,17 @@
     },
     "Argument": {
       "description": "Warning: Arguments construct command-line parameters that may contain user-provided input. This creates potential command injection risks if clients execute commands in a shell environment. For example, a malicious argument value like ';rm -rf ~/Development' could execute dangerous commands. Clients should prefer non-shell execution methods (e.g., posix_spawn) when possible to eliminate injection risks entirely. Where not possible, clients should obtain consent from users or agents to run the resolved command before execution.",
-      "anyOf": [
-        {
-          "$ref": "#/definitions/PositionalArgument"
-        },
-        {
-          "$ref": "#/definitions/NamedArgument"
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "enum": ["positional", "named"]
         }
-      ]
+      },
+      "required": ["type"],
+      "if": {"properties": {"type": {"const": "positional"}}},
+      "then": {"$ref": "#/definitions/PositionalArgument"},
+      "else": {"$ref": "#/definitions/NamedArgument"}
     },
     "StdioTransport": {
       "type": "object",
@@ -334,6 +341,7 @@
         },
         "url": {
           "type": "string",
+          "pattern": "^https?://[^\\s]+$",
           "description": "URL template for the streamable-http transport. Variables in {curly_braces} reference argument valueHints, argument names, or environment variable names. After variable substitution, this should produce a valid URI.",
           "example": "https://api.example.com/mcp"
         },
@@ -363,8 +371,8 @@
         },
         "url": {
           "type": "string",
-          "format": "uri",
-          "description": "Server-Sent Events endpoint URL",
+          "pattern": "^https?://[^\\s]+$",
+          "description": "URL template for the sse transport. Variables in {curly_braces} reference argument valueHints, argument names, or environment variable names. After variable substitution, this should produce a valid URI.",
           "example": "https://mcp-fs.example.com/sse"
         },
         "headers": {
@@ -490,14 +498,17 @@
         "remotes": {
           "type": "array",
           "items": {
-            "anyOf": [
-              {
-                "$ref": "#/definitions/StreamableHttpTransport"
-              },
-              {
-                "$ref": "#/definitions/SseTransport"
+            "type": "object",
+            "properties": {
+              "type": {
+                "type": "string",
+                "enum": ["streamable-http", "sse"]
               }
-            ]
+            },
+            "required": ["type"],
+            "if": {"properties": {"type": {"const": "streamable-http"}}},
+            "then": {"$ref": "#/definitions/StreamableHttpTransport"},
+            "else": {"$ref": "#/definitions/SseTransport"}
           }
         },
         "_meta": {


domdomegg (commented):
I am undecided about this. I am not used to this syntax in JSON schemas, and worry it potentially makes things more complex.

That said, I also see the benefits to users of having the schema validation give clearer errors.

Though for what its worth good validators could figure this out already, e.g. TypeScript generated off the JSON schema does already work.

BobDickinson:
> I am undecided about this. I am not used to this syntax in JSON schemas, and worry it potentially makes things more complex.

> Though for what its worth good validators could figure this out already, e.g. TypeScript generated off the JSON schema does already work.

I share the concern.  I really tried to find a cleaner syntax that would work.  For example, you would think this might work:

```json
{
  "transport": {
    "oneOf": [
      {
        "type": "object",
        "properties": {
          "type": { "const": "stdio" }
        },
        "required": ["type"],
        "$ref": "#/definitions/StdioTransport"
      },
      {
        "type": "object", 
        "properties": {
          "type": { "const": "sse" }
        },
        "required": ["type"],
        "$ref": "#/definitions/SseTransport"
      },
      {
        "type": "object",
        "properties": {
          "type": { "const": "streamable-http" }
        },
        "required": ["type"],
        "$ref": "#/definitions/StreamableHttpTransport"
      }
    ]
  }
}
```

But it failed in the same way in both Go and Javascript validation.

I also struggled picking through the detailed schema validation results on both platforms looking for a pattern that would help me filter out results that shouldn't apply (the only limitation being it had to be general purpose - I wasn't ready to special case the situations I happened to know about), and I really couldn't figure out a solution there either.

This was literally my last resort and the only thing I could make work.  I'm going to push for requiring schema validation, and for that to be acceptable I think the schema errors need to be reasonably clean, which is the only reason I'm pursuing this. Otherwise, I prefer the schema as-is. 

If you didn't care about re-using the transport types between packages and remotes, it's possible you could inline them all into one Transport type (for each) and make the discrimination on type work, but that has its own issues in terms of schema hygiene.

BobDickinson:
There are some even more degenerate cases.  I had a server with a named arg with an invalid format (an existing published server that uses "integer" instead of "number").  The schema validation produced the correct error (bad format on a named arg) but not before giving four errors for violating positional argument constraints (because of the oneOf and exhaustive validation without distinction).  Number 10 below is the only semantically meaningful error.

```bash
6. [error] packages.0.packageArguments.0.format (schema)
   value must be one of "string", "number", "boolean", "filepath"
   Reference: #/definitions/Input/properties/format/enum from: [#/definitions/ServerDetail]/properties/packages/items/[#/definitions/Package]/properties/packageArguments/items/[#/definitions/Argument]/anyOf/0/[#/definitions/PositionalArgument]/allOf/0/[#/definitions/InputWithVariables]/allOf/0/[#/definitions/Input]/properties/format/enum

7. [error] packages.0.packageArguments.0.type (schema)
   value must be "positional"
   Reference: #/definitions/PositionalArgument/allOf/1/properties/type/enum from: [#/definitions/ServerDetail]/properties/packages/items/[#/definitions/Package]/properties/packageArguments/items/[#/definitions/Argument]/anyOf/0/[#/definitions/PositionalArgument]/allOf/1/properties/type/enum

8. [error] packages.0.packageArguments.0 (schema)
   missing required fields: 'valueHint'
   Reference: #/definitions/PositionalArgument/allOf/1/anyOf/0/required from: [#/definitions/ServerDetail]/properties/packages/items/[#/definitions/Package]/properties/packageArguments/items/[#/definitions/Argument]/anyOf/0/[#/definitions/PositionalArgument]/allOf/1/anyOf/0/required

9. [error] packages.0.packageArguments.0 (schema)
   missing required fields: 'value'
   Reference: #/definitions/PositionalArgument/allOf/1/anyOf/1/required from: [#/definitions/ServerDetail]/properties/packages/items/[#/definitions/Package]/properties/packageArguments/items/[#/definitions/Argument]/anyOf/0/[#/definitions/PositionalArgument]/allOf/1/anyOf/1/required

10. [error] packages.0.packageArguments.0.format (schema)
   value must be one of "string", "number", "boolean", "filepath"
   Reference: #/definitions/Input/properties/format/enum from: [#/definitions/ServerDetail]/properties/packages/items/[#/definitions/Package]/properties/packageArguments/items/[#/definitions/Argument]/anyOf/1/[#/definitions/NamedArgument]/allOf/0/[#/definitions/InputWithVariables]/allOf/0/[#/definitions/Input]/properties/format/enum
```

Also, could we try to special-case our way around the confusing errors?  Sure, I guess.  But the schema really needs to be able to travel.  Anyone should be able to use the schema with any validator they choose and get reasonable output.

BobDickinson:
OK, so this is probably the most "offends the senses" part of the change:

```json
"transport": {
  "type": "object",
  "required": ["type"],
  "properties": {
    "type": {"enum": ["stdio", "streamable-http", "sse"]}
  },
  "if": {"properties": {"type": {"const": "stdio"}}},
  "then": {"$ref": "#/definitions/StdioTransport"},
  "else": {
    "if": {"properties": {"type": {"const": "streamable-http"}}},
    "then": {"$ref": "#/definitions/StreamableHttpTransport"},
    "else": {"$ref": "#/definitions/SseTransport"}
  },
  "description": "Transport protocol configuration for the package"
},
```

The official docs seem to prefer an `allOf` with multiple `if`/`then` for this case (more than two variants).  It is equivalent schema, just different representation.

https://json-schema.org/understanding-json-schema/reference/conditionals#ifthenelse

And that would look like this:

```json
"transport": {
  "type": "object",
  "required": ["type"],
  "properties": {
    "type": {"enum": ["stdio", "streamable-http", "sse"]}
  },
  "allOf": [
    {
      "if": {"properties": {"type": {"const": "stdio"}}},
      "then": {"$ref": "#/definitions/StdioTransport"}
    },
    {
      "if": {"properties": {"type": {"const": "streamable-http"}}},
      "then": {"$ref": "#/definitions/StreamableHttpTransport"}
    },
    {
      "if": {"properties": {"type": {"const": "sse"}}},
      "then": {"$ref": "#/definitions/SseTransport"}
    }
  ],
  "description": "Transport protocol configuration for the package"
},
```

The latter is maybe a little more human readable, but it's also twice a many lines.  I'm ambivalent, but if others prefer the latter, I'm fine with it.  If we changed to this form, I'd do it all three places for consistency.

I've implemented and tested it locally to make sure it produces the same results.

BobDickinson:
I was unaware of #629 until just now (that's very nice, BTW).  So even if we wanted to do this, it seems like the correct mechanism would be to generate it from the openapi.yaml.  Is that correct @domdomegg?

OpenAPI does support an explicit `discriminator` mechanism that would make this pretty straightforward (and a clear mapping) - generating the simulated discrimination using if/then in the schema from each `discriminator` in the OpenAPI spec (assuming we added those).

Anyway, it's unclear if there is support for this, but if there is, I'm assuming the PR will be a very different shape.  I'd be happy to take a crack (or not, if someone closer to the schema generate code wanted to do it).

domdomegg:
> OpenAPI does support an explicit discriminator mechanism that would make this pretty straightforward (and a clear mapping) - generating the simulated discrimination using if/then in the schema from each discriminator in the OpenAPI spec (assuming we added those).

#629 is pretty basic - just meant to keep the two in sync, and want to avoid special mapping if we can! (Might have misunderstood this - don't have that much context on OpenAPI discriminators).

BobDickinson:
OK, I'm not totally in love with this, but if we were going to do it, this PR is the general idea.  I reverted my manual schema changes and added explicit discrimination to the openapi spec (which does support it) and then modified the schema generation code to generate the allOf/if/then construct when it encounters an openapi discriminator.  The resulting schema produces clean / specific errors in the cases we're discussing here.  

The readability has admittedly suffered somewhat. Though I think readability generally took a hit with the generated schema (which was well worth it IMO), and it's getting complex enough that it's already not a super user-friendly doc.  I would point to the openapi specs or docs tools that display them for readability and double down on the schema validator producing specific, actionable error messages.

BobDickinson:
I'm going to close this PR for the sake of repo hygiene. I get that this is a hard sell. I'll spend some time in the validator code and see if I can algorithmically trim the feedback (I don't think I can do it in a fully general purpose way, but I think I can do it for the couple of cases we have in the schema).

PR #631, opened 2025-10-07
Improve golang type documentation and examples

domdomegg:
Enhanced Go struct tags throughout `pkg/api/v0/types.go` and `pkg/model/types.go` to better align with the OpenAPI specification.

## Changes

- Added comprehensive `doc` tags with detailed field descriptions matching OpenAPI
- Added `example` tags showing valid values for fields (e.g., `"io.github.user/weather"`, `"npx"`)
- Added `format` tags (`uri`, `date-time`) for better type specification
- Added `pattern` and `enum` tags for validation constraints
- Added `placeholder` field to `Input` struct matching OpenAPI spec
- Removed redundant comments in favor of inline struct tags

These changes improve code clarity and make the types more self-documenting, while maintaining compatibility with the OpenAPI schema definitions in `docs/reference/api/openapi.yaml`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,26 +22,12 @@ jobs:
       uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00
       with:
         go-version-file: 'go.mod'
-
-    - name: Cache Go modules
-      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830
-      with:
-        path: |
-          ~/.cache/go-build
-          ~/go/pkg/mod
-        key: ${{ runner.os }}-go-${{ github.sha }}-${{ hashFiles('**/go.sum', '.golangci.yml') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
-
-    - name: Download dependencies
-      run: go mod download
-
-    - name: Install golangci-lint
-      run: |
-        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.4.0
+        cache: true
 
     - name: Run lint
-      run: make lint
+      uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8
+      with:
+        version: v2.4.0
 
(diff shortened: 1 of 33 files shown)


domdomegg:
Most of this came from diffing the specs with:

```bash
# Start the server
make dev-compose

# Wait for server to be ready, then fetch and compare
curl -s http://localhost:8080/openapi.yaml | yq -o json | jq -S > /tmp/localhost-openapi.json
yq -o json docs/reference/api/openapi.yaml | jq -S > /tmp/docs-openapi.json
diff -u /tmp/docs-openapi.json /tmp/localhost-openapi.json > schemas.diff
```

And then going through and trying to align these as much as possible

domdomegg:
I have no idea why go is panicking now: https://github.com/modelcontextprotocol/registry/actions/runs/18327735919/job/52196150090?pr=631

Stacktrace not very helpful, and the changes really don't seem like they would cause this.

rdimitrov:
> I have no idea why go is panicking now: https://github.com/modelcontextprotocol/registry/actions/runs/18327735919/job/52196150090?pr=631
> 
> Stacktrace not very helpful, and the changes really don't seem like they would cause this.

I hope you don't mind I tried fixing this and from what I found it's caused by Huma which apparently panics if both `default` and `omitempty` are set for embedded properties properties. 

For the sake of the tests, I've removed the default labels for the failing properties (IsRequired, IsSecret and IsRepeated, included in my last commit), but this is still present in the reference openapi spec, so we have to decide if we keep it or come with a workaround for this.

domdomegg:
> I hope you don't mind I tried fixing this and from what I found it's caused by Huma which apparently panics if both default and omitempty are set for embedded properties properties.

Sounds good, thanks for fixing! Maybe we should report it upstream to huma as well.

> For the sake of the tests, I've removed the default labels for the failing properties (IsRequired, IsSecret and IsRepeated, included in my last commit), but this is still present in the reference openapi spec

I think it's fine if there's a little divergence - at least this PR brings them much closer.

rdimitrov:
@domdomegg - in light of the huma issue note that to reproduce it this was happening for embedded properties (booleans in our case, but the type might not be related)

PR #633, opened 2025-10-08
Introduce a cron job that syncs the prod base into staging

rdimitrov:
## Motivation and Context

The following PR adds a workflow that should:
* Runs daily but also can be manually triggered
* Drops the staging database and copies the prod database into it

This should improve our release confidence by reducing the likelihood of regressions or unexpected breakages.
 
## How Has This Been Tested?

No

## Breaking Changes

No (I assume no one relies on staging keeping its state)

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/sync-db.yml
+++ b/.github/workflows/sync-db.yml
@@ -1,0 +1,219 @@
+name: Sync Production DB to Staging
+
+on:
+  schedule:
+    # Run daily at 2 AM UTC (during low-traffic hours)
+    - cron: '0 2 * * *'
+  workflow_dispatch:  # Allow manual triggering
+
+permissions:
+  contents: read
+
+jobs:
+  sync-database:
+    name: Sync Prod DB to Staging
+    runs-on: ubuntu-latest
+    environment: staging
+    concurrency:
+      group: sync-staging-database
+      cancel-in-progress: false
+    steps:
+      - name: Authenticate to Google Cloud (Production)
+        uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093
+        with:
+          credentials_json: ${{ secrets.GCP_PROD_SERVICE_ACCOUNT_KEY }}
+
+      - name: Setup Google Cloud SDK
+        uses: google-github-actions/setup-gcloud@aa5489c8933f4cc7a4f7d45035b3b1440c9c10db
+        with:
+          project_id: mcp-registry-prod
+          install_components: gke-gcloud-auth-plugin
+
+      - name: Get prod GKE credentials
+        run: |
+          gcloud container clusters get-credentials mcp-registry-prod \
+            --zone=us-central1-b \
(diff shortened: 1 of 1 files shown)


rdimitrov:
Initially I wanted to leverage the existing backups so we can validate them as well, but from what I found they are whole cluster backups and not solely on the database layer and so this seemed easier. Happy to revisit if some of you have opinions about this 👍

domdomegg:
> Initially I wanted to leverage the existing backups so we can validate them as well, but from what I found they are whole cluster backups and not solely on the database layer and so this seemed easier. Happy to revisit if some of you have opinions about this 👍

Appreciate it might be more difficult, but I think I would be very excited for this. Will enable us to test our backups, and likely allow us to have lower privileges to do the sync (i.e. not having to login to prod db, instead just read from the backups bucket).

domdomegg (approved):
I think I am mostly okay with this, but would appreciate maybe an extra set of eyes on it given the potential security implications.

domdomegg:
@claude can you review, with particular attention to any unintended security or reliability implications. Read about the project from the docs for context about what this is.

rdimitrov:
Yeah, I too lean more towards the backup approach 👍 I'll update the PR to it instead.

PS. I'll dismiss you review just so someone doesn't merge it accidentally

rdimitrov:
@claude - can you review it again

.github/workflows/sync-db.yml
```diff
@@ -0,0 +1,374 @@
...
+          echo "secret_key=$SECRET_KEY" >> $GITHUB_OUTPUT
+
+          echo "✓ Backup credentials extracted from prod"
+
+      - name: Switch to staging cluster
+        uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093
+        with:
+          credentials_json: ${{ secrets.GCP_STAGING_SERVICE_ACCOUNT_KEY }}
+
+      - name: Configure staging cluster access
```
domdomegg, line 70: Nice, this looks really good!

The one thing I'm scared about is accidentally overwriting the prod database in this way. I guess it all kind of hinges on this command.

Is there a way we can have more confidence we're not going to overwrite prod because e.g. this step fails unexpectedly, or we change the workflow later on and don't realize the implications. Maybe @claude has ideas?

domdomegg: We perhaps could also intentionally log out of prod and remove the kubeconfig once we have the database backup

rdimitrov: I've updated the change so we have a few more safety checks + logging out after we got the backup credentials 👍

rdimitrov:
@claude - can you review the changes again assessing the potential risks of accidentally affecting the prod environment? Explore cases like one or more jobs failing, services timing out, etc. We want to be sure we don't expose the prod environment to any potential errors.

PR #634, opened 2025-10-08
Draft: Refactor package types to use dedicated structs per package type (NPM, PyPI, NuGet, OCI, MCPB)

rdimitrov:
## Motivation and Context

The following PR aims to refactor package types to use dedicated structs per package type (NPM, PyPI, NuGet, OCI, MCPB)

**Notes:**
* Opened this PR with the plan (PACKAGE_TYPE_REFACTOR.md) and the initial definition of the structs so we can agree if the approach is okay before moving forward with the implementation. 

## Breaking Changes

Yes

## Types of changes

- [ ] Documentation update

## Checklist


beshkenadze:
Hey! 
Just wondering — why NPM specifically? The package could also come from bunx, deno, or even be a standalone binary, so maybe a more general runtime type would fit better.

rdimitrov:
> Hey! Just wondering — why NPM specifically? The package could also come from bunx, deno, or even be a standalone binary, so maybe a more general runtime type would fit better.

@beshkenadze - hey 👋 

Someone should probably correct me as I'm not that acquainted with the npm ecosystem, but `npm` in our context is the registry (the location where the package is published). From my understanding npx, bunx, deno and so forth are runtimes that fetch & run npm packages from that registry (by default). 

Also this PR solely aim to refactor what is already there and it doesn't introduce any new package types.

beshkenadze:
@rdimitrov thx for the clarification! 

Here’s the current implementation:
```json
{
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-09-29/server.schema.json",
  "packages": [
    {
      "registryType": "npm",
      "registryBaseUrl": "https://registry.npmjs.org",
      "identifier": "@modelcontextprotocol/server-brave-search",
      "version": "1.0.2",
      "fileSha256": "fe333e598595000ae021bd27117db32ec69af6987f507ba7a63c90638ff633ce",
      "runtimeHint": "npx",
      "transport": { "type": "stdio" }
    }
  ]
}
```

What if the package is hosted in a different registry (e.g., GitHub Packages) or executed directly in a TypeScript runtime such as Bun or Deno?

In that case, would registryType: "npm" still make sense, or should we consider a more general value for non-NPM sources?

Also — where would be the best place to discuss package type definitions more broadly (outside this PR)? Maybe in a separate issue or design doc?

rdimitrov:
> What if the package is hosted in a different registry (e.g., GitHub Packages) or executed directly in a TypeScript runtime such as Bun or Deno?
> 
For this case I would expect `registryBaseUrl` has `npm.pkg.github.com` instead of `https://registry.npmjs.org`, right?

> Also — where would be the best place to discuss package type definitions more broadly (outside this PR)? Maybe in a separate issue or design doc?

Yep, I think either Discord or a separate Discussion in GitHub would be best 👍 In any case, thanks for raising this 🙏

docs/reference/server-json/server.schema.json
```diff
@@ -190,7 +274,117 @@
...
+        "mapping": {
+          "mcpb": "#/components/schemas/MCPBPackage",
+          "npm": "#/components/schemas/NPMPackage",
+          "nuget": "#/components/schemas/NuGetPackage",
+          "oci": "#/components/schemas/OCIPackage",
+          "pypi": "#/components/schemas/PyPIPackage"
+        },
+        "propertyName": "registryType"
+      },
+      "oneOf": [
```
asklar: i think this may be limiting - e.g. we need other package types that are not in this list, like "on_device"

pkg/model/types.go
```diff
@@ -16,20 +16,36 @@ type Transport struct {
...
-	FileSHA256           string          `json:"fileSha256,omitempty"`
-	RunTimeHint          string          `json:"runtimeHint,omitempty"`
-	Transport            Transport       `json:"transport,omitempty"`
-	RuntimeArguments     []Argument      `json:"runtimeArguments,omitempty"`
-	PackageArguments     []Argument      `json:"packageArguments,omitempty"`
+	// Identifier is the package identifier:
+	//   - For NPM/PyPI/NuGet: package name or ID
+	//   - For OCI: full image reference (e.g., "ghcr.io/owner/repo:v1.0.0")
+	//   - For MCPB: direct download URL
+	Identifier string `json:"identifier" minLength:"1"`
```
asklar, line 35: could you preserve formatting so diffs are easier to review and we review only what's actually changed :)

docs/reference/server-json/server.schema.json
rdimitrov: I agree, adding other package types is not in the scope of this PR 👍

pkg/model/types.go
rdimitrov: Sure, can you elaborate more on this?

rdimitrov:
> I agree with changing OCI to just having the identifier field and combining everything there, but think we can do that with the current model.
> 
> I don't think we should split the package definitions up like this - I worry it makes it more likely that the definitions drift further from the implementation (because Golang does not support unions) and looking at it from a client perspective it looks like more models to handle. I think also it might make extending it with custom package types for subregistries harder? I haven't looked super closely, but think they'd then not be compliant with this spec from what I can see.

@domdomegg - I realised I forgot to update the title which might have confused you as this was my initial take, i.e. I started with an interface and different go structs for each type, but I quickly realised it's not good since it became too complicated. 

This approach is different. The Go type remains the same and the main change is on the validation part and on the schema so we can use it for another form of validation). Can you check again in case I've mislead you by thinking it's different go types whereas it's only on the schema (so we can ensure OCI doesn't have versions or npm has a file digest for example). 

Edit: Or I was wrong and you did refer to the schema suggesting we should keep it looser and don't imply the separation on its level? _Apologies for going back and forth_

beshkenadze:
> > What if the package is hosted in a different registry (e.g., GitHub Packages) or executed directly in a TypeScript runtime such as Bun or Deno?
> 
> For this case I would expect `registryBaseUrl` has `npm.pkg.github.com` instead of `https://registry.npmjs.org`, right?
> 
> > Also — where would be the best place to discuss package type definitions more broadly (outside this PR)? Maybe in a separate issue or design doc?
> 
> Yep, I think either Discord or a separate Discussion in GitHub would be best 👍 In any case, thanks for raising this 🙏

FYI: https://github.com/modelcontextprotocol/registry/discussions/645

rdimitrov:
Note: I've tested the PR against prod data earlier and the migration seemed to work fine - 
```
⏺ Migration Testing Complete ✅

Successfully tested migration 009 with real production data (1080 servers). Here's the validation summary:

Migration Results:

- Total servers: 1,080
- Servers with packages: 599 (updated by migration)
- Package breakdown:
  - NPM: 420 packages
  - PyPI: 138 packages
  - OCI: 75 packages
  - MCPB: 46 packages
  - NuGet: 20 packages

Validation Checks (All Passed):

1. ✅ All packages have transport field: 0 packages missing transport
2. ✅ OCI canonical format: 0 OCI packages with old version or registryBaseUrl fields
3. ✅ MCPB cleanup: 0 MCPB packages with redundant version or registryBaseUrl fields
4. ✅ Forbidden field removal: 0 non-MCPB packages with fileSha256 field
5. ✅ NPM/PyPI/NuGet unchanged: Still have version and registryBaseUrl as expected

Sample Validation:

- OCI packages: Correctly formatted with canonical references like ghcr.io/victoriametrics-community/mcp-victoriametrics:v1.13.1
- MCPB packages: Retained fileSha256, removed redundant fields
- NPM packages: Unchanged with version and registryBaseUrl preserved

The migration is working correctly on production data!
```

cmd/publisher/commands/init.go
```diff
@@ -267,36 +267,63 @@ func createServerJSON(
...
-	// Determine registry type and base URL
-	var registryType, registryBaseURL string
+	// Create package based on type
+	var pkg model.Package
+
 	switch packageType {
 	case model.RegistryTypeNPM:
-		registryType = model.RegistryTypeNPM
-		registryBaseURL = model.RegistryURLNPM
+		pkg = model.Package{
```
tadasant, line 275: [can be a follow up PR if you agree] I feel like we should use the "flavors" of Package types internally, and just map each of them to a "list of supported registry types" that can be expanded over time as support is added for more registry types. Instead of bespoke recreating each of the flavors on a per-type basis from Package in this switch statement.

docs/reference/api/openapi.yaml
```diff
@@ -337,6 +318,118 @@ components:
           items:
             $ref: '#/components/schemas/KeyValueInput'
 
+    RegistryPackage:
```
tadasant: `RegistryPackage` as a name feels a little overloaded. Maybe `LanguagePackage`?

docs/reference/api/openapi.yaml
```diff
@@ -337,6 +318,118 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/PackageBase'
+        - type: object
+          required:
+            - identifier
+            - fileSha256
+          properties:
+            registryType:
+              type: string
+              const: mcpb
```
tadasant: Can we drop this const in favor of an example?

tadasant: Same for the other consts on the other Package types

internal/validators/registries/pypi.go
```diff
@@ -39,6 +39,11 @@ func ValidatePyPI(ctx context.Context, pkg model.Package, serverName string) err
 		return ErrMissingVersionForPyPi
 	}
 
+	// Validate that MCPB-specific fields are not present
```
tadasant, line 42: We should probably include a unit test on this logic branch (for npm, nuget as well). Unit test could be generically "does not allow fields that aren't explicitly in the LanguagePackage object"

internal/validators/registries/npm.go
```diff
@@ -41,6 +41,11 @@ func ValidateNPM(ctx context.Context, pkg model.Package, serverName string) erro
 		return ErrMissingVersionForNPM
 	}
 
+	// Validate that MCPB-specific fields are not present
+	if pkg.FileSHA256 != "" {
+		return fmt.Errorf("NPM packages must not have 'fileSha256' field - this is only for MCPB packages")
```
domdomegg, line 46: minor nit, but can probably drop `this is only for MCPB packages`, in case we support other packages with this in future. I think error is clear enough without it.

internal/validators/registries/oci_ref_parser.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// ParseOCIReference parses a canonical OCI image reference.
+// Supported formats:
+//   - registry/namespace/image:tag
+//   - registry/namespace/image@digest
+//   - registry/namespace/image:tag@digest
+//   - namespace/image:tag (defaults to docker.io)
+//   - image:tag (defaults to docker.io/library)
+func ParseOCIReference(ref string) (*OCIReference, error) {
```
domdomegg, line 32: I'm a little scared about implementing this ourselves, but see there's lots of test coverage. I think I'd prefer to find a battle tested narrow library for this if there is one, but no biggie if not.

rafaeljusto:
Hi there 👋  Is there a new schema to match these breaking changes?

I'm using https://static.modelcontextprotocol.io/schemas/2025-09-29/server.schema.json, but it reports an issue when adjusting the `server.json`:

image

I couldn't find any new version on the [CHANGELOG](https://github.com/modelcontextprotocol/registry/blob/84387c96053f4fe8aa7b76e1f8352b04c49f6f1f/docs/reference/server-json/CHANGELOG.md). I'm trying to fix the following error when uploading:

> Error: publish failed: server returned status 400: {"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"registry validation failed for package 0 (teamwork/mcp): OCI packages must not have 'registryBaseUrl' field - use canonical reference in 'identifier' instead (e.g., 'docker.io/owner/image:1.0.0')"}]}

domdomegg:
@rdimitrov are you update the changelog + schema as above?

Also maybe we should make the changelogs more discoverable in general? Maybe even send a link to them in 422 responses (when Huma validation fails, which is probably most indicative of an API change).

rdimitrov:
@domdomegg - on it 👍

rafaeljusto:
Thank you! 🙂

rdimitrov:
@rafaeljusto - hey, the updated schema should be available at https://static.modelcontextprotocol.io/schemas/2025-10-11/server.schema.json 👍 

And again, thank you for raising this! Let us know in case there's anything else you notice! Cheers 🙏

rafaeljusto:
@rdimitrov Thank you for the quick fix!

I'm getting a deprecated schema error now, even though I'm using the latest one 🤔 
```
Error: deprecated schema detected: https://static.modelcontextprotocol.io/schemas/2025-10-11/server.schema.json.
Migrate to the current schema format for new servers.
📋 Migration checklist: https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/CHANGELOG.md#migration-checklist-for-publishers
📖 Full changelog with examples: https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/CHANGELOG.md
```

Maybe we should update this variable?
https://github.com/modelcontextprotocol/registry/blob/ba609e1c17affe828105165bab5b76df7995657d/pkg/model/constants.go#L41

rdimitrov:
Yes! I just realised it too, filing a PR in a second 🙏

rdimitrov:
@rafaeljusto - sorry it took a bit longer but decided to add one more change there. Once this completes it should be ready - https://github.com/modelcontextprotocol/registry/actions/runs/18576909491/job/52963984140 🙏

rafaeljusto:
It worked! 🎉  Thanks again!!

PR #641, opened 2025-10-09
Fix sync-db workflow: k8up restore job detection and timing issues

rdimitrov:
## Motivation and Context

The following PR addresses the failing sync job - https://github.com/modelcontextprotocol/registry/actions/runs/18374711984/job/52345853050

**Root Cause**

1. Job detection failure: The workflow was looking for a job with a specific labeling pattern (k8up.io/owned-by=restore), but k8up creates jobs with a naming convention `restore-<restore-name>`
2. Race condition: The cleanup step (with if: always()) was deleting the prod-to-staging-sync-credentials secret even when the restore job hadn't been created yet or was still running

**The Fix**

1. Updated job detection logic: Now searches for jobs by name pattern restore-$RESTORE_NAME instead of relying on labels
2. Added better diagnostics: If the job isn't found, the workflow now shows:
  - Restore resource status
  - Any existing restore jobs
  - k8up operator logs
3. Fixed cleanup race condition: The cleanup step now waits up to 2 minutes for any running restore jobs to complete before deleting credentials
4. Improved error handling: Better logging when restore jobs fail, including checking if credentials still exist

**The changes ensure:**
- The workflow correctly finds the k8up restore job
- Credentials aren't deleted while jobs are still running
- Better visibility into failures for easier debugging

## Types of changes

- [ ] Documentation update

## Checklist


rdimitrov:
@claude - can you review this? The intent is to fix the problem of the failing workflow at https://github.com/modelcontextprotocol/registry/actions/runs/18374711984/job/52345853050

PR #642, opened 2025-10-09
Add support for on-device packages in MCP Registry

asklar:
- Introduced `on_device` registry type for pre-installed servers.
- Added optional `manifest` and `__dirname` fields in package objects.
- Updated validation logic to accept on-device packages without remote checks.
- Enhanced documentation and examples for on-device server usage.

## Types of changes

- [x] Documentation update

## Checklist


asklar:
thanks @domdomegg! Starting with the manifest (+ __dirname) in the server's `_meta` sounds good, we can re-evaluate promoting that based on adoption. 

Ideally, `package` would have its own `_meta` so that we could add it there instead of at the server level. Any idea why that's not present?

I'll update the PR to put it in `server._meta` for now.

rdimitrov:
> 
> Ideally, `package` would have its own `_meta` so that we could add it there instead of at the server level. Any idea why that's not present?
> 
> 
> 
> I'll update the PR to put it in `server._meta` for now.

I think for now we wanted to keep it on the higher level so we don't bloat the format. For example on our side, we handle package specific meta by having a map of extensions under the _meta vendor which we map like - map[package]:{}

domdomegg:
Hmm some of the MCPB tests are failing in CI - are you able to look into these and fix them?

docs/reference/server-json/server.schema.json
```diff
@@ -234,13 +234,14 @@
           "type": "string"
         },
         "registryType": {
-          "description": "Registry type indicating how to download packages (e.g., 'npm', 'pypi', 'oci', 'nuget', 'mcpb')",
```
zooozialmrshdi-afk: +آر

Issue #509, opened 2025-09-18
Enable exporters for resource observability, add kubelet logs in filelog receiver


pree-dew:
@rdimitrov As per discussion here https://github.com/modelcontextprotocol/registry/pull/478, adding todos:

1. Analyse exporters for node and pod resource which include node exporter and metric server.
2. Add filelog receiver for kubelet logs 
3. Analyse kubernetes events for different failure mode.

pree-dew:
@rdimitrov I have started with first point. 
For 3rd point I will add the failure modes and source details first for discussion before enabling it.

rdimitrov:
@pree-dew - I think this looks good 👍 Thank you once again! 🙏

pree-dew:
@rdimitrov Figured out the solution for all requirements here, without using any external exporter. I tried other exporters like cadvisor, kube state , metrics but was looking where with minimum moving components we can get it done.

- Using otel collector to fetch kubernetes events and ship it as log, this will help us to figure out any issue that we see with respect to deployments etc
- Using kubelet stats to get node, pod level resource utilisation metrics for cpu, memory, disk and network. 

I will be raising a PR once I verify these points:
1. What is the cardinality impact of including these metrics and logs 
2. What all failure modes can be captured without missing critical ones.
3. How frequently we want to ship this.
4. Any unbounded label in any of the metric

pree-dew:
@rdimitrov PR [https://github.com/modelcontextprotocol/registry/pull/646](url)

pree-dew:
Pending items:
- Specify all failure modes that will be covered
- Prepare dashboard for resource metrics
- Prepare dashboard for failure modes covered by k8s events.

@rdimitrov

pree-dew:
@domdomegg @rdimitrov Observing on staging as per the PR (https://github.com/modelcontextprotocol/registry/pull/646#issuecomment-3423355961), so far looks good. 

Cardinality is under control and k8s events have started coming as logs. I will observe it for 1 more day so that impact of deployments on cardinality is more clear with stats.

<img width="1056" height="825" alt="Image" src="https://github.com/user-attachments/assets/28814630-9a05-4a4e-a01e-69c40be8fb4a" />


<img width="1056" height="825" alt="Image" src="https://github.com/user-attachments/assets/988f97ba-d303-4522-8d81-5b3cca77efa8" />

pree-dew:
[...]

PR #646, opened 2025-10-09
Add failure modes telemetry

pree-dew:
Telemetry to cover failure modes which are not covered by container logs and metrics for finding resource constraints.

## Motivation and Context
When there is any issue with registry container we should be notified.

## How Has This Been Tested?
- Local seup

## Breaking Changes
- No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
- No additional exporter is used, taken advantage of opentelemetry collector 
- It covers metrics related to resource constraints, currently only limited to default namespace.
- Takes cares of kubernetes events as logs which are the source of figuring out any problem with service, covers all such scenarios where pod is not able to start yet and get missed because there are no container logs for such cases. Limited to default namespace.
- Taken care of daemonset deployment i.e. deploying otel collector as agent by using correct filtering.
- Cardinality contributing factors are only pod ids (but have to observe more), node ids will not increase cardinality as scale up will lead to limited nodes. 
- Shipping of metrics for resources happens every 60s and list of metrics that will be emitted [https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kubeletstatsreceiver/metadata.yaml](url)

- Container errors

Screenshot 2025-10-10 at 1 21 14 AM

- Resource metrics

Screenshot 2025-10-10 at 1 23 51 AM

### DIFF
--- a/deploy/pkg/k8s/monitoring.go
+++ b/deploy/pkg/k8s/monitoring.go
@@ -232,26 +232,11 @@ func deployOtelCollectorDaemonSet(ctx *pulumi.Context, cluster *providers.Provid
 							pulumi.Map{
 								"type":       pulumi.String("regex_parser"),
 								"id":         pulumi.String("extract_metadata_from_filepath"),
-								"regex":      pulumi.String(`^.*\/(?P<namespace>[^_]+)_(?P<pod_name>[^_]+)_(?P<uid>[a-f0-9\-]{36})\/(?P<container_name>[^\._]+)\/(?P<restart_count>\d+)\.log`),
+								"regex":      pulumi.String(`^.*\/[^_]+_[^_]+_(?P<uid>[a-f0-9\-]{36})\/[^\._]+\/(?P<restart_count>\d+)\.log`),
 								"parse_from": pulumi.String("attributes[\"log.file.path\"]"),
 								"cache": pulumi.Map{
 									"size": pulumi.Int(128),
 								},
-							},
-							pulumi.Map{
-								"type": pulumi.String("move"),
-								"from": pulumi.String("attributes.container_name"),
-								"to":   pulumi.String("resource[\"k8s.container.name\"]"),
-							},
-							pulumi.Map{
-								"type": pulumi.String("move"),
-								"from": pulumi.String("attributes.namespace"),
-								"to":   pulumi.String("resource[\"k8s.namespace.name\"]"),
-							},
-							pulumi.Map{
-								"type": pulumi.String("move"),
-								"from": pulumi.String("attributes.pod_name"),
-								"to":   pulumi.String("resource[\"k8s.pod.name\"]"),
 							},
 							pulumi.Map{
 								"type": pulumi.String("move"),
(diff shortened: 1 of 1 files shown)


pree-dew:
Issue https://github.com/modelcontextprotocol/registry/issues/509

pree-dew:
@rdimitrov @domdomegg @tadasant  Is there a possibility where we can run a deployment on staging for some time then push to production? Wanted to check the cardinality number for this release before it goes to production, something like this(screenshot), here it is high because of so many deployments I have done that won't be case for production. I have done very good testing around this but still wanted to see if this is an option.

Screenshot 2025-10-10 at 1 27 54 AM

rdimitrov:
> Is there a possibility where we can run a deployment on staging for some time then push to production? Wanted to check the cardinality number for this release before it goes to production, something like this(screenshot), here it is high because of so many deployments I have done that won't be case for production. I have done very good testing around this but still wanted to see if this is an option.

I believe what we have now is what you describe - all commits to main get deployed to staging and only the latest release gets deployed to prod. Would that work for you or perhaps you'll need something else?

pree-dew:
@rdimitrov That works assuming once we verify the above details then only release happens for production env, basically we might have to wait for other release to be deployed meanwhile.

Just being cautious here, looking at workflow files, looks like everything is getting deployed on production, anywhere that I can check the `latest release` part? just to make sure if this approach will work.

rdimitrov:
You can check the workflows for how the container images are published but overall we use the `main` floating tag for staging and the `latest` floating tag for prod of which:
* The `main` tag follows all commits to main
image

* The `latest` tag follows all released versions, i.e. `v1.x.y`, etc.
image

pree-dew:
@rdimitrov Thank you for clearing this out 🙂 

I should have clarified this earlier, my bad. I am not looking at release process for specific tag, I wanted to get the stats for this branch before this gets deployed to production. In `deploy.yml` every commit to main is deploying to staging and then to production. When this branch will be merged it will get deployed to both environment. But to understand the cardinality as per above screenshot, I have to collect stats from staging env first, verify it and if cardinality is under control the deploy it to production.

domdomegg:
> In deploy.yml every commit to main is deploying to staging and then to production. When this branch will be merged it will get deployed to both environment

Ah yeah, this is how it used to work. But we changed it a couple weeks ago so that deploy only deploys main to staging by default. And to get to production you need to create a GitHub release.

domdomegg (approved):
Happy to try this in staging!

pree-dew:
Super @domdomegg I will do a final testing on my end once again and we can then take it to staging.

pree-dew:
@domdomegg @rdimitrov 
- Did 2-3 deployments, looks under control, out of bound labels are only pod ids. 
- Simulated Image pull back error also, it means events logs are also coming

Screenshot 2025-10-19 at 2 58 53 PM


============================================================



Screenshot 2025-10-19 at 3 00 33 PM




**We can try this on staging**

domdomegg:
Should get deployed to staging automatically :)

pree-dew:
Thank you @domdomegg

PR #655, opened 2025-10-13
feat: add OpenAPI tags for improved SDK generation and API organization

beshkenadze:
## Summary
Adds OpenAPI tags to improve SDK generation and API documentation organization, as discussed in #644.

## Changes
- **Go Code**: Define tag metadata (name + description) in `internal/api/router/router.go` using Huma's `api.OpenAPI().Tags`
- **Reference Spec**: Update `docs/reference/api/openapi.yaml` to match the generated spec
- **Tags Added**: servers, publish, auth, admin, health, ping

## Benefits
- ✅ Improved SDK generation with proper service grouping (e.g., `serversService`, `publishService`)
- ✅ Better API documentation organization in Scalar UI, Swagger UI, and other tools
- ✅ Code-first approach ensures reference spec stays in sync via compliance tests

## Testing
- ✅ Compliance test passes: `go test -v ./internal/api -run TestOpenAPIEndpointCompliance`
- ✅ All 4 reference endpoints verified as implemented
- ✅ TypeScript SDK regenerated successfully with proper tag-based grouping

## Related
Closes #644

### DIFF
--- a/internal/api/router/router.go
+++ b/internal/api/router/router.go
@@ -149,22 +149,6 @@ func NewHumaAPI(cfg *config.Config, registry service.RegistryService, mux *http.
 			Name:        "publish",
 			Description: "Operations for publishing MCP servers to the registry",
 		},
-		{
-			Name:        "auth",
-			Description: "Authentication and authorization endpoints",
-		},
-		{
-			Name:        "admin",
-			Description: "Administrative operations (restricted access)",
-		},
-		{
-			Name:        "health",
-			Description: "Health check and system status endpoints",
-		},
-		{
-			Name:        "ping",
-			Description: "Simple connectivity test endpoint",
-		},
 	}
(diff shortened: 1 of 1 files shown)


beshkenadze:
@rdimitrov Hey!

I’m not sure if manually editing the OpenAPI file is a good idea.

I wrote a small Go script that exports the OpenAPI spec without running the server directly.
Since I’m not very confident in my Go skills and don’t have much experience with it, I’m not sure if it’s worth including.
I can either add it to this PR or submit it as a separate one.

I also added (just for testing) a Scalar UI setup (similar to Swagger UI) and a generated TypeScript SDK.
Scalar UI is quite convenient for testing APIs externally and can be hosted as static files.
If you think it could be useful, I can also submit it in a separate PR.

PR #681, opened 2025-10-16
Pass ldflags to the Dockerfile

rdimitrov:
## Motivation and Context

The following fixes an issue where the ldflags were not passed to the Dockerfile, so the binary cannot populate its version-related variables accordingly

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         uses: sigstore/cosign-installer@d7543c93d881b35a8faa02e8e3605f69b7a1ce62
 
       - name: Install Syft
-        uses: anchore/sbom-action/download-syft@v0.20.6
+        uses: anchore/sbom-action/download-syft@v0.20.7
 
       - name: Run GoReleaser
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -10,7 +10,14 @@ RUN go mod download
 COPY . .
 
 ARG GO_BUILD_TAGS
-RUN go build ${GO_BUILD_TAGS:+-tags="$GO_BUILD_TAGS"} -o /build/registry ./cmd/registry
+ARG VERSION=dev
+ARG GIT_COMMIT=unknown
+ARG BUILD_TIME=unknown
```
domdomegg, line 15: Is it possible to get the git version and build time inside the Dockerfile build process? Just to help keep things standardized between local dev <> CI as much as possible.

rdimitrov: Technically we could but I haven't seen it being the practice (also this way we add a dependency on git inside the Dockerfile which might feel weird). Anyway, your comment brought up something I missed that it is worth adding the same information for local builds so I'll file a follow up on that 🚀

PR #708, opened 2025-10-24
Use ko in favour of docker build

rdimitrov:
## Motivation and Context

The following PR leverages ko instead of docker build for building the container image.

The motivation for the move is that ko has been one of the widely adopted ways to build Go applications, but overall its benefits are:
* Build time is much faster - around 70-80%
* The resulting image size is much smaller (also more secure) - around 50% smaller

The changes in this PR are:
* Updated the workflows for staging and release to prod to use it
* Updated the local development + the integration tests to use it (the integration tests overwrite the base image from chainguards one to alpine mostly because we rely on wget to check the health status and you don't get that in the stripped down one)
 
From the user point of view everything should continue to work as it is.
 
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,8 @@ services:
       - MCP_REGISTRY_ENABLE_REGISTRY_VALIDATION=${MCP_REGISTRY_ENABLE_REGISTRY_VALIDATION:-true}
     ports:
       - 8080:8080
+    volumes:
+      - ./data:/data:ro
     restart: "unless-stopped"
 
   postgres:


domdomegg:
Appreciate they both spit out OCI images, but I'm a bit nervous about moving to ko over Docker - feels potentially less standardized etc. than Docker.

Is the main benefit/motivation local dev speed? If so that's maybe reasonable and the trade-off is worth it. I'm not super convinced that image size / hardening is what we should be optimizing for rn if that's a key reason though.

rdimitrov:
There's the option to use ko for building the prod/staging images and continue to have docker for local builds, but I was concerned that this may introduce some risk of the two getting out of sync (although in our case it's not really that important since we build the project and set 3 variables). 

Personally my main benefits are trying to improve the build time. Image size does help speed up deployment too but in this case I think it's insignificant.

rdimitrov:
Had to dismiss the approval as I noticed we are not including the seed.json as part of the image anymore so for local development we have to mount the `data` folder so the seed.json is available to the container.

Issue #494, opened 2025-09-16
Drop the name-based validation of Remote.url (just validate it is a URL)


ashubham:
**Describe the bug**
When trying to run 
```
mcp-publisher publish
```

We get this error:

```
Publishing to https://registry.modelcontextprotocol.io...
Error: publish failed: server returned status 400: 
{"title":"Bad Request","status":400,"detail":"Failed to publish server",
"errors":[
  {"message":"remote URL https://agent.thoughtspot.app/mcp does not match namespace 
      io.github.thoughtspot/mcp-server: remote URL host agent.thoughtspot.app does not match 
      publisher domain thoughtspot.github.io"
  }
]}
```

While the docs clearly state that when `io.github.companyname` scheme is used, any remote URL is allowed, as long as authenticated to github.

https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md#url-validation-requirements

<img width="848" height="122" alt="Image" src="https://github.com/user-attachments/assets/88b48505-33b6-448d-9653-f4b0b6ac490e" />

**Expected behavior**
This should successfully publish.

tadasant:
Thank you for raising this!

For context, I think this is currently a mistake in the docs (i.e. implementation is as-intended, docs are incorrect):

> For io.github.username/* namespaces: No URL restrictions (but you must authenticate via GitHub)

As designed, our goal was to prevent people from submitting Remote.url's they do not control. So we required that your namespace verification match up with the Remote URL.

_However_, thinking through this further, I see several issues with that goal:
- Someone might namespace their MCP servers like `com.companyname.servers/my-mcp` but the server might be hosted at `mcp.companyname/mcp`
- If you're using a third party hosting service, you can't align
- As you (@ashubham) are pointing out, there is demand for being able to point to a server from an `io.github` namespace. And because we don't allow name mutability (only by creating an entirely new server), this is pretty prohibitive for you.

So I think I'm convinced we should drop this requirement altogether. Just allow people to set `Remote.url` to any valid URL they want; no need for alignment with the reverse DNS namespace. In theory, that could lead to people spamming the registry with many entries under different `name`s for the same `Remote.url`, but in practice I think sub-registries will be good at filtering this kind of thing out, and there will be little/no incentive to spamming in this way.

[...]

PR #728, opened 2025-10-29
Prepare docs for modelcontextprotocol.io

jonathanhefner:
This PR reorganizes and extends the existing documentation in preparation for publishing on https://modelcontextprotocol.io/.

The current plan is for docs to remain in this repo, and be synced to the `modelcontextprotocol/modelcontextprotocol` whenever necessary (e.g., accompanying a release of the registry).  I have included a script in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1725 (`bin/pull-registry-docs.sh`) that can be run manually to perform the sync.

If we later decide to permanently move the docs to the `modelcontextprotocol/modelcontextprotocol` repo, that also works.

This PR puts all https://modelcontextprotocol.io/ docs into `docs/modelcontextprotocol-io/`.  Docs outside that subdirectory, such as `docs/administration/admin-operations.md`, are intended to be accessible solely via this repo.  One major exception is the `docs/reference/server-json/` subdirectory, which will depend on the resolution of https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649.  However, I think we could address that in a follow-up PR.

### DIFF
--- a/docs/reference/server-json/official-registry-requirements.md
+++ b/docs/reference/server-json/official-registry-requirements.md
@@ -10,7 +10,6 @@ While the [generic server.json format](./generic-server-json.md) defines the bas
 
 - **Namespace authentication** - Servers are published under appropriate namespaces
 - **Package ownership verification** - Publishers actually control referenced packages
-- **Remote server URL match** - Remote server base urls match namespaces
 - **Restricted registry base urls** - Packages are from trusted public registries
 - **`_meta` namespace restrictions** - Restricted to `publisher` key only
 
@@ -25,10 +24,6 @@ See the [publishing guide](../../guides/publishing/publish-server.md) for authen
 All packages must include metadata proving the publisher owns them. This prevents impersonation and ensures authenticity (see more reasoning in [#96](https://github.com/modelcontextprotocol/registry/issues/96)).
 
 For detailed verification requirements for each registry type, see the [publishing guide](../../guides/publishing/publish-server.md).
-
-## Remote Server URL Match
-
-Remote servers must use URLs that match the publisher's domain from their namespace. For example, `com.example/server` can only use remote URLs on `example.com` or its subdomains.
 
 ## Restricted Registry Base URLs
 


jonathanhefner:
CI fails with:

```
Validating docs/modelcontextprotocol-io/remote-servers.mdx: found 6 examples

  Example 4 (line 75):
    Validating against server.schema.json: ✅
    Validating with Go Validator: ❌
      Error: remote URL https://email.example.com/mcp does not match namespace io.github.username/email-integration-mcp: remote URL host email.example.com does not match publisher domain username.github.io

  Example 6 (line 123):
    Validating against server.schema.json: ✅
    Validating with Go Validator: ❌
      Error: remote URL https://email.example.com/mcp does not match namespace io.github.username/email-integration-mcp: remote URL host email.example.com does not match publisher domain username.github.io
```

My understanding is that there are no restrictions for the remote URL of a `io.github.*` namespaced server:

https://github.com/modelcontextprotocol/registry/blob/18cde88a7e1c7916f1433c1f131a8de2fec8fe65/docs/guides/publishing/publish-server.md?plain=1#L422

The purpose of the above examples is to demonstrate that.

Is CI incorrect, or do we not currently support that usage?

jonathanhefner:
> Is CI incorrect, or do we not currently support that usage?

Related issues: https://github.com/modelcontextprotocol/registry/issues/494 and https://github.com/modelcontextprotocol/registry/issues/739

jonathanhefner:
Since it looks like #769 is on track to address #494 / #739, I removed the "URL Restrictions" section of the docs.  The remaining CI failure will be fixed once #769 is merged.

jonathanhefner:
#769 has been merged, I have rebased on `main`, and all tests now pass! :tada:

docs/modelcontextprotocol-io/authentication.mdx
```diff
@@ -0,0 +1,290 @@
+---
+title: How to Authenticate an MCP Server for Publishing
```
domdomegg: I think this is more 'How to authenticate yourself when publishing an MCP server' or 'Logging into the registry' or something like this.

Also might be useful to clarify this applies only to the official MCP registry, and that subregistries should use the auth spec. We're looking at migrating the official registry onto the (new) auth spec (see #751).

docs/modelcontextprotocol-io/todo/faq.mdx
```diff
@@ -0,0 +1,75 @@
+# Frequently Asked Questions
```
domdomegg: Should this live somewhere else?

docs/modelcontextprotocol-io/authentication.mdx
jonathanhefner: Updated. :+1:

docs/modelcontextprotocol-io/todo/faq.mdx
jonathanhefner: Yes, at one point, I was hoping to distribute this information to other docs, but there isn't an obvious place for everything, so I think having an FAQ (especially for _frequent_ questions) is fine.  Moved it out of `todo/`.

jonathanhefner:
> Generally looks good. Possibly we should just move all the docs into .mdx format and onto the website, to avoid having this weird split where half the docs are in different places.

I think some docs should continue to live in this repository.  For example, all of the docs in `docs/administration/`, `docs/contributing/`, and `docs/design/`.

I'm open to porting other docs over to `modelcontextprotocol.io`, but I want to assess needs first, because I'm not sure how everything should be organized.  (And there will also likely be more changes pending [SEP-1649](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1649).)

PR #733, opened 2025-10-30
Use the official govulncheck-action

rdimitrov:
## Motivation and Context

Better to use the action with proper versioning instead of latest.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/modelcontextprotocol/registry
 
-go 1.24.6
+go 1.24.9
 
 require (
 	cloud.google.com/go/kms v1.23.0


rdimitrov:
The failing check is expected, see #732

rdimitrov:
> Am okay to merge, will defer to you. But I think I actually prefer if we can keep the workflow primarily bash commands as easier to debug e.g. locally than custom actions.

I like that too but when running govulncheck locally it is a bit weird unfortunately. The reason is it uses the Go version available on your host (not the one specified in the go.mod file) so if people are not aware of this it may give false positive results since almost always the host version of Go is equal or greater than the one in the go.mod file.

PR #756, opened 2025-11-07
Add registry authorization specification

### LINKED ISSUE #751: Introduce auth spec into Registry
Requirements from @toby:
- Goal is to allow private entries into an MCP registry
- Users who auth into an MCP registry should be able to see different available MCP servers based on what permissions they have
- Should be equal to (or at least similar) to how auth in the MCP spec works

Toby says Harald from VS Code and others can provide feedback, etc, to help nail down requirements here.

This would _not_ be functionality we expose in the public registry - it'd just be a spec that downstream sub-registries can (optionally) adopt.

domdomegg:
Add a standardized OAuth 2.1-based authentication specification for MCP sub-registries, aligned with the MCP Authorization Specification.

This enables:
- Private registry entries with access control
- Permission-based server visibility
- Client code reuse (clients can use their existing MCP auth implementation)

The spec is intended for downstream sub-registries. The official registry remains public for reading.

Fixes #751

### DIFF
--- a/docs/reference/api/registry-authorization.md
+++ b/docs/reference/api/registry-authorization.md
@@ -1,159 +1,26 @@
-# Registry Authorization Specification
+# Registry Authorization
 
-A standardized authentication and authorization mechanism for MCP registries, based on the [MCP Authorization Specification](https://modelcontextprotocol.io/specification/draft/basic/authorization).
+MCP registries wishing to implement authentication SHOULD follow the [MCP Authorization Specification](https://modelcontextprotocol.io/specification/draft/basic/authorization).
 
-**Scope:** This specification is intended for downstream sub-registries (e.g., private company registries). The official modelcontextprotocol.io registry remains public for reading, and uses a different auth system for publishing for legacy reasons (this may change in future).
+## How it works
 
-## Architecture
+The registry acts as an OAuth 2.1 Resource Server, identical to how MCP servers work. This means:
 
-Sub-registries implementing this specification act as OAuth 2.1 Resource Servers, identical to how MCP servers work:
+- **MCP clients** can reuse their existing MCP authorization implementation without any changes
+- **Registries** validate access tokens the same way MCP servers do
+- **Users** get a consistent login experience across MCP servers and registries
 
-- **Registry** = OAuth 2.1 Resource Server (validates access tokens)
-- **MCP Client** = OAuth 2.1 Client (requests registry resources with tokens)
-- **Authorization Server** = Issues tokens and manages authentication
+## Registry-Specific Scopes
 
-This allows clients to reuse their existing MCP authorization implementation without modification.
+Registries MAY use these scopes:
 
-## Client Requirements
+- `mcp-registry:read` - List and read server metadata
+- `mcp-registry:write` - Publish, update, and delete servers
 
-Clients MUST follow the [MCP Authorization Specification](https://modelcontextprotocol.io/specification/draft/basic/authorization) when authenticating to registries. The flow is identical to MCP server authentication.
+These are recommendations - registries may use any set of scopes they deem sensible.
 
-### Discovery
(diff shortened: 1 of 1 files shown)


domdomegg:
@claude please review

rdimitrov:
Looks good to me 🚀 I think we are all in favour of keeping it as close to the existing MCP spec as possible but let's wait a few more days so others can chime in if needed 👍

rdimitrov (dismissed):

toby (dismissed):

domdomegg:
Just wanting to also get the opinions of a few security professionals in Anthropic as well, will merge pending their review.

docs/reference/api/registry-authorization.md
```diff
@@ -0,0 +1,159 @@
...
+  "resource": "https://registry.example.com",
+  "authorization_servers": ["https://auth.example.com"],
+  "scopes_supported": ["registry:read", "registry:write"],
+  "bearer_methods_supported": ["header"]
+}
+```
+
+### Authorization Server Discovery
+
+Clients MUST discover the authorization server using standard OAuth 2.0 / OpenID Connect well-known endpoints, following the same priority order specified in the MCP Authorization Specification.
```
henrythe9th: OAuth 2.0+ or 2.1 as it's 2.1 everywhere else

docs/reference/api/registry-authorization.md
```diff
@@ -0,0 +1,159 @@
...
+
+**For sub-registries:** Industry-standard OAuth 2.1 with flexible permissions and proven security model.
+
+**For users:** Consistent login experience across MCP servers and registries.
+
+## Comparison with Official Registry Authentication
+
+The official modelcontextprotocol.io registry uses a custom JWT-based authentication system for publishing servers. Reading servers remains public and does not require authentication.
+
+We may in future look at moving the official registry to this standardized OAuth 2.1-based system for consistency.
```
henrythe9th: is there a benefit (or con) for not keeping the official registry in sync with this spec?

docs/reference/api/registry-authorization.md
```diff
@@ -0,0 +1,159 @@
...
+Registries MAY implement permission-based filtering of server lists. When a client provides an access token, registries SHOULD filter the response based on the token's claims, excluding servers the user does not have permission to access.
+
+Example approach:
+
+```json
+{
+  "name": "io.github.acme/internal-tools",
+  "title": "ACME Internal Tools",
+  "visibility": "private",
+  "allowed_scopes": ["registry:read:acme-internal"]
```
cjellick: What's the function of this field? Does this need to map to a scope in the token?

domdomegg: Have removed this - not actually necessary.

docs/reference/api/registry-authorization.md
domdomegg: Fixed! And simplified this in general to avoid discrepancies like these.

docs/reference/api/registry-authorization.md
domdomegg: Main benefit is that people only need to implement one auth spec consistently across servers and registries. Most MCP clients will already have all the code for handling the MCP auth spec, so if we reuse that they get registry auth for free.

There's no particular con from keeping them synced, except for potentially that the MCP auth spec is a little more complex than say a username/password or fixed API key (although roughly equivalent things can be achieved with the client credentials grant OAuth flow).

domdomegg:
Okay, after security review basically everyone seems happy with this, except with the recommendation to slim this spec down to just pointing at the main MCP spec. This avoids duplication and reduces the change of inconsistencies between the specs.

Have enabled auto-merge so this goes in once someone GitHub stamps :)

marcpradas-IAG:
thank you guys, haven't seen it so far but waiting for it half a year at least

copilot still has it in their milestones for april XD (slow people)

PR #757, opened 2025-11-07
Add simple web UI for browsing the registry

### LINKED ISSUE #99: UI for the official registry
Primary purpose: sunset the [modelcontextprotocol/servers README](https://github.com/modelcontextprotocol/servers/blob/main/README.md).

domdomegg:
## Summary

Adds a clean, minimal web interface at `/` for browsing MCP servers in the registry. Built as a single self-contained HTML file embedded in the Go binary.

## Features

- Responsive grid layout (1/2/3 columns based on screen size)
- Search servers by name
- Filter to show only latest versions (default on)
- "Recently Updated" section for servers updated in the last hour
- Expandable cards with full server metadata and JSON viewer
- Debug modal to switch between production/staging/local/custom API endpoints
- Full URL parameter support and browser history navigation

## Implementation

- Single HTML file (~400 lines) with Tailwind CSS via CDN
- Zero build dependencies
- Embedded using `//go:embed` 
- Served directly from router at root path
- Uses stable v0.1 API endpoints

Fixes #99

---

AI disclosure: this is 99% vibed with Claude. It's intended as a v0 to get something up and working, and then maybe at a later date we can replace it with something more real.

### DIFF
--- a/internal/api/router/router.go
+++ b/internal/api/router/router.go
@@ -189,7 +189,10 @@ func NewHumaAPI(cfg *config.Config, registry service.RegistryService, mux *http.
 		if r.URL.Path == "/" {
 			// Serve UI at root
 			w.Header().Set("Content-Type", "text/html; charset=utf-8")
-			w.Write([]byte(v0.GetUIHTML()))
+			_, err := w.Write([]byte(v0.GetUIHTML()))
+			if err != nil {
+				http.Error(w, "Failed to write response", http.StatusInternalServerError)
+			}
 			return
 		}
 


domdomegg:
lil video demo:

https://github.com/user-attachments/assets/de7024fc-bb51-4f45-a09f-9dd14ab2f56d

rdimitrov (approved):
Nice! 🚀 

nit: Maybe add a follow up that updates the docs/readme so we can advertise there's now a UI available 👍

domdomegg:
> nit: Maybe add a follow up that updates the docs/readme so we can advertise there's now a UI available 👍

Sounds good - will add this once we deploy to production.

PR #771, opened 2025-11-15
Add flags for publish subcommand

### LINKED ISSUE #767: publish subcommand's `--file` flag not working
**Describe the bug**

**To Reproduce**
1. place the server.json in `.registry/server.json`.
2. Run `mcp-publisher publish --file .registry/server.json --dry-run`
3. Error `Error: server.json not found. Run 'mcp-publisher init' to create one`

**Expected behavior**

Expected dry run to run on the `.registry/server.json` file.

**Logs**
If applicable, add logs to help explain your problem.

Here's the error message:
`Error: server.json not found. Run 'mcp-publisher init' to create one`

**Additional context**

Yuan325:
Add flags for publish subcommand.

## Motivation and Context

We'll need to use the `--file` flag since we store our server.json in a folder. 

The options that was noted in the docs [here](https://github.com/modelcontextprotocol/registry/blob/d9c2e506a98f7ab03fe602c3dd6c932d1b1f963b/docs/reference/cli/commands.md#mcp-publisher-publish) was not implemented. This PR added the 3 flags that was stated.

## How Has This Been Tested?

Tested locally

## Breaking Changes

No breaking changes. The existing way of defining file `mcp-publisher publish server.json` will still work.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Fixes #767

### DIFF
--- a/cmd/publisher/commands/publish.go
+++ b/cmd/publisher/commands/publish.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"net/http"
@@ -17,14 +18,21 @@ import (
 )
 
 func PublishCommand(args []string) error {
+	publishFlags := flag.NewFlagSet("publish", flag.ExitOnError)
+	serverFile := publishFlags.String("file", "server.json", "Path to server.json (default: ./server.json)")
+	registryURLFlag := publishFlags.String("registry", "", "Registry URL override")
+	dryRun := publishFlags.Bool("dry-run", false, "Validate without publishing")
+	if err := publishFlags.Parse(args); err != nil {
+		return fmt.Errorf("failed to parse flags: %w", err)
+	}
+
 	// Check for server.json file
-	serverFile := "server.json"
 	if len(args) > 0 && !strings.HasPrefix(args[0], "-") {
-		serverFile = args[0]
+		serverFile = &args[0]
 	}
 
 	// Read server.json
-	serverData, err := os.ReadFile(serverFile)
+	serverData, err := os.ReadFile(*serverFile)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return fmt.Errorf("server.json not found. Run 'mcp-publisher init' to create one")
@@ -71,8 +79,15 @@ Migrate to the current schema format for new servers.
 
 	token := tokenInfo["token"]
 	registryURL := tokenInfo["registry"]
+	if *registryURLFlag != "" {
+		registryURL = *registryURLFlag
+	}
 	if registryURL == "" {
 		registryURL = DefaultRegistryURL
+	}
+
+	if *dryRun {
+		return nil
 	}
 
 	// Publish to registry


Yuan325:
@domdomegg @rdimitrov I am unable to add reviewers to this PR (and wasn't sure who should I request for reviews). Please take a look, thank you!! :)

cmd/publisher/commands/publish.go
```diff
@@ -71,10 +79,17 @@ Migrate to the current schema format for new servers.
 
 	token := tokenInfo["token"]
 	registryURL := tokenInfo["registry"]
```
domdomegg, line 81: I think the registryUrl probably won't be helpful as your token determines which registry you have authed to

cmd/publisher/commands/publish.go
```diff
@@ -17,14 +18,21 @@ import (
 )
 
 func PublishCommand(args []string) error {
+	publishFlags := flag.NewFlagSet("publish", flag.ExitOnError)
+	serverFile := publishFlags.String("file", "server.json", "Path to server.json (default: ./server.json)")
```
domdomegg, line 22: My general philosophy is to avoid configuration options as much as possible - each additional flag is another lever of complexity to maintain, debug, document, etc. And they compose exponentially so you increase the surface area for problems a lot.

I can see this being helpful, although I'd ask first why:
- the default isn't working, e.g. you can't name it server.json. Note that you could put this in a subfolder and run publish from there
- failing that, whether it couldn't compose already with other tools (e.g. move the server.json into the right place or a tmp folder, then run publish, then clean up).

domdomegg:
Also I appreciate these flags are bizarrely in the docs, which seems like a mistake for something not implemented, so we do need to fix something (either the docs or the implementation). I'd lean towards removing the options from the docs.

cc @rdimitrov I wonder what you think here?

cmd/publisher/commands/publish.go
Yuan325: @domdomegg  The `--file` flag is a requirement for our planned expansion to support multiple servers within the registry. These servers will consist of (1) multiple prebuilt tools for specific database connections (typically via `stdio`) -- each being one server and (2) custom tools that users can build using their own configuration files.

Since the startup args differ between these server types, we will be using different `server.json` for each server. Currently, we only have the custom tools server published, and planning to expand it next.

Yuan325: The second method, although not ideal, will probably work: copying the file to server.json -> run publish -> clean up -> do the same for the remaining file.

rdimitrov:
> Also I appreciate these flags are bizarrely in the docs, which seems like a mistake for something not implemented, so we do need to fix something (either the docs or the implementation). I'd lean towards removing the options from the docs.
> 
> cc @rdimitrov I wonder what you think here?

@domdomegg - Yeah, I think I'm also a bit unsure about this and I think I'm more in favour of updating the docs to reflect our current state, at least for now. 

Note that I'm not dismissing the issue, but my main reason behind this is it's a matter of time we move the CLI to a proper framework, like cobra, which would give us not only a very good foundation but most probably address most of these issues for free (we've chatted with @toby about this at some point). I'm hoping to help contribute that in the next weeks unless someone beats me to it of course 👍

Yuan325:
@domdomegg @rdimitrov Got it, thank you! I'll close this PR as not fixed

Issue #389, opened 2025-09-09
Add 'Server List Filtering' into OpenAPI spec


formulahendry:
In https://github.com/modelcontextprotocol/registry/pull/374, it supports `updated_since`, `search`, and `version` query parameters.
Please also help make them include them in the official OpenAPI spec, so that other subregistries could align and follow them.

Especially the `search` and `version` query parameters are very helpful to client maintainers which consume the regsitry APIs:

- `search` is helpful for query and pagination.
- `version=latest` is helpful for client to avoid show dupilcated MCP server with multiple versions.

domdomegg:
Yeah, I think it's likely that we add something like this to the official spec in time.

Some discussion here about whether we should do this, where I suggested in the first instance that we shouldn't so that we can experiment a bit with the filters: https://github.com/modelcontextprotocol/registry/issues/291#issuecomment-3208819441

tadasant:
Yeah I'd like to build confidence that these are the right shapes before we bake them into the standard.

To give an extreme counter-example: [WoT's Discovery spec uses JSONPath/XPath/SPARQL](https://www.w3.org/TR/wot-discovery/#jsonpath-semantic) for exposing search capabilities. I'd like to see someone think through the argument for why we wouldn't eventually want something sophisticated like that, especially when pushing this federated, opinionated subregistry model.

axel7083:
The `version` query parameter only support `latest`, I don't see much use cases where you would search by version other than latest, so maybe `latest` should be boolean optional query parameter instead?

tadasant:
> The version query parameter only support latest, I don't see much use cases where you would search by version other than latest, so maybe latest should be boolean optional query parameter instead?

The query parameter should support any string there (let us know if you see otherwise).

The only way to uniquely identify a server.json file is a combination of `name` and `version`, so `version` is very critical (it's conceptually part of a composite primary key), and I don't think there is an opportunity to simplify there.

rdimitrov:
Closing this out. All three filters requested here (`search`, `updated_since`, `version`) are documented in the OpenAPI spec at `docs/reference/api/openapi.yaml` (and rendered at `registry.modelcontextprotocol.io/docs`). The broader discussion in-thread about more sophisticated filter shapes is a separate design question; worth a fresh, narrower issue if that direction resurfaces.

Issue #24, opened 2025-05-09
MCP Endpoints


patwhite:
**Is your feature request related to a problem? Please describe.**
The registry should really expose MCP endpoints, both as a general principal that MCP first makes a lot of sense, but almost more specifically because hosts / models will definitely want to interact with this.

**Describe the solution you'd like**
MCP Endpoints - this needs to be horizontally scalable so I'll selfishly suggest https://github.com/Traego/scaled-mcp

**Describe alternatives you've considered**
REST is fine, but I think MCP should be default for new projects within this overall spec.

VishalYadavCF:
I didn't understand "selfishly" until i opened the repository of scaled-mcp

tadasant:
For the use cases and scope for which we are designing, REST makes more sense. I agree that at some point, having some sort of spec for how an registry implementation should expose itself via MCP could be helpful, but by default I'd say there's nothing stopping someone from just building an MCP server bridge and using that. We can bake it into the core architecture if there is meaningful need to do so down the line.

thoorp:
How about extending https://modelcontextprotocol.io/mcp ?

clareliguori:
+1 for supporting the HTTP MCP protocol. I would very much like my AI coding assistant to be able to suggest relevant MCP servers when I'm building an agent, by connecting to the MCP registry APIs as MCP tools.

tadasant:
Having further developed thoughts on this: I actually now don't think we should implement this. I'm open to being convinced otherwise, but here is my reasoning --

The target end-consumers of the MCP Registry are [subregistries](https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/consuming/use-rest-api.md). It is _not_ MCP Clients directly, nor agents (wielding MCP clients). The MCP Registry is a technical infrastructure building block that centralizes publication of MCP server versions and metadata for subregistries to ingest and mirror, but is not architected (or resourced) to solve "discovery" for potentially millions of end-users and all kinds of verticals and user personas. That is why we fan out the final discovery layer to the notion of subregistries. 

Exposing MCP endpoints would be contrary to this design goal - it would encourage direct usage by MCP end-users of the MCP Registry. Maintenance practicalities aside, the UX of agentically trying to parse "a centralized registry that has all millions of MCP server versions ever built"  is not going to go well - so I'm not convinced this is something that would be practical to use, even if we committed to maintaining it.

patwhite:
[...]

PR #777, opened 2025-11-17
Add registry extensions specification

### LINKED ISSUE #776: Extensions spec
We receive lots of suggestions to improve the MCP registry. Like:
- free-text / "smart" search ( https://github.com/modelcontextprotocol/registry/issues/389, https://github.com/modelcontextprotocol/registry/issues/135 )
- MCP server ( https://github.com/modelcontextprotocol/registry/issues/24 )
- skills ( https://github.com/modelcontextprotocol/registry/issues/721 )
- client MCP documentation
- clients registry

This puts us in the unfortunate position of having to choose between:
- Implementing something that might not get broad usage, and then it being very hard to unship because to do so would cause a breaking registry API spec change. Then we're stuck with this maintenance burden forever.
- Not implementing things that might be useful. Then we lose out on registry being better.

Develop some convention around common extensions, ideally so that:
- there are conventions around common extensions, so that people may start to adopt them / AI agents know how to use them
- there are conventions around how to build extensions, so that things are generally standardised ish & easy to adopt
- anyone can build custom extensions if they like, and be unlikely to clash with others

Ideally we also try to build things out such that it's easy to compose things together.

domdomegg:
Introduces a standardized way for registries to provide experimental or community-driven features without committing them to the core API specification.

## Summary

- Extensions live under `/v0/x/<namespace>/<extension>` URL structure
- Simple conventions: REST, JSON responses, cursor pagination, optional auth
- Composable design building on core APIs where possible

This enables experimentation with features like search (#389) and MCP server integration (#24) while keeping the core API minimal and stable.

Fixes #776

### DIFF
--- a/docs/reference/api/extensions.md
+++ b/docs/reference/api/extensions.md
@@ -1,0 +1,70 @@
+# Registry Extensions Specification
+
+A standardized way for registries to provide experimental or community-driven features without committing them to the core API specification.
+
+## Motivation
+
+[The core generic registry API](./generic-registry-api.md) intentionally stays minimal to ensure stability and broad adoption. Extensions provide a path for:
+
+- **Experimentation**: Try new features without core API changes
+- **Community innovation**: Anyone can implement custom extensions
+- **Gradual adoption**: Popular extensions may inform future core API features
+- **Avoiding breaking changes**: Failed experiments can be deprecated without API versioning churn
+
+## URL Structure
+
+Extensions live under the `/v0/x/` prefix:
+
+```
+/v0/x/<namespace>/<extension>[/<path>]
+```
+
+**Components:**
+- `<namespace>`: Reverse domain ownership (e.g., `com.example`, `io.github.username`)
+- `<extension>`: Extension name (lowercase, hyphens for word separation)
+- `<path>`: Extension-specific path structure (optional)
+
+**Examples:**
+```
+/v0/x/com.example/search?q=database
+/v0/x/com.example/stats
+/v0/x/io.github.username/custom-feature
+```
+
+## Conventions
+
+Where possible:
+- Follow standard REST conventions, return simple JSON responses, and avoid special headers
+- For list endpoints, use cursor-based pagination matching the core API
+- Extensions requiring authentication **SHOULD** follow the [Registry Authorization Specification](./registry-authorization.md)
+- Build open-source implementations in a composable way on top of the core APIs (e.g. as opposed to via custom database integration)
+
+## Implementation Requirements
+
+Registries implementing extensions **SHOULD** namespace extensions properly to avoid conflicts.
+
+Clients consuming extensions **MUST** gracefully handle missing extensions.
+
+## Example
+
+A simple server stats extension:
+
+```bash
+GET /v0/x/com.example/stats
+```
+
+```json
+{
+  "totalServers": 1234,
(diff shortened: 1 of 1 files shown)


domdomegg:
Areas that I think people might have the most issue with:

1. Namespacing. Maybe this should be a MUST to prevent conflicts? Maybe namespacing is too heavy for simple extensions. It does make the URLs ugly but my guess is humans generally won't see them so maybe it's fine?

2. URL structure. Why `/x/`? Why not `/extensions/` or `/ext/`. Or maybe even `/experimental/`? This is mainly to have shorter URLs (but appreciate this is at odds a bit with the comment above about humans not really reading this).

3. No extension discovery mechanism. It's in "future considerations", but haven't actually specced out what this looks like. I _think_ it's okay to go ahead as-is and if people need this discovery mechanism we can add it later. Perhaps an array of strings at `/v0/x`? Or maybe array of objects with name key, and then people can put other things in later if we need to extend.

docs/reference/api/extensions.md
```diff
@@ -0,0 +1,70 @@
...
+/v0/x/com.example/stats
+/v0/x/io.github.username/custom-feature
+```
+
+## Conventions
+
+Where possible:
+- Follow standard REST conventions, return simple JSON responses, and avoid special headers
+- For list endpoints, use cursor-based pagination matching the core API
+- Extensions requiring authentication **SHOULD** follow the [Registry Authorization Specification](./registry-authorization.md)
```
domdomegg, line 39: See https://github.com/modelcontextprotocol/registry/pull/756

docs/reference/api/extensions.md
```diff
@@ -0,0 +1,70 @@
...
+{
+  "totalServers": 1234,
+  "totalVersions": 5678,
+  "recentPublishes": 42
+}
+```
+
+## Future Considerations
+
+- **Extension discovery**: A potential `/v0/x` endpoint to list available extensions
```
rdimitrov, line 66: Do you think we should make this mandatory?

rdimitrov (approved):
Looks great! I've added one suggestion but we can do a follow up with it if you agree it makes sense 👍

blkt:
I'm a bit concerned by the fact that version number for the extension would be the same as the official one. This means it cannot be bumped when the extension changes.

I'm not really keen on having two version numbers in one path, isn't it better to suggest `/x/<namespace>/<extension>/<version>` instead?

domdomegg:
> I'm a bit concerned by the fact that version number for the extension would be the same as the official one. This means it cannot be bumped when the extension changes.
> 
> I'm not really keen on having two version numbers in one path, isn't it better to suggest `/x/<namespace>/<extension>/<version>` instead?

Maybe this is a bit silly, but I was imagining that if extensions want versioning they should append it on.

So I see v0 as the version for the overall registry spec, and how you do a v0 extension is put it under /v0/x/...

But if you want your extension to be versioned, you have `/v0/x/<namespace>/<extension>/<version>` (appreciate it's getting a bit clunky though)

docs/reference/api/extensions.md
domdomegg: Yeah - I think either:
- an array of strings that are extension paths
- an array of objects with `path` entries. same as above, but gives us room for extension in future

Thoughts?

PR #782, opened 2025-11-18
Add DELETE method to the OpenAPI specification

rdimitrov:
## Motivation and Context

The following adds a DELETE method to the reference API specification of the registry. 

The endpoint is consistent with the rest of the spec and provides a standardised way for registry implementations to support deletion if they choose to.

**Details:**
* Endpoint: DELETE /v0/servers/{serverName}/versions/{version}
* Uses tags: [publish] (same as POST /v0/publish for write operations)
* Requires security: bearerAuth authentication
* Returns ServerResponse on success (200) - matching POST pattern
* Same error responses: 401 (Unauthorized), 403 (Forbidden), 500 (Internal error)
* Added 404 (Not Found) for when server/version doesn't exist
* Parameters match the GET endpoint format with URL-encoding examples
* Marked as "(Optional)" in summary
* Explicit note that it's not implemented by the official registry
* States it's included for standardization purposes

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/docs/reference/api/openapi.yaml
+++ b/docs/reference/api/openapi.yaml
@@ -322,6 +322,16 @@ paths:
                   error:
                     type: string
                     example: "Failed to publish server"
+        '501':
+          description: Not Implemented - Registry does not support publishing
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: "Publishing is not supported by this registry"
 components:
   securitySchemes:
     bearerAuth:


docs/reference/api/openapi.yaml
```diff
@@ -135,6 +135,79 @@ paths:
...
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  error:
+                    type: string
+                    example: "Failed to delete server version"
```
blkt, line 210: **nit:** it would be nice to return a **501 Not Implemented** in case the server does not support server version deletion.

rdimitrov: Thanks! I also added it to `publish` since we've missed it there too 👍

rdimitrov:
> Looks good - can we also add a line in [the docs](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/api/generic-registry-api.md)?

Addressed 🚀

PR #787, opened 2025-11-19
Implement support for searching by type

### LINKED ISSUE #781: Discover remote servers (filter by type)
Registry APIs are used by AI Agents to discover existing servers to be connected.
As of today, the search API provide pagination and filter by name, but there's no way to select only the servers with a remote configuration. Before finding the desired (remote) server, the client may need to fetch multiple pages for nothing.

Adding a type filter to the query to further qualify the output:
`/v0/servers?type=remote&search=github`

This should return all the servers with at least one remote configuration and matching the given filters.
Similarily we could have `type=package` to list all servers with at least one package configuration. This is probably a less real-world use case.

rdimitrov:
## Motivation and Context

The following PR adds a query parameter for searching by the distribution type (remote, npm, nuget, etc). This is so registry clients supporting nuget can query just for those (less load on the registry server + less load on the client to go through all results).

Note that this is not a breaking change. 

## How Has This Been Tested?

Locally + unit tests

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Fixes: #781


docs/reference/api/openapi.yaml
```diff
@@ -61,6 +61,18 @@ paths:
           schema:
             type: string
             example: "1.2.3"
+        - name: type
+          in: query
+          description: |
+            Filter by distribution type. Allows discovering servers based on how they are made available to users.
+
+            Example values include: 'remote', 'npm', 'pypi', 'oci', 'nuget', 'mcpb'
```
domdomegg, line 69: Would it be worth changing `remote` to `sse` and `streamable-http`? I know that at Anthropic we maybe have plans to deprecate `sse` so being able to filter by the actual remote type might be useful - and it also aligns with filtering by the package type.

domdomegg:
Meta: I might want one extra reviewer to check this as well - adding an endpoint is something that is hard to take back later so better to double check. Maybe we can discuss later today in maintainers meeting?

docs/reference/api/openapi.yaml
rdimitrov: I think that's reasonable 👍 Would it be a problem if we continue to have remote though as an option that aggregates all remote servers?

rdimitrov:
> Meta: I might want one extra reviewer to check this as well - adding an endpoint is something that is hard to take back later so better to double check. Maybe we can discuss later today in maintainers meeting?

Just to chime in that this adds a property to an existing endpoint and not a new one 👍

docs/reference/api/openapi.yaml
domdomegg: maybe a slight preference for making this composable in some way, e.g. you can specify the type multiple times or you can comma-separate types or something? then we don't need remote, you can just specify `streamable-http,sse`

tadasant (changes requested):
In concept, I am supportive of adding "filter" functionality to the registry spec. It's clear that for ETL processes pulling down entries from the Registry or a sub-registry, they will often just want to pull down a subset based on downstream support of features.

However, I'm not a fan of adding a `type` query parameter here. If we do this for `type`, there's an explosion of other filters we would create a precedent for adding. For example `uses_template_variables`, `has_git_repository`, `uses_semantic_versioning`, etc etc. It wouldn't be obvious to me why we would support `type` and not those others.

To land this, I think we need to make the feature more generic. Rather then encoding every possible "filter" into the spec, the caller should be able to express a filter that the server then applies on its response.

One option is to take a page from [WoT Discovery and use JSONPath](https://www.w3.org/TR/wot-discovery/#jsonpath-semantic). The queries it would create are ugly, but it'd be a nice standardized way to allow consumers to filter whatever fields they care about. My main concern with this approach would be whether the server side implementation is very nonperformant, and whether we would need to lock down what is allowed. A solution for that: the generic schema could support JSONPath, but the public registry API could just support very specific predefined JSONPath queries (this `type` one being one of them).

I think I'd be good with a JSONPath solution, though also am curious for more/alternative ideas that would fulfill the "flexible for consumers" kind of requirement here.

@claude do you have any ideas? Propose a few alternatives to JSONPath, and share your opinion on whether JSONPath is a good solution here for a generic filtering capability.

tadasant:
After poking around some more, I can't find any good precedent for introducing something like what I'm suggesting into a major piece of ecosystem infra. In fact it looks like NuGet used to use OData and migrated away in their v3, and no other reasonable precedent exists.

I actually wonder if `4. Predefined Named Filters` from Claude's response might be the best way to go. No performance concerns, still flexible and every sub-registry can have their own "predefined" sets that they care about. It's actually not that different from your original proposal conceptually, where you are introducing a new notion of `type` that could mean `remote`, could mean local `npm`, etc. We just tweak by:
- Stepping up the abstraction layer (`filter` instead of `type` as the key)
- Only define the key and not the values in the sub-registry spec (we don't need to make decisions on whether `'remote', 'npm', 'pypi', 'oci', 'nuget', 'mcpb'` etc are the right values - any sub-registry can come up with any filters they please; maybe the official registry can offer some popular ones that folks might standardize around, but we could tweak them without breaking API changes over time)

What do you think @rdimitrov? Would that still fulfill your needs with the benefit of avoiding the explosion of modeling for us?

@claude what's your deeper take on this path for the MCP Registry?

tadasant:
I don't agree with Claude's suggestion to split into 2 functionalities (would rather we just focus on `filter` as a key), FWIW.

jonathanhefner:
Rather than `type`, perhaps `category`?  Then servers could automatically be placed in zero or more categories (e.g., `["remote", "sse", "streamable-http"]`).  And we could add arbitrary categories in the future.

Though that does introduce the question of whether `?category=` with multiple categories should be a union versus an intersection.

tadasant:
> Rather than `type`, perhaps `category`? Then servers could automatically be placed in zero or more categories (e.g., `["remote", "sse", "streamable-http"]`). And we could add arbitrary categories in the future.
> 
> Though that does introduce the question of whether `?category=` with multiple categories should be a union versus an intersection.

@jonathanhefner did you see my suggestion of `filter` instead of `type`? Curious what you think of `filter`.

I don't love category personally, because category is typically a mutually exclusive designator (items can only be in one category). `tags` would be more typically like what you are proposing (item can be in multiple), but indeed the comma-separation being union or intersection is a very real problem.

I do think making the key `filter` and allowing custom values solves all these problems (a sub-registry could implement `remote-and-npm` as a filter just as easily as `remote-or-npm`). Maybe eventually (or even out of the gate) we make a list of "recommended `filter` values to implement".

jonathanhefner:
> did you see my suggestion of `filter` instead of `type`? Curious what you think of `filter`.

> I don't love category personally, because category is typically a mutually exclusive designator (items can only be in one category).

I was interpreting `filter` as mutually exclusive. :laughing:  But also, I was interpreting `filter` as manually defined rather than automatically defined.  For example, if we add support for a new package registry, would we automatically have a new filter for that registry?  For `category` or `tag`, I would assume "yes".  If the answer for `filter` is also "yes", then I think that sounds good! :smiley:

tadasant:
> For example, if we add support for a new package registry, would we automatically have a new filter for that registry? For `category` or `tag`, I would assume "yes". If the answer for `filter` is also "yes", then I think that sounds good! 😃

Hmm - I think my position would be that I **don't** think we should automatically register these values (whether they be `categories`, `tags`, or `filters`) and rather either leave them completely free form, or manually provide guidance (for example, start with saying `remote`, `npm`, `pypi`, `oci`, `nuget`, and `mcpb` are the recommended values to implement; but not required for spec compliance). We could "automate" the registration if we find this list becomes hard to maintain for some reason.

But yes, I think whether its `categories`, `tags`, or `filters`; we _could_ do some automatic registration scheme all the same at some point if we want to.

rdimitrov:
Hey, sorry for being out for a bit and not replying. 

I usually like the direction of making things flexible, but I’m a bit worried about leaving values undefined in the registry spec.

Whether that's `type` or `filter` or some other parameter we add, if we do it without specifying how its values are interpreted we’re opening the door to divergence between implementations. One registry might interpret it one way, another registry in a completely different way (same goes for clients) and suddenly clients have to implement special-case logic for each registry. We can see how this adds fragmentation and is probably the opposite of what we want from a “standard” API.

In my view the API spec is effectively the contract we are creating for the ecosystem and I think it's fair to be more explicit and deterministic about how some of these things are defined and expected to work. In a way it's our responsibility to ensure clients have predictable behavior, registries have clear requirements and we keep interoperability intact.

I imagine @BobDickinson can chime in here too since it is related to the problem he's helping us solve on the server.json schema side.

tadasant:
> Hey, sorry for being out for a bit and not replying.
> 
> I usually like the direction of making things flexible, but I’m a bit worried about leaving values undefined in the registry spec.
> 
> Whether that's `type` or `filter` or some other parameter we add, if we do it without specifying how its values are interpreted we’re opening the door to divergence between implementations. One registry might interpret it one way, another registry in a completely different way (same goes for clients) and suddenly clients have to implement special-case logic for each registry. We can see how this adds fragmentation and is probably the opposite of what we want from a “standard” API.
> 
> In my view the API spec is effectively the contract we are creating for the ecosystem and I think it's fair to be more explicit and deterministic about how some of these things are defined and expected to work. In a way it's our responsibility to ensure clients have predictable behavior, registries have clear requirements and we keep interoperability intact.
> 
> I imagine @BobDickinson can chime in here too since it is related to the problem he's helping us solve on the server.json schema side.

What is your current view on what the right design is that addresses (or deliberately trades off) all the concerns outlined to-date?

I think the following works quite well:
- `filter=` is the query parameter key
- It takes a comma separated string (that implies OR logic)
- We provide example values of `sse`, `streamable`, `npm`, `pypi`, `oci`, `nuget`, and `mcpb`
- Implementation guidance is that it is compliant to add a new value if you need some combination of something different (e.g. "streamable HTTP of all server.json's of schema version 2025-10-17 or later"). If it's generalizable enough for other folks, open a PR with it as another recommended example.

I think this works because:
- It provides foundational structure (the `filter` field) that registry implementors are encouraged to rally around.
- We continue to follow MCP's general principle of empowering use cases in the wild before trying to standardize them across vendors (example values that implementors can choose to extend as they nudge forward the set of use cases they're solving for)
- We do not create a long-tail bottleneck of maintainers adding one-off filter combinations to the spec [this is the big one I feel fairly strongly that we need _some_ solution for] and block implementors on being spec-compliant on this bottleneck when they have legitimate long tail filter scenarios they want to enable

rdimitrov:
Yeah, of course, let me sanity-check my understanding of the `filter` direction as I think it touches a broader question about what we consider part of the API contract.

I also want to acknowledge that this argument applies to my `type` proposal as well (so not exactly in ones favour or the other 😃). Overall because we haven’t standardised the package types yet values like `npm` and `pypi` are only examples and aren’t meaningful from a spec/contract perspective unless they’re required (which they are not). Adding a loosely defined property like `filter` on top feels like it compounds the ambiguity (`type` at least implies using the available package types thus why I think I prefer it more)

So up in the thread when we say things like “define the key but not the values” or “recommended but not required values” I'm cautious to support it as it seems to imply:
* registries can be compliant while ignoring or redefining `filter=npm`
* clients can’t rely on consistent cross-registry behavior
* the spec introduces one more parameter without deterministic semantics

In my mind it's not spec compliant if an implementation changes while the spec stays the same and the result is broken interoperability. For example, if tomorrow we decide to rename `npm` to `node` that would break clients and probably sub-registries all over the place while still being “spec-compliant” which is usually a sign that the contract wasn’t well specified in the first place. 

I’m all in and very supportive of keeping things simple and extensible but this makes me wonder what “spec-compliant” is actually guaranteeing here? Are we comfortable with a situation where clients and registries cannot work together, yet still claim spec compliance?

If we agree on that then whether we have `type` and/or `filter` properties it's okay from my side  since we should explicitly define the allowed values (or their subset at least). In any case I don't think we should loose a lot of time discussing this and would rather prefer to make a decision so we can move forward, but it would help if we acknowledge our stance on what spec compliant means and what not because this defines what a breaking change means too.

tadasant:
> Yeah, of course, let me sanity-check my understanding of the `filter` direction as I think it touches a broader question about what we consider part of the API contract.
> 
> I also want to acknowledge that this argument applies to my `type` proposal as well (so not exactly in ones favour or the other 😃). Overall because we haven’t standardised the package types yet values like `npm` and `pypi` are only examples and aren’t meaningful from a spec/contract perspective unless they’re required (which they are not). Adding a loosely defined property like `filter` on top feels like it compounds the ambiguity (`type` at least implies using the available package types thus why I think I prefer it more)
> 
> So up in the thread when we say things like “define the key but not the values” or “recommended but not required values” I'm cautious to support it as it seems to imply:
> 
> * registries can be compliant while ignoring or redefining `filter=npm`
> * clients can’t rely on consistent cross-registry behavior
> * the spec introduces one more parameter without deterministic semantics
> 
> In my mind it's not spec compliant if an implementation changes while the spec stays the same and the result is broken interoperability. For example, if tomorrow we decide to rename `npm` to `node` that would break clients and probably sub-registries all over the place while still being “spec-compliant” which is usually a sign that the contract wasn’t well specified in the first place.
> 
> I’m all in and very supportive of keeping things simple and extensible but this makes me wonder what “spec-compliant” is actually guaranteeing here? Are we comfortable with a situation where clients and registries cannot work together, yet still claim spec compliance?
> 
> If we agree on that then whether we have `type` and/or `filter` properties it's okay from my side since we should explicitly define the allowed values (or their subset at least). In any case I don't think we should loose a lot of time discussing this and would rather prefer to make a decision so we can move forward, but it would help if we acknowledge our stance on what spec compliant means and what not because this defines what a breaking change means too.

@rdimitrov all fair points, you've probably convinced me that `filter=*` with no further definition is probably the wrong path forward, primarily for this point:

> In my mind it's not spec compliant if an implementation changes while the spec stays the same and the result is broken interoperability. For example, if tomorrow we decide to rename `npm` to `node` that would break clients and probably sub-registries all over the place while still being “spec-compliant” which is usually a sign that the contract wasn’t well specified in the first place.

How about this, modifying my above proposal:
- `filter=` is the query parameter key
- It takes a comma separated string (that implies OR logic)
- We provide ~~example~~ **reserved** values of `sse`, `streamable`, `npm`, `pypi`, `oci`, `nuget`, and `mcpb`. Sub-registries SHOULD implement these. If they choose to support these values, they MUST implement them in alignment with how our official spec treats them.
- Sub-registries MAY implement additional filter values, but MUST vendor prefix them if they do so. e.g. `com.pulsemcp%2Fsse-and-streamable`, `com.pulsemcp%2F2025-12-11-or-newer`

If we find sub-registries using patterns often, we can lift those patterns into the reserved non-prefixed values over time.

rdimitrov:
**Update:** So after our last community meeting we decided we'll go with the last proposal by @tadasant above, i.e. have a filter property that supports a reserved set of values + allow other registry implementations to extend on top. 

I'll rebase my implementation to reflect that 👍

PR #803, opened 2025-11-27
Fix misleading server.json schema URL label

tadasant:
The schema in this repo may contain unreleased changes, so labeling it with a dated version URL is misleading. Change to use a `draft` URL instead, and add documentation for the schema contribution process.

- Update $id in server.schema.json to use draft URL
- Update example $schema URL in openapi.yaml
- Add Draft (Unreleased) section to CHANGELOG.md
- Add CONTRIBUTING.md with process for making/releasing schema changes

Written by Claude Code, reviewed and tweaked by me.

### DIFF
--- a/docs/reference/api/openapi.yaml
+++ b/docs/reference/api/openapi.yaml
@@ -709,7 +709,7 @@ components:
           type: string
           format: uri
           description: JSON Schema URI for this server.json format
-          example: "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json"
+          example: "https://static.modelcontextprotocol.io/schemas/draft/server.schema.json"
         packages:
           type: array
           items:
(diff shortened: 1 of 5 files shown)


domdomegg (commented):
Do we want people to use 'draft' version? Also given `https://static.modelcontextprotocol.io/schemas/draft/server.schema.json` doesn't exist I worry this might cause problems if people end up using this etc.

tadasant:
> Do we want people to use 'draft' version? Also given `https://static.modelcontextprotocol.io/schemas/draft/server.schema.json` doesn't exist I worry this might cause problems if people end up using this etc.

I don't think we generally want people to use the draft version -- they should use one (ideally the latest one) we've finalized in `static`.

Fair point that it's confusing that it doesn't exist at that URL. Maybe would be worth pointing it more directly at the current-draft. But I think we can take that as a follow-on to this; more critical IMO to make it clear the version in this repo is not exactly the same as the latest one in `static`.

Edit: follow up Issue: https://github.com/modelcontextprotocol/registry/issues/814

Issue #743, opened 2025-10-31
`mcp-publisher init` does not strip `git+` from repository URL


jonathanhefner:
When using the `npm init` command to create a new `package.json`, the `repository.url` will be prefixed with `git+`.  For example:

```jsonc
// package.json
{
  "name": "my-package",
  "version": "1.0.0",
  "repository": {
    "type": "git",
    "url": "git+https://github.com/jonathanhefner/my-package.git"
  }
}
```

Then, when using `mcp-publisher init` to generate `server.json`, `mcp-publisher` does not strip the `git+` prefix, resulting in an invalid URL:

```jsonc
// server.json
{
  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
  "name": "io.github.jonathanhefner/my-package",
  "repository": {
    "url": "git+https://github.com/jonathanhefner/my-package",
    "source": "github"
  }
}
```

Avish34:
@tadasant  / @domdomegg Is this fixed? If not, I can take it up.

domdomegg:
Not sure it's fixed - @Avish34 happy for you to pick it up.

PR #809, opened 2025-11-29
Fix issue-743 by sanitizing the URL

Avish34:
This PR aims to fix https://github.com/modelcontextprotocol/registry/issues/743#event-20985516896

## Motivation and Context

This is needed to ensure we are sanitizing the URL before creating server.json to be inline with the validations.

## How Has This Been Tested?

Tested via UTs and also tested CLI locally to ensure the expected behavior.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/cmd/publisher/commands/init.go
+++ b/cmd/publisher/commands/init.go
@@ -213,6 +213,10 @@ func detectDescription() string {
 }
 
 func detectRepoURL() string {
+	sanitizeURL := func(url string) string {
+		return strings.TrimPrefix(url, "git+")
+	}
+
 	// Try git remote
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
@@ -233,11 +237,11 @@ func detectRepoURL() string {
 		if json.Unmarshal(data, &pkg) == nil {
 			if repo, ok := pkg["repository"].(map[string]any); ok {
 				if url, ok := repo["url"].(string); ok {
-					return strings.TrimSuffix(url, ".git")
+					return sanitizeURL(strings.TrimSuffix(url, ".git"))
 				}
 			}
 			if repo, ok := pkg["repository"].(string); ok {
-				return strings.TrimSuffix(repo, ".git")
+				return sanitizeURL(strings.TrimSuffix(repo, ".git"))
 			}
 		}
 	}


Avish34:
@tadasant  @domdomegg  Please review!!

tadasant:
@claude please review this PR

cmd/publisher/publisher.exe
tadasant, line 1: Unintentional .exe commit?

tadasant (changes requested):
I'm not sure introducing a utils pattern is necessary here. Can we just add it to the current flow in `detectRepoURL`?

cmd/publisher/publisher.exe
Avish34: Yeah, silly of me,

Avish34:
> I'm not sure introducing a utils pattern is necessary here. Can we just add it to the current flow in `detectRepoURL`?

sure!!

cmd/publisher/publisher.exe
Avish34: @tadasant  Pushed the changes, please review.

cmd/publisher/commands/init.go
```diff
@@ -212,7 +211,17 @@ func detectDescription() string {
 	return "An MCP server that provides [describe what your server does]"
 }
 
-func detectRepoURL() string {
+func detectRepoURL(packageType string) string {
```
tadasant: Is it necessary to pipe around `packageType` here? Couldn't we just "strip git+ from the prefix if it exists" regardless of the packageType and keep this change fairly minimal? Or is there some case you're imagining where we'd want git+ as a prefix?

tadasant (changes requested):

cmd/publisher/commands/init.go
Avish34: No, I was just thinking that if someday going forward a case arises for a different package manager where we need to do some changes to repoURL generated by them. In that case, the changes will be minimal a that time. I'm happy to remove the packageType params if you think it won't be much helpful right now. Happy to address the comments, let me know your thoughts.

tadasant: I'd rather we do the simpler change now and do the bigger change later if the need arises (I'm not sure it definitely will). Thank you!

Avish34: Done @tadasant , please review.

Avish34: unrelated failure in sanity, good to merge.

tadasant (approved):

Issue #738, opened 2025-10-30
Registry base URL is not the proper entry point for NuGet


joelverhagen:
**Describe the bug**

Today, the `registryBaseUrl` serves as a pointer to which registry the package is stored in. This maybe makes sense for some package registries (i.e. base URL = entry point) but for NuGet the entry point is not a base URL, it is a URL to a JSON service index.

For example, NuGet clients configure "https://api.nuget.org/v3/index.json" NOT "https://api.nuget.org". 

This line currently violates the service discovery protocol we have for NuGet:
https://github.com/modelcontextprotocol/registry/blob/18cde88a7e1c7916f1433c1f131a8de2fec8fe65/internal/validators/registries/nuget.go#L51
(technically this should read the service index JSON document and get the README URL pattern with the `ReadmeUriTemplate/6.13.0` service ID.

It is not correct to assume the `/v3/index.json` or `/v3-flatcontainer/%s/%s/readme` can be appended to a NuGet registry base URL to get a valid endpoint. This works for NuGet.org, but not other NuGet feed implementations.

**Expected behavior**

Ideally, for NuGet, the server.json property would be `registryUrl` in which case we would provide/default to https://api.nuget.org/v3/index.json. 

Generally speaking, I think `registryUrl` is a little more generic and conceptually also supports a base URL if that is idiomatic for the given package registry type.

But the property name is already chosen and locked in the schema. Also, there is a data on PROD using that property name and value.

I propose:

1. Introduce an additional property `registryUrl`
2. For NuGet.org, recommend `registryUrl` but accept `https://api.nuget.org` and internally treat that as `registryUrl = https://api.nuget.org/v3/index.json`.
3. Enhance the NuGet validator code to do proper service discovery.

I am happy to implement all of this.

PR #813, opened 2025-12-02
Fix NuGet.org base URL default, improve API integration

joelverhagen:
## Motivation and Context

This resolves https://github.com/modelcontextprotocol/registry/issues/738. There are two issues with the NuGet.org integration today:

1. Change the `registryBaseUrl` for NuGet to `https://api.nuget.org/v3/index.json`. `https://api.nuget.org` is not the entry point for our service and is not specific enough. `https://api.nuget.org/v3/index.json` is the value that users put into their tooling to connect to our public registry (it is installed by default in .NET client tooling).
2. Use the service index API to get the URL pattern for package READMEs.
   - This URL is not guaranteed to be stable. The service index URL (`registryBaseUrl`) is, however.

Additionally, I polished up the error messages so various error modes are clearer. When packages are published to NuGet.org, they take 2-5 minutes to become available due to validations we run.

Some possible states that are now handled and described more helpfully are:
- Version exists, but the package has no README
- The ID exists, but the version does not
- The ID does not exist at all

There are only 3 authors (me, another MSFT person, and a community member) explicitly setting the old value. The DB migration will fix the issue. A future publish will fail with the validation error but the error message is clear.

`registry type and base URL do not match: 'https://api.nuget.org' is not valid for registry type 'nuget'. Expected: https://api.nuget.org/v3/index.json`

## How Has This Been Tested?

- Unit tests, new and existing
- Manual test locally
- Set old data in the DB, `docker compose up`, confirm the migration updates the specific package

## Breaking Changes

Yes, but minimal impact expected, unless there is a dependency on the specific `https://api.nuget.org` value downstream. I am not aware of any such dependency, and the new value is more proper and specific.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

For more information about the NuGet API see: https://learn.microsoft.com/en-us/nuget/api/overview#service-index (I wrote most of the docs).

### DIFF
--- a/complete.md
+++ b/complete.md
@@ -658,7 +658,7 @@ Only trusted public registries are supported. Private registries and alternative
 **Supported registries:**
 - **NPM**: `https://registry.npmjs.org` only
 - **PyPI**: `https://pypi.org` only
-- **NuGet**: `https://api.nuget.org/v3/index.json` only (`https://api.nuget.org` is supported for backwards compatibility)
+- **NuGet**: `https://api.nuget.org/v3/index.json`
 - **Docker/OCI**: `https://docker.io` only
 - **MCPB**: `https://github.com` releases and `https://gitlab.com` releases only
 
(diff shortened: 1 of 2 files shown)


joelverhagen:
I have verified the change on staging.registry.modelcontextprotocol.io. Thanks for merging, @tadasant!

PR #818, opened 2025-12-04
Fix command documentation for publish

joelverhagen:
## Motivation and Context

The `--file=PATH` pattern does not work. The path is an optional, positional argument not a named option.
The `--dry-run` option is not supported.

Look at the implementation: https://github.com/modelcontextprotocol/registry/blob/d092dd6390f1c64add83303517c4d00bfc65433e/cmd/publisher/commands/publish.go#L19

## How Has This Been Tested?

Try these options yourself, they are ignored.

## Types of changes

- [x] Documentation update

## Checklist

### DIFF
--- a/docs/reference/cli/commands.md
+++ b/docs/reference/cli/commands.md
@@ -215,30 +215,26 @@ For detailed guidance on the publishing process, see the [publishing guide](../.
 
 **Usage:**
 ```bash
-mcp-publisher publish [options]
+mcp-publisher publish [PATH]
 ```
 
 **Options:**
-- `--file=PATH` - Path to server.json (default: `./server.json`)
-- `--registry=URL` - Registry URL override
-- `--dry-run` - Validate without publishing
+- `PATH` - Path to server.json (default: `./server.json`)
 
 **Process:**
 1. Validates `server.json` against schema
-2. Verifies package ownership (see [Official Registry Requirements](../server-json/official-registry-requirements.md))
-3. Checks namespace authentication
-4. Publishes to registry
+2. Publishes the `server.json` to the registry server URL specified in the login token
+3. Server: Verifies package ownership (see [Official Registry Requirements](../server-json/official-registry-requirements.md))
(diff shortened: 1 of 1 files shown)


docs/reference/cli/commands.md
```diff
@@ -215,30 +215,26 @@ For detailed guidance on the publishing process, see the [publishing guide](../.
...
-- `--registry=URL` - Registry URL override
-- `--dry-run` - Validate without publishing
+- `PATH` - Path to server.json (default: `./server.json`)
 
 **Process:**
 1. Validates `server.json` against schema
-2. Verifies package ownership (see [Official Registry Requirements](../server-json/official-registry-requirements.md))
-3. Checks namespace authentication
-4. Publishes to registry
+2. Publishes the `server.json` to the registry server URL specified in the login token
```
donnie-msft, line 226: I'm not sure this is correct. The code appears to look for the server URL inside the `server.json` file itself.
My interpretation was login would use the parameter, but publishing either looks in `server.json` or falls back to the default (production) registry URL. I assume that if login was for another URL like staging, that it'll fail to authenticate in that case.

Reference: https://github.com/modelcontextprotocol/registry/blob/d092dd6390f1c64add83303517c4d00bfc65433e/cmd/publisher/commands/publish.go#L72-L76

donnie-msft (commented):
As someone who was relying on this document when working on a release pipeline, I appreciate this update to make it correct!

docs/reference/cli/commands.md
joelverhagen: See these lines. We read the token from disk (home dir), deserialize it, and then look for the "registry" property:
https://github.com/modelcontextprotocol/registry/blob/d092dd6390f1c64add83303517c4d00bfc65433e/cmd/publisher/commands/publish.go#L58-L76

donnie-msft: ah, it's from an intermediate login file, not the server.json or similar. Thanks

PR #833, opened 2025-12-09
Add IP-based rate limiting to protect against abuse

### LINKED ISSUE #826: Implement rate limits to address mis-use causing downtime
See: https://discord.com/channels/1358869848138059966/1437868748890571045/1447619393667469353

> Maybe the limit should be something like 60 reqs / minute, 1000 / hour per IP? Could be just on /servers to start. Still pretty lenient but would get rid of these breaking spikes

tadasant:
Written by Claude Code, reviewed by me. When we get this in `main` I'll test it out against staging to make sure it's acting as intended.

## Summary
- Implements IP-based rate limiting middleware to address service disruptions caused by excessive requests (#826)
- Enforces dual rate limits: 60 requests/minute and 1000 requests/hour per IP address
- Fully configurable via environment variables (`MCP_REGISTRY_RATE_LIMIT_*`)
- Excludes health/ping/metrics endpoints from rate limiting
- Returns RFC 7807 problem+json responses with Retry-After header when rate limited

## Configuration Options
| Variable | Default | Description |
|----------|---------|-------------|
| `MCP_REGISTRY_RATE_LIMIT_ENABLED` | `true` | Enable/disable rate limiting |
| `MCP_REGISTRY_RATE_LIMIT_REQUESTS_PER_MINUTE` | `60` | Max requests per minute per IP |
| `MCP_REGISTRY_RATE_LIMIT_REQUESTS_PER_HOUR` | `1000` | Max requests per hour per IP |

## Implementation Details
- Uses token bucket algorithm via `golang.org/x/time/rate` (already a dependency)
- Properly handles `X-Forwarded-For` and `X-Real-IP` headers (since registry is deployed behind NGINX ingress with `use-forwarded-headers=true`)
- All IPs are validated using `net.ParseIP()` to handle malformed input
- Memory protection via MaxVisitors limit (100k) with LRU eviction
- Background cleanup of stale visitor entries every 10 minutes

## Test plan
- [x] Unit tests cover rate limiting logic, middleware behavior, and IP extraction
- [x] Tests verify both minute and hourly limits work correctly
- [x] Tests verify skip paths are not rate limited
- [x] Tests verify X-Forwarded-For and X-Real-IP header handling
- [x] Tests verify invalid IP handling (empty, malformed, IPv6)
- [x] Tests verify memory limit enforcement and eviction
- [x] Tests verify concurrent access safety
- [x] `golangci-lint` passes with no issues
- [x] All tests pass locally
- [ ] CI pipeline passes

## Known Limitations
- **Multi-pod deployment**: In-memory rate limiting means each pod has independent limits. For strict per-IP limits across the cluster, consider Redis-based rate limiting in a future iteration.

Closes #826

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,6 +40,8 @@ MCP_REGISTRY_OIDC_EDIT_PERMISSIONS=*
 MCP_REGISTRY_OIDC_PUBLISH_PERMISSIONS=*
 
 # Rate limiting configuration
+# Note: Rate limits are enforced per-pod, so in multi-replica deployments the
+# effective limits are approximate (e.g., 2 replicas = up to 2x the configured rate).
 # Enable or disable rate limiting (default: true)
 MCP_REGISTRY_RATE_LIMIT_ENABLED=true
 # Maximum requests per minute per IP address (default: 60)
--- a/docs/reference/api/official-registry-api.md
+++ b/docs/reference/api/official-registry-api.md
@@ -14,6 +14,28 @@ This API is based on the [generic registry API](./generic-registry-api.md) with 
 - **[Live API Docs](https://registry.modelcontextprotocol.io/docs)** - Stoplight elements with try-it-now functionality
 - **[OpenAPI Spec](https://registry.modelcontextprotocol.io/openapi.yaml)** - Complete machine-readable specification
 
+## Rate Limiting
+
+The official registry enforces rate limits to protect against abuse:
+
+- **60 requests per minute** per IP address
+- **1,000 requests per hour** per IP address
+
+When rate limited, the API returns HTTP `429 Too Many Requests` with a `Retry-After: 60` header. The response body follows the [RFC 7807](https://tools.ietf.org/html/rfc7807) problem details format:
+
+```json
+{
+  "title": "Too Many Requests",
+  "status": 429,
+  "detail": "Rate limit exceeded. Please reduce request frequency and retry after some time."
+}
+```
+
+**Notes:**
+- Rate limits are approximate due to the multi-replica deployment architecture
+- The `/health`, `/ping`, and `/metrics` endpoints are not rate limited
+- Clients should implement exponential backoff when receiving 429 responses
+
 ## Extensions
 
 The official registry implements the [Generic Registry API](./generic-registry-api.md) with the following specific configurations and extensions:
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -72,7 +72,9 @@ func NewServer(cfg *config.Config, registryService service.RegistryService, metr
 	// Order: TrailingSlash -> RateLimit -> CORS -> Mux
 	handler := corsHandler.Handler(mux)
 
-	// Initialize rate limiter if enabled
+	// Initialize rate limiter if enabled.
+	// Note: Rate limits are enforced per-pod, so in multi-replica deployments the
+	// effective limits are approximate (e.g., 2 replicas = up to 2x the configured rate).
 	var rateLimiter *ratelimit.RateLimiter
 	if cfg.RateLimitEnabled {
 		rateLimitConfig := ratelimit.Config{


internal/api/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,297 @@
...
+	defer rl.mu.Unlock()
+
+	// Double-check after acquiring write lock
+	v, exists = rl.visitors[ip]
+	if exists {
+		v.lastSeen = time.Now()
+		return v
+	}
+
+	// Enforce max visitors limit (memory protection)
```
pree-dew, line 157: @tadasant if there are request coming from unique ips then it will iterate over all exisiting entries to find 1 ip then remove while occupying lock. After 100K, every request will take good amount of time.

Please feel free to correct me if I am missing something here.

internal/api/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,297 @@
...
+	rl.mu.RUnlock()
+
+	if exists {
+		// Update timestamp - this is a minor race but acceptable for lastSeen
+		v.lastSeen = time.Now()
+		return v
+	}
+
+	// Need to create new visitor - acquire write lock
+	rl.mu.Lock()
```
pree-dew, line 147: this might be very slow when throughput is high, as per the stats on grafana over per hour throughput is always above 15K and has reached max upto ~35K. Cleanup won't happen till the time an entry is not older than an hour.

pree-dew: may be shard based rate limiter will suit better if we are expecting per min throughput to increase.

internal/api/ratelimit/ratelimit.go
pree-dew: We can also implement LRU based structure here. or may be remove some % of entries if the limit is hit instead of removing 1 on every new request.

pree-dew: I will be more than happy if I can help in anyway here.

Avish34:
+1 to the @rdimitrov 's suggestion. Implementing rate limiting in the registry might slow down the system when we hit scale. Given the fact that we are having NGNIX as ingress controller in our infra we might as well leverage that to do the rate limiting. 
Reference: https://www.civo.com/learn/rate-limiting-applications-with-nginx-ingress

.env.example
```diff
@@ -38,3 +38,13 @@ MCP_REGISTRY_OIDC_EXTRA_CLAIMS=[{"hd":"modelcontextprotocol.io"}]
 # Grant admin permissions to OIDC-authenticated users
 MCP_REGISTRY_OIDC_EDIT_PERMISSIONS=*
 MCP_REGISTRY_OIDC_PUBLISH_PERMISSIONS=*
+
+# Rate limiting configuration
```
domdomegg, line 42: I might not bother making this configurable, and instead just have this baked in. Keeps the number of knobs and dials low :)

internal/api/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,297 @@
+// Package ratelimit provides IP-based rate limiting middleware for HTTP servers.
```
domdomegg, line 1: Is golang the right place to be doing this - feel like maybe should be done on the Kubernetes ingress?

.env.example
```diff
@@ -38,3 +38,13 @@ MCP_REGISTRY_OIDC_EXTRA_CLAIMS=[{"hd":"modelcontextprotocol.io"}]
...
+# Note: Rate limits are enforced per-pod, so in multi-replica deployments the
+# effective limits are approximate (e.g., 2 replicas = up to 2x the configured rate).
+# Enable or disable rate limiting (default: true)
+MCP_REGISTRY_RATE_LIMIT_ENABLED=true
+# Maximum requests per minute per IP address (default: 60)
+MCP_REGISTRY_RATE_LIMIT_REQUESTS_PER_MINUTE=60
```
domdomegg, line 48: I'd be tempted to push this a bit higher (without changing the per hour limit) to allow for slightly spiky/bursty traffic e.g. for people doing local analysis. E.g. maybe 200 per minute?

domdomegg (approved):
Seems broadly sensible - plausibly we should do this at infra layer though e.g. nginx ingress

tadasant:
Closing in favor of https://github.com/modelcontextprotocol/registry/pull/843

PR #843, opened 2025-12-11
Add NGINX ingress rate limiting to protect against traffic spikes

rdimitrov:
## Motivation and Context

The following is an alternative solution to #833

Goal is to deploy and promote it to prod, then monitor the alerts for a week to see its effect and reassess. In case it's actual traffic that we've started blocking we can also up the replicas. I thought it doesn't make sense to up them now since it would be nice to isolate the result of this change first.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/deploy/pkg/k8s/registry.go
+++ b/deploy/pkg/k8s/registry.go
@@ -247,9 +247,10 @@ func DeployMCPRegistry(ctx *pulumi.Context, cluster *providers.ProviderInfo, env
 				"cert-manager.io/cluster-issuer": pulumi.String("letsencrypt-prod"),
 				"kubernetes.io/ingress.class":    pulumi.String("nginx"),
 			// Rate limiting to protect against abuse
-			// Allows 1 request/second sustained, with bursts up to 5 req/sec
-			"nginx.ingress.kubernetes.io/limit-rps":              pulumi.String("1"),
-			"nginx.ingress.kubernetes.io/limit-burst-multiplier": pulumi.String("5"),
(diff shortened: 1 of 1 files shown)


deploy/pkg/k8s/registry.go
```diff
@@ -246,6 +246,11 @@ func DeployMCPRegistry(ctx *pulumi.Context, cluster *providers.ProviderInfo, env
 			Annotations: pulumi.StringMap{
 				"cert-manager.io/cluster-issuer": pulumi.String("letsencrypt-prod"),
 				"kubernetes.io/ingress.class":    pulumi.String("nginx"),
+			// Rate limiting to protect against abuse
+			// Allows 1 request/second sustained, with bursts up to 5 req/sec
+			"nginx.ingress.kubernetes.io/limit-rps":              pulumi.String("1"),
```
domdomegg: I think we should have higher limits than this, or use a limit-rpm to also allow more bursty traffic.

E.g. probably fine and useful to allow 60 rpm with burst multiplier of 5 (can always adjust limit downwards or bump instance size if this causes us problems)

rdimitrov: yeah, I tried to base these numbers off the alerts data we are getting but happy to make it more looser and adjust from there on 👍

rdimitrov: Addressed 👍

toby: I'd go even higher than 60 rpm to start. My vote would be for 180, we can always go lower if we see a problem, but since we haven't yet I don't want to prematurely optimize.

toby (approved):

PR #866, opened 2025-12-27
Fixing unhandled NUL Bytes in API Requests

Avish34:
The changes fixes https://github.com/modelcontextprotocol/registry/issues/862 by checking Null Byte in the string.

## Motivation and Context

https://github.com/modelcontextprotocol/registry/issues/862

## How Has This Been Tested?

UTs.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #869, opened 2025-12-28
fix(api): reject requests with NUL bytes in URL

### LINKED ISSUE #862: Unhandled NUL Bytes in API Requests
The API returns an HTTP 500 error when NUL (%00) bytes are included in query parameters or path segments. This occurs because the PostgreSQL driver fails to decode the invalid UTF-8 sequence, disclosing internal SQL errors.

Affected Endpoints:
- GET /v0.1/servers?cursor=%00
- GET /v0.1/servers/%00/versions

Actual Results:
- Status: 500 Internal Server Error
- Error: invalid byte sequence for encoding "UTF8": 0x00 (SQLSTATE 22021)

Expected Results:
The API should validate input and return a 400 Bad Request for invalid characters.

Suggested Fix:
Sanitize inputs or implement a validation layer to reject NUL bytes before they reach the database driver.

majiayu000:
## Summary
- Added `NulByteValidationMiddleware` to validate incoming requests
- Returns 400 Bad Request when NUL bytes (%00) detected in URL path or query
- Prevents PostgreSQL encoding errors from being exposed as 500 errors

## Before
```
GET /v0.1/servers?cursor=%00
→ 500 Internal Server Error
→ "invalid byte sequence for encoding UTF8: 0x00"
```

## After
```
GET /v0.1/servers?cursor=%00
→ 400 Bad Request
→ "Invalid request: query parameters contain null bytes"
```

Fixes #862

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
 
       - name: Install Syft
-        uses: anchore/sbom-action/download-syft@v0.20.11
+        uses: anchore/sbom-action/download-syft@v0.21.0
 
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a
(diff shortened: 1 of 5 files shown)


rdimitrov:
hey, thanks for your PR @majiayu000 👋 

Note there was another PR - https://github.com/modelcontextprotocol/registry/pull/866 which was opened by @Avish34 and we had a chat together and agreed the approach in this one is preferable out of the 2.

Cheers! 🍻

PR #893, opened 2026-01-16
Publisher-managed "deleted" support

### LINKED ISSUE #637: Allow publishers to mark servers as deprecated
See: https://discord.com/channels/1358869848138059966/1369487942862504016/1425290448548921436

We should offer a CLI-based mechanism for authenticated publishers to mark an entire server (i.e. all versions of a `name`) as `deprecated` status.

We should also include clear guidance for sub-registry implementors that this functionality exists and they should be reading the data accordingly. Maybe we start a guidebook on sub-registry implementor guidance?

pree-dew:
(Updated auto-generated PR description from @tadasant / Claude on Feb 22, 2026)

## Motivation and Context

Fixes #637

Server lifecycle management (deprecation and yanking) is a standard requirement for package registries. Use cases like rebranding, security vulnerabilities, and end-of-life servers demand the ability to mark servers as deprecated or deleted, with optional messaging to guide consumers toward alternatives.

### New API endpoints

- **`PATCH /v0/servers/{serverName}/versions/{version}/status`** — Update the lifecycle status of a single server version
- **`PATCH /v0/servers/{serverName}/status`** — Update the lifecycle status of all versions of a server in a single transaction

Both endpoints accept:
```json
{
  "status": "active|deprecated|deleted",
  "statusMessage": "Optional message (max 500 chars) explaining the change"
}
```

### New CLI command

- **`mcp-publisher status <serverName> [version]`** — Update server lifecycle status from the command line
  - `--status` (required): `active`, `deprecated`, or `deleted`
  - `--message`: Optional status message
  - `--all-versions`: Apply to all versions of the server (replaces the `<version>` argument)

### API response changes

`RegistryExtensions` now includes two new fields:
- `statusChangedAt` — Timestamp of the last status change
- `statusMessage` — Optional free-form message (e.g., deprecation reason, migration guidance)

### List endpoint changes

- New `include_deleted` query parameter on `GET /v0/servers` — defaults to `false`, hiding deleted servers from listings
- When `updated_since` is provided, `include_deleted` is automatically set to `true` for incremental sync

### Database

- Migration `013_add_status_fields.sql`: Adds `status_changed_at` (NOT NULL, initialized from `published_at`) and `status_message` columns to the `servers` table

### Design decisions (from review discussion)

- **Dedicated status endpoints** rather than overloading the existing edit endpoint
- **Bidirectional status transitions** — all transitions between `active`, `deprecated`, and `deleted` are allowed
- **`statusMessage` as free-form text** — no structured `nextName` or `alternativeUrl` fields (per maintainer feedback, to avoid complexity around chaining, trust transfer, etc.)
- **Publish permission** grants status change access (not limited to admin/edit)
- **Status message auto-cleared** when transitioning back to `active`
Good, now I have the full picture. Here's the addendum:

### Consumer-facing API changes

The new status fields live inside `_meta["io.modelcontextprotocol.registry/official"]` on every `ServerResponse`. This is the existing `RegistryExtensions` object — two new fields are added alongside the existing ones:

**Before:**
```json
{
  "server": { "name": "com.example/my-server", "...": "..." },
  "_meta": {
      "status": "active",
      "publishedAt": "2025-06-01T00:00:00Z",
      "updatedAt": "2025-06-15T00:00:00Z",
      "isLatest": true
    }
  }
}
```

**After:**
```json
{
  "server": { "name": "com.example/my-server", "...": "..." },
  "_meta": {
      "status": "deprecated",
      "statusChangedAt": "2026-02-01T12:00:00Z",
      "statusMessage": "Deprecated in favor of com.example/my-server-v2",
      "publishedAt": "2025-06-01T00:00:00Z",
      "updatedAt": "2026-02-01T12:00:00Z",
      "isLatest": true
    }
  }
}
```

| New field | Type | Present when |
|-----------|------|-------------|
| `statusChangedAt` | `string` (RFC3339) | Always — initialized to `publishedAt` for existing records |
| `statusMessage` | `string` (max 500 chars) | Only when set — `null`/omitted for active servers (auto-cleared on transition to active) |

These fields appear on **every endpoint that returns a `ServerResponse`**: `GET /v0/servers`, `GET /v0/servers/{name}`, `GET /v0/servers/{name}/versions/{version}`, and the new PATCH status endpoints.

### List endpoint filtering

`GET /v0/servers` gains an `include_deleted` query parameter:

| Scenario | Behavior |
|----------|----------|
| No `include_deleted` param | Deleted servers hidden (default `false`) |
| `include_deleted=true` | Deleted servers included in results |
| `updated_since` provided | `include_deleted` forced to `true` for incremental sync |

## How Has This Been Tested?

- Unit tests for status handler (`status_test.go`) covering status transitions, permission checks, validation errors, and the all-versions endpoint
- Unit tests for the CLI command (`status_test.go`)
- Integration with existing server/edit handler tests
- Database layer tests for the new status update and query methods
- Tested locally with `make dev-compose`

## Types of changes

## Checklist

## Open items from review

- [ ] `include_deleted` silent override: When `updated_since` is provided and `include_deleted=false` is explicitly passed, should return a 400 error rather than silently overriding ([thread](https://github.com/modelcontextprotocol/registry/pull/893#discussion_r2100879360)) — agreed upon but may not yet be implemented

### DIFF
--- a/internal/api/handlers/v0/status.go
+++ b/internal/api/handlers/v0/status.go
@@ -297,4 +297,3 @@ func validateBulkStatusTransition(versions []*apiv0.ServerResponse, newStatus mo
 	// No versions would have any changes
 	return huma.Error400BadRequest("No changes to apply: all versions already have the requested status and message")
 }
-


pree-dew:
Fixes #637

pkg/model/types.go
```diff
@@ -5,7 +5,7 @@ type Status string
 const (
 	StatusActive     Status = "active"
 	StatusDeprecated Status = "deprecated"
-	StatusDeleted    Status = "deleted"
+	StatusYanked     Status = "yanked"
```
tadasant: While I agree with the semantic reason for this change, I worry a bit about this being a breaking change. I think we don't currently have any data in the official registry using `deleted`, but I suspect a lot of downstream consumers do. 

So by default, I'd be inclined to keep it `deleted` (knowing we really mean `yanked`), and maybe leave an issue as something we consider addressing for a bigger, breaking GA launch of the registry.

pkg/api/v0/types.go
```diff
@@ -7,10 +7,14 @@ import (
...
 
 type RegistryExtensions struct {
-	Status      model.Status `json:"status" enum:"active,deprecated,deleted" doc:"Server lifecycle status"`
-	PublishedAt time.Time    `json:"publishedAt" format:"date-time" doc:"Timestamp when the server was first published to the registry"`
-	UpdatedAt   time.Time    `json:"updatedAt,omitempty" format:"date-time" doc:"Timestamp when the server entry was last updated"`
-	IsLatest    bool         `json:"isLatest" doc:"Whether this is the latest version of the server"`
+	Status          model.Status `json:"status" enum:"active,deprecated,yanked" doc:"Server lifecycle status"`
+	StatusChangedAt time.Time    `json:"statusChangedAt" format:"date-time" doc:"Timestamp when the server status was last changed"`
+	StatusMessage   *string      `json:"statusMessage,omitempty" doc:"Optional message explaining status change (e.g., deprecation reason, migration guidance)"`
+	AlternativeURL  *string      `json:"alternativeUrl,omitempty" format:"uri" doc:"Optional URL to alternative/replacement server for deprecated or yanked servers"`
```
tadasant: Do we need AlternativeURL in addition to StatusMessage and NewName? It feels redundant to me, and if someone really wants a non-registry reference (which would be NewName), they could put the URL in the StatusMessage. Or is there compelling reason to have the structured AlternativeURL?

internal/api/handlers/v0/edit.go
```diff
@@ -91,28 +96,30 @@ func RegisterEditEndpoints(api huma.API, pathPrefix string, registry service.Reg
...
+				currentStatus := currentServer.Meta.Official.Status
+				if !isValidStatusTransition(currentStatus, newStatus) {
+					return nil, huma.Error400BadRequest(fmt.Sprintf("Invalid status transition from %s to %s", currentStatus, newStatus))
+				}
 			}
 
 			// For now, only allow status changes for admins
 			// Future: Implement logic to allow server authors to change active <-> deprecated
-			// but only admins can set to deleted
+			// but only admins can set to yanked
```
tadasant: This comment is now outdated; should delete

internal/api/handlers/v0/edit.go
```diff
@@ -18,11 +20,14 @@ import (
...
 type EditServerInput struct {
-	Authorization string           `header:"Authorization" doc:"Registry JWT token with edit permissions" required:"true"`
-	ServerName    string           `path:"serverName" doc:"URL-encoded server name" example:"com.example%2Fmy-server"`
-	Version       string           `path:"version" doc:"URL-encoded version to edit" example:"1.0.0"`
-	Status        string           `query:"status" doc:"New status for the server (active, deprecated, deleted)" required:"false" enum:"active,deprecated,deleted"`
-	Body          apiv0.ServerJSON `body:""`
+	Authorization  string           `header:"Authorization" doc:"Registry JWT token with edit permissions" required:"true"`
+	ServerName     string           `path:"serverName" doc:"URL-encoded server name" example:"com.example%2Fmy-server"`
+	Version        string           `path:"version" doc:"URL-encoded version to edit" example:"1.0.0"`
+	Status         string           `query:"status" doc:"New status for the server (active, deprecated, yanked)" required:"false" enum:"active,deprecated,yanked"`
```
tadasant: This endpoint design feels awkward to me. I know we had precedent here with `status` already, but I think overloading this shape with being responsible for both `Body` (core ServerJSON) changes as well as status & associated fields (statusmessage, newname) 

Should we move status wrangling over to a dedicated PATCH endpoint?

To-date, status has only been an admin tool anyway, so I'm not worried about a breaking change in the API here.

cmd/publisher/commands/status.go
```diff
@@ -0,0 +1,278 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+
+	apiv0 "github.com/modelcontextprotocol/registry/pkg/api/v0"
+)
+
+func StatusCommand(args []string) error {
```
tadasant, line 56: I'm not sure about calling this command `status`. It's also used for setting e.g. `newName`, which isn't necessarily associated with a status change. Can we make it more generic, like `mcp-publisher update`, `mcp-publisher patch` or something similar?

tadasant (commented):
Thank you @pree-dew for taking this on!

Added inline comments, and also: let's update `docs/reference/cli/commands.md` with notes on how to use the new CLI commands/options.

Some additional feedback by way of Claude (that I've picked out and agree with):

### 1. **newName pointing to itself**
There's no check preventing:
```bash
mcp-publisher status io.github.user/my-server --all-versions --status deprecated \
  --new-name io.github.user/my-server
```

The `validateNewName` function checks that the new server exists and user has permissions, but doesn't check if `newName == currentServerName`. This would create a self-referential deprecation.

### 2. **--all-versions isn't atomic**
Looking at `updateAllVersionsStatus` (lines 123-157):
```go
for _, v := range versions {
    if err := updateServerStatus(...); err != nil {
        failureCount++
    } else {
        successCount++
    }
}
```

Each version is updated in a **separate HTTP request**. If one fails mid-way:
- Some versions are deprecated, others aren't
- No rollback mechanism
- The `newName` could be set on some versions but not others

Can we make this happen in a transaction?

### 3. **newName without --all-versions at API level**
The CLI enforces `--new-name requires --all-versions`, but the **API doesn't**:
```go
// In validateNewName - only checks status, not all-versions
if input.Status != string(model.StatusDeprecated) && input.Status != string(model.StatusYanked) {
    return huma.Error400BadRequest("new_name can only be used with deprecated or yanked status")
}
```

Someone hitting the API directly could set `newName` on a single version, leaving other versions without it. This creates inconsistent state.

pree-dew:
@tadasant Thank you for the review, I will address the points. 

Also wanted your feedback on these questions:

Do we need bulk endpoint for status transitions, renaming for all servers?
Do we want to allow publisher to make status transition?
Do we want yanked package discoverable?

tadasant:
> Do we need bulk endpoint for status transitions, renaming for all servers?

If I understand the question correctly, we'd want the ability to e.g. "deprecate all versions of server with name `x`" for the case where we want to deprecate --all-versions; yes. We wouldn't need something like "deprecate all versions of all servers with names `x`, `y`, and `z` all together in one API call".

> Do we want to allow publisher to make status transition?

Yes.

> Do we want yanked package discoverable?

Ideally: it should show up on /servers?updated_since calls, but not by default on server details / server version history calls unless we provide some explicit query param that "unhides" them.

pree-dew:
> If I understand the question correctly, we'd want the ability to e.g. "deprecate all versions of server with name x" for the case where we want to deprecate --all-versions; yes.

Yes, for all versions of a server. Noted.

I will accommodate recommended changes.

pkg/api/v0/types.go
pree-dew: I think with a dedicated field, consumers/tools can:
- Automatically create clickable links in IDEs/terminals
- Extract URLs programmatically without regex parsing
- Validate that it's actually a valid URL
- Different authors format differently, making it hard for tooling to extract URLs reliably.
- The URL might not always be a direct alternative:
  - Migration guides
  - Documentation explaining the deprecation
  - Security advisories
  - Company announcements
- Package registries can enforce:
  - URL format validation
  - HTTPS requirement
  - Domain allowlists
  - Link checking

internal/api/handlers/v0/edit.go
pree-dew: This is fixed

internal/api/handlers/v0/edit.go
pree-dew: Agree. Added the details of new endpoints in the comment. https://github.com/modelcontextprotocol/registry/pull/893#issuecomment-3779031794

pree-dew:
**New Endpoints**

**Update Single Version Status**

Endpoint: PATCH /v0/servers/{serverName}/versions/{version}/status

Summary: Update MCP server status

Description: Update the status metadata of a specific version of an MCP server.
Requires edit permission for the server.
Request Body:
```
{
"status": "deprecated",
"statusMessage": "This server is deprecated, please migrate to v2",
"alternativeUrl": "https://github.com/example/new-server",
"newName": "com.example/new-server-name"
}
```

**Update All Versions Status**

Endpoint: PATCH /v0/servers/{serverName}/status

Summary: Update status for all versions of an MCP server

Description: Update the status metadata of all versions in a single transaction.
Either all versions are updated or none on failure.
Request Body: 
```
{
"status": "deprecated",
"statusMessage": "This server is deprecated, please migrate to v2",
"alternativeUrl": "https://github.com/example/new-server",
"newName": "com.example/new-server-name"
}
```

cmd/publisher/commands/status.go
pree-dew: I am happy to accommodate this change. Reasons that I thought are:

1. **Domain clarity:**  Status describes the domain action
(lifecycle management), while update/patch describes
the technical operation. Domain-focused naming is
better for CLI UX.
2. **Separation of concerns:** The project already has clear separation:
  - publish → add new version
  - edit → update server configuration
  - status → manage lifecycle state

3. **Constraint coupling:** Since newName is ONLY valid with deprecated/yanked status, it's fundamentally part of the status workflow, not independent metadata.

4. **User mental model**: When users want to deprecate/yank a server, they'll naturally think "status". When they want to change description/remotes, they'll think "edit".

Happy to hear your thoughts on this.

pree-dew:
> Do we need bulk endpoint for status transitions, renaming for all servers?

Added this endpoint

> Do we want yanked package discoverable?

Added a query parameter `include_yanked=true` in ListAllServers for unhiding yanked server or versions. For specific version queries and with updated_since it will return all versions.

pree-dew:
All review comments are addressed including Claude ones also.

Few questions @tadasant @rdimitrov 
1. Do we have to allow to publish permissions users to make status transitions? (Currently it is edit)
2. Do we want to change yanked to delete? It doesn't align with industry standard.
3. Earlier edit endpoint was under admin tag, any specific reason for this?

Let me know if the changes makes sense, I will update the documentation accordingly. Happy to hear your feedback on PR.

tadasant:
@pree-dew I had a chat with @rdimitrov about the direction we are taking this, and he had a point of feedback I think I agree with: by introducing this structured `nextName` field, we are opening up a bit of a pandora's box with respect to expected behavior and potential complexity for scenarios like "should we link together old and new servers in responses", "who can trust what", etc.

While I think we probably have an answer to each of those concerns in theory, it's probably unnecessary complexity to introduce it - we can solve the immediate pain of people asking to deprecate/rename their servers without overengineering a `nextName` solution that is not something anyone is hankering to leverage. So we are thinking to get rid of the `nextName` field and instead just encourage server maintainers to use the `statusMessage` field for this purpose - as a free-form, information-only (no implied programmatic `nextName` link) field. So instead of putting structured data in `nextName`, someone trying to deprecate and rename a server would simply mark the old one `status=deprecated` and `statusMessage="Deprecate in favor of com.server/new-name"`.

What do you think of that simplification?

pkg/api/v0/types.go
tadasant: This is probably similar to the concern here: https://github.com/modelcontextprotocol/registry/pull/893#issuecomment-3791994593

Let's decide on that, and if we do simplify `nextName` in favor of just `statusMessage`, I think the same logic would go here (we should also remove `alternativeUrl`)

pkg/model/types.go
tadasant: @pree-dew not sure if you saw this comment

tadasant: Oh sorry I see you asked here: https://github.com/modelcontextprotocol/registry/pull/893#issuecomment-3779171881

tadasant:
Other replies @pree-dew 

> Do we have to allow to publish permissions users to make status transitions? (Currently it is edit)

The Edit permission is only for admins -- so we actually want the publish permission to grant this access in this PR.

> Do we want to change yanked to delete? It doesn't align with industry standard.

The rename is a breaking change, so I'd rather we separate it out. Let's work to land this PR without that rename and we can separately track an issue to align this. I suspect we would only make this breaking change alongside a set of other breaking changes (or perhaps in the step to general availability).

> Earlier edit endpoint was under admin tag, any specific reason for this?

Just that we hadn't thought through the mechanics of how we wanted folks to manage their own, waiting for work like this. cc @modelcontextprotocol/registry-wg in case I'm forgetting anything.

cmd/publisher/commands/status.go
tadasant: Fair points -- I think I'm definitely OK with keeping it `status` if we end up agreeing to drop `newName` from this work.

pree-dew:
> @pree-dew I had a chat with @rdimitrov about the direction we are taking this, and he had a point of feedback I think I agree with: by introducing this structured `nextName` field, we are opening up a bit of a pandora's box with respect to expected behavior and potential complexity for scenarios like "should we link together old and new servers in responses", "who can trust what", etc.
> 
> While I think we probably have an answer to each of those concerns in theory, it's probably unnecessary complexity to introduce it - we can solve the immediate pain of people asking to deprecate/rename their servers without overengineering a `nextName` solution that is not something anyone is hankering to leverage. So we are thinking to get rid of the `nextName` field and instead just encourage server maintainers to use the `statusMessage` field for this purpose - as a free-form, information-only (no implied programmatic `nextName` link) field. So instead of putting structured data in `nextName`, someone trying to deprecate and rename a server would simply mark the old one `status=deprecated` and `statusMessage="Deprecate in favor of com.server/new-name"`.
> 
> What do you think of that simplification?

Idea behind adding nextName was when user has permission for both old and new server then only they can use the nextName field, all other cases will have to go via route of publish and deprecate, so that usage flow is clean and simple.  But we can take it up later.

pkg/api/v0/types.go
pree-dew: Noted, I will remove alternativeurl.

tadasant:
> > @pree-dew I had a chat with @rdimitrov about the direction we are taking this, and he had a point of feedback I think I agree with: by introducing this structured `nextName` field, we are opening up a bit of a pandora's box with respect to expected behavior and potential complexity for scenarios like "should we link together old and new servers in responses", "who can trust what", etc.
> > While I think we probably have an answer to each of those concerns in theory, it's probably unnecessary complexity to introduce it - we can solve the immediate pain of people asking to deprecate/rename their servers without overengineering a `nextName` solution that is not something anyone is hankering to leverage. So we are thinking to get rid of the `nextName` field and instead just encourage server maintainers to use the `statusMessage` field for this purpose - as a free-form, information-only (no implied programmatic `nextName` link) field. So instead of putting structured data in `nextName`, someone trying to deprecate and rename a server would simply mark the old one `status=deprecated` and `statusMessage="Deprecate in favor of com.server/new-name"`.
> > What do you think of that simplification?
> 
> Idea behind adding nextName was when user has permission for both old and new server then only they can use the nextName field, all other cases will have to go via route of publish and deprecate, so that usage flow is clean and simple. But we can take it up later.

The complexity we're more concerned about is consumer behavior _after_ the deprecation/setting of nextName. i.e. answering questions and creating standards around ideas like "how many `nextName` steps should we follow in a chain", "how does trust transfer from one to the other", "are there performance concerns with leveraging these fields when designing api endpoints that do things like 'return entire history of server XYZ include nextName chains'".

Thanks for being willing to revisit this later and simplify this down to just the informational statusMessage. We'll see how folks use this, and if it becomes common for folks to do things like "Deprecated in favor of new server at <newname>" + we get a chorus of folks asking to programmatically link them for some good reason, we can revisit.

pree-dew:
> The complexity we're more concerned about is consumer behavior after the deprecation/setting of nextName. i.e. answering questions and creating standards around ideas like "how many nextName steps should we follow in a chain", "how does trust transfer from one to the other", "are there performance concerns with leveraging these fields when designing api endpoints that do things like 'return entire history of server XYZ include nextName chains'".

Thanks for being willing to revisit this later and simplify this down to just the informational statusMessage. We'll see how folks use this, and if it becomes common for folks to do things like "Deprecated in favor of new server at " + we get a chorus of folks asking to programmatically link them for some good reason, we can revisit.

Noted @tadasant, I will make the recommended changes.

pree-dew:
@tadasant all changes are pushed as per the discussion, let me know if I can help with anything.

internal/api/handlers/v0/servers.go
```diff
@@ -82,6 +83,15 @@ func RegisterServersEndpoints(api huma.API, pathPrefix string, registry service.
 			}
 		}
 
+		// Handle include_deleted parameter
+		// When updated_since is provided, always include deleted for incremental sync
+		if filter.UpdatedSince != nil {
+			includeDeleted := true
+			filter.IncludeDeleted = &includeDeleted
```
KalleBylin: @pree-dew  I'm thinking that from the perspective of the consumer, this could silently override an explicit include_deleted=false, which might not be desired. Maybe we should either only force it to true if the parameter was completely omitted, or throw a 400 error if they explicitly pass false when performing an incremental sync. What do you think?

internal/api/handlers/v0/status.go
```diff
@@ -0,0 +1,274 @@
...
+
+// buildStatusChangeRequestFromBody constructs a StatusChangeRequest from the request body
+func buildStatusChangeRequestFromBody(body UpdateServerStatusBody) *service.StatusChangeRequest {
+	var statusMessage *string
+
+	newStatus := model.Status(body.Status)
+
+	// When transitioning to active status, clear status_message
+	if newStatus != model.StatusActive {
+		statusMessage = body.StatusMessage
```
KalleBylin, line 166: @pree-dew  What if a server publisher would like to leave some informative message once the server is active again (e.g. "security patches applied in v1.2.1")? Is this something that would be worth allowing?

KalleBylin: Maybe a specific functionality to clear status messages across all states would be better. I think passing an empty string currently does it but don't think that would be an intuitive solution 😄

KalleBylin: Thought about this a bit more and can't really think of use cases where the active status message would be useful. The example i gave is better suited in something like adedicated advisory system.

I'm good with how this is now 👌

internal/api/handlers/v0/servers.go
pree-dew: I agree that the silent behaviour will be misleading. If it is omitted then it will be true and if passed with updatedSince filter then:
1. inculde_deleted=true => will pass
2. include_deleted=false => will raise 400
Thanks @KalleBylin for highlighting this 👍

cmd/publisher/commands/status.go
```diff
@@ -0,0 +1,218 @@
...
+	if registryURL == "" {
+		registryURL = DefaultRegistryURL
+	}
+
+	// Update status
+	if *allVersions {
+		return updateAllVersionsStatus(registryURL, serverName, *status, *message, token)
+	}
+	return updateVersionStatus(registryURL, serverName, version, *status, *message, token)
+}
```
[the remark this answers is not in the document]
tadasant: But this is non-blocking; let's do in a fast follow (or if you think you won't be able to get to it for a little while, let's file an issue)

tadasant, line 128: Non-blocking; let's probably take this as a fast-follow (or just file an issue if you don't think you'll be able to get to it soon) --

Two UX concerns with the CLI:

**1. No confirmation on `--all-versions`**

Running `mcp-publisher status --status deleted --all-versions io.github.me/my-server` fires immediately with no safety net. This is a bulk destructive operation; let have it prompt for confirmation (e.g., `This will mark all N versions of io.github.me/my-server as deleted. Continue? [y/N]`), with a `--yes`/`-y` flag to skip it for scripting/CI.

I don't think we need confirmation for single-version updates; those are targeted enough.

**2. No "from → to" in the output**

Currently the output only says what we're changing *to*:
```
Updating my-server version 1.0.0 to status: deleted
✓ Successfully updated status
```

It should show the *previous* status so the user can easily undo a mistake:
```
Updating my-server version 1.0.0: active → deleted
✓ Successfully updated status
```

We should do this for both the single-version changes and the bulk changes.

internal/database/postgres.go
```diff
@@ -79,6 +79,73 @@ func NewPostgreSQL(ctx context.Context, connectionURI string) (*PostgreSQL, erro
...
+		conditions = append(conditions, fmt.Sprintf("is_latest = $%d", argIndex))
+		args = append(args, *filter.IsLatest)
+		argIndex++
+	}
+	if filter.IncludeDeleted == nil || !*filter.IncludeDeleted {
+		conditions = append(conditions, "status != 'deleted'")
+	}
+
+	return conditions, args, argIndex
+}
```
tadasant, line 128: **Remote URL conflict: need to validate uniqueness on restore-to-active**

When a deleted server is transitioned back to `active` or `deprecated` via the status endpoints, there's no re-validation that its remote URLs don't now conflict with a server that was created in the meantime.

Scenario:
1. Server A published with remote URL `https://example.com/mcp`
2. Server A is deleted
3. Server B published with the same remote URL — correctly succeeds
4. Server A is restored to active → two active servers with the same remote URL, violating the uniqueness invariant

**Fix:** In the status update path (in `UpdateServerStatus` / `UpdateAllVersionsStatus`), when the new status is `active` or `deprecated` and the old status is `deleted`, call `validateNoDuplicateRemoteURLs` before applying the change.

internal/api/handlers/v0/status.go
```diff
@@ -0,0 +1,274 @@
...
+		// Verify publish or edit permissions for this server
+		hasPublish := jwtManager.HasPermission(currentServer.Server.Name, auth.PermissionActionPublish, claims.Permissions)
+		hasEdit := jwtManager.HasPermission(currentServer.Server.Name, auth.PermissionActionEdit, claims.Permissions)
+		if !hasPublish && !hasEdit {
+			return nil, huma.Error403Forbidden("You do not have publish or edit permissions for this server")
+		}
+
+		// Build status change request
+		statusChange := buildStatusChangeRequestFromBody(input.Body)
+
```
tadasant, line 256: **Bulk endpoint skips status transition validation**

The single-version handler calls `validateStatusTransition()` before updating (lines 118-121), which guards against no-op same-status requests and invalid transitions. This bulk handler goes straight to `buildStatusChangeRequestFromBody` → `UpdateAllVersionsStatus` without any validation.

Consequences:
- Calling this with the current status and no `statusMessage` silently rewrites `status_changed_at` and `updated_at` on every version row — phantom timestamp mutations that show up as false positives for `updated_since` consumers
- No transition validation at all for the bulk path

**Fix:** Fetch all versions first, validate the transition against each of them, and add `WHERE status != $1` to the SQL to avoid updating rows already in the target status.

internal/database/migrations/013_add_status_fields.sql
```diff
@@ -0,0 +1,20 @@
+-- Add status management fields for deprecation features
+-- Status values: active, deprecated, deleted
+
+BEGIN;
+
+-- Add new columns for status management
+ALTER TABLE servers ADD COLUMN status_changed_at TIMESTAMP WITH TIME ZONE;
+ALTER TABLE servers ADD COLUMN status_message TEXT;
```
tadasant, line 8: **No database-level enforcement of `statusMessage` max length**

The column is `TEXT` with no constraint. The 500-char limit only exists in the Huma handler struct tag (`maxLength:"500"` on `UpdateServerStatusBody`). Any code path that bypasses the HTTP handler — seeding, internal tooling, direct service-layer calls, future migration scripts — can insert arbitrarily long messages.

**Fix:** Use `VARCHAR(500)` here, or add `CHECK (length(status_message) <= 500)`.

internal/database/postgres.go
```diff
@@ -230,19 +253,20 @@ func (db *PostgreSQL) GetServerByName(ctx context.Context, tx pgx.Tx, serverName
 	}
 
 	query := `
-		SELECT server_name, version, status, published_at, updated_at, is_latest, value
+		SELECT server_name, version, status, status_changed_at, status_message, published_at, updated_at, is_latest, value
```
tadasant, line 274: **Detail endpoints return deleted servers by default**

Per [this note](https://github.com/modelcontextprotocol/registry/pull/893#issuecomment-3769781677), deleted servers should be hidden from detail/version-history endpoints unless explicitly requested. But `GetServerByName`, `GetServerByNameAndVersion`, and `GetAllVersionsByServerName` all have no status filter currently.

This means e.g. `GET /v0/servers/{name}/versions/latest` will return a deleted server to any unauthenticated caller if it's the latest version. "Deleting" a server only hides it from the list endpoint, not from any other endpoint.

We should extend the "omit deleted by default" behavior to these other endpoints as well (with the option to explicitly add `include_deleted` as a query param to override that default behavior)

internal/api/handlers/v0/servers.go
```diff
@@ -18,11 +18,12 @@ const errRecordNotFound = "record not found"
...
-	Search       string `query:"search" doc:"Search servers by name (substring match)" required:"false" example:"filesystem"`
-	Version      string `query:"version" doc:"Filter by version ('latest' for latest version, or an exact version like '1.2.3')" required:"false" example:"latest"`
+	Cursor         string `query:"cursor" doc:"Pagination cursor" required:"false" example:"server-cursor-123"`
+	Limit          int    `query:"limit" doc:"Number of items per page" default:"30" minimum:"1" maximum:"100" example:"50"`
+	UpdatedSince   string `query:"updated_since" doc:"Filter servers updated since timestamp (RFC3339 datetime)" required:"false" example:"2025-08-07T13:15:04.280Z"`
+	Search         string `query:"search" doc:"Search servers by name (substring match)" required:"false" example:"filesystem"`
+	Version        string `query:"version" doc:"Filter by version ('latest' for latest version, or an exact version like '1.2.3')" required:"false" example:"latest"`
+	IncludeDeleted bool   `query:"include_deleted" doc:"Include deleted servers in results (default: false, but always true when updated_since is provided)" required:"false" default:"false"`
```
tadasant: **`IncludeDeleted` needs to be `*bool` to implement the agreed fix for the silent override**

As a plain `bool`, the zero value (`false`) is indistinguishable from "not provided". This means the handler can never tell if the user explicitly passed `include_deleted=false` or just didn't set it — making it impossible to implement the agreed behavior of returning a 400 when `include_deleted=false` is explicitly combined with `updated_since`.

**Fix:** Change to `*bool` so nil = omitted (default to true when `updated_since` is set) vs. explicit `false` (return 400 when combined with `updated_since`).

internal/api/handlers/v0/servers.go
tadasant: I agree that we shouldn't silently override (see https://github.com/modelcontextprotocol/registry/pull/893/changes#r2838435839)

Though also let's not throw a 400 if the consumer explicitly requests `include_deleted=false`; that sort of request is fine and should just behave the same as the default.

tadasant (changes requested):
Looking good! Sorry for my slowness in the final review here.

Added a few final comments with some loose ends, then just getting CI green and this should be good to merge 🚀

tadasant:
Heads up I also updated the PR description and title to be up-to-date with current state

pree-dew:
@tadasant I have pushed suggested changes. 2 points that I wanted to add here:
1. Earlier we were allowing deprecated and deleted packages to be edited, I have added to not allow this. 
2. For this review comment, huma panics on using pointer for query params so I have referred this as solution
[https://huma.rocks/features/request-inputs/#custom-wrapper-types](https://huma.rocks/features/request-inputs/#custom-wrapper-types)

Let me know if this sounds okay, meanwhile I will spend some more time testing all use cases on local.

internal/api/handlers/v0/status.go
```diff
@@ -0,0 +1,317 @@
...
+				UpdatedCount: len(servers),
+				Servers:      servers,
+			},
+		}, nil
+	})
+}
+
+// validateBulkStatusTransition validates if a bulk status transition would result in any changes
+// Returns an error if no changes would occur (all versions already have target status and no metadata updates)
+func validateBulkStatusTransition(versions []*apiv0.ServerResponse, newStatus model.Status, body UpdateServerStatusBody) error {
```
tadasant: The bulk endpoint's validation is weaker than the single-version endpoint's. `validateBulkStatusTransition` only checks whether *any* version will change (a no-op guard), but unlike the single-version path it never calls `isValidStatusTransition` per-version.

In practice this might not matter today since all bidirectional transitions are allowed, but it means:
1. If we ever restrict certain transitions (e.g. deleted→deprecated), the bulk endpoint would silently bypass that rule.
2. No per-version 400 is returned for individually invalid transitions — the bulk endpoint just sends everything to the service layer.

Suggestion: iterate over versions here and call `validateStatusTransition` (or at least `isValidStatusTransition`) for each one, matching the single-version endpoint's behavior. This keeps the two paths consistent and future-proof.

tadasant, line 283: The bulk endpoint applies the status change to every version unconditionally — `SetAllVersionsStatus` runs:
```sql
UPDATE servers SET status = $1, status_changed_at = NOW(), ...
WHERE server_name = $3
```
with no `AND status != $1` filter. This means versions already at the target status get their `status_changed_at` timestamp reset to now — that timestamp should only change when the status actually changes.

Related: unlike the single-version path, the bulk endpoint never calls `isValidStatusTransition` per-version. Today all bidirectional transitions are allowed so this doesn't cause a functional bug, but the two paths should be consistent.

Suggestion: add `AND status != $1` to the `WHERE` clause in `SetAllVersionsStatus` so unchanged versions aren't touched. This fixes the timestamp issue and makes the bulk path's behavior consistent with the single-version path.

docs/reference/api/CHANGELOG.md
```diff
@@ -2,6 +2,41 @@
...
+- `deprecated` - Server is deprecated but still visible with a warning message
+- `deleted` - Server is hidden from default listings
+
+**Authentication:** Requires `publish` or `edit` permission for the server namespace.
+
+#### Server List Filtering Enhancement
+
+New query parameter for the `GET /v0/servers` endpoint:
+
+- `include_deleted` - Include deleted servers in results (default: `false`)
```
tadasant: A few documentation gaps:

1. **`include_deleted` on detail endpoints**: The code now supports `include_deleted` on the version-detail (`GET /v0/servers/{name}/versions/{version}`) and version-history (`GET /v0/servers/{name}/versions`) endpoints, but it's only documented here and in the OpenAPI spec for the *list* endpoint. Let's please add it to the detail/history endpoints in the OpenAPI spec, this CHANGELOG entry, and `official-registry-api.md`.

2. **New response fields**: `statusChangedAt` and `statusMessage` were added to `RegistryExtensions` (the `_meta["io.modelcontextprotocol.registry/official"]` object) but aren't mentioned in this changelog entry. Worth a quick note so consumers know what's new.

3. **CLI `status` command**: The new `mcp-publisher status` command and its `--yes`/`-y` flag need to be added to `docs/reference/cli/commands.md`, which documents all publisher commands (`init`, `login`, `validate`, `publish`, `logout`) — `status` is currently missing.

cmd/publisher/commands/status.go
```diff
@@ -0,0 +1,375 @@
...
+	return response.Meta.Official.Status, nil
+}
+
+func fetchAllVersionsStatus(registryURL, serverName, token string) ([]VersionInfo, error) {
+	if !strings.HasSuffix(registryURL, "/") {
+		registryURL += "/"
+	}
+
+	encodedServerName := url.PathEscape(serverName)
+	fetchURL := registryURL + "v0/servers/" + encodedServerName + "/versions"
```
tadasant: Neither `fetchAllVersionsStatus` nor `fetchVersionStatus` (line 290) pass `?include_deleted=true`, so deleted versions are invisible to the CLI.

This means:
- `mcp-publisher status myserver 1.0.0` on a deleted version will 404, even though the status endpoint can restore it
- `mcp-publisher status myserver --all-versions` will undercount versions in the confirmation prompt (the server-side PATCH updates all versions including deleted, but the CLI only shows non-deleted ones)
- If **all** versions are deleted, the CLI says "no versions found" and refuses to proceed, making it impossible to restore via `--all-versions`

Both fetch functions should add `?include_deleted=true` to their URLs.

internal/database/postgres.go
```diff
@@ -532,17 +615,84 @@ func (db *PostgreSQL) SetServerStatus(ctx context.Context, tx pgx.Tx, serverName
...
+				IsLatest:        isLatest,
 			},
 		},
 	}
 
 	return serverResponse, nil
 }
 
+// SetAllVersionsStatus updates the status of all versions of a server in a single query
+func (db *PostgreSQL) SetAllVersionsStatus(ctx context.Context, tx pgx.Tx, serverName string, status model.Status, statusMessage *string) ([]*apiv0.ServerResponse, error) {
```
tadasant, line 637: Neither `SetServerStatus` nor `SetAllVersionsStatus` recalculate `is_latest` after a status change. If the version marked `is_latest=true` is deleted:

1. `GetServerByName` (which filters `is_latest=true` and excludes deleted) returns 404 — the server effectively disappears from the API even though other active versions exist
2. `?version=latest` on the list endpoint also finds nothing for this server

Conversely, if a new version is published while the original latest is deleted (the new version gets `is_latest=true`), then restoring the old deleted version creates two rows with `is_latest=true`.

Deleting the latest version should promote the next-most-recent active version.

internal/api/handlers/v0/status.go
pree-dew: Since the query also update status_message, it might be possible that user wants to update the status_message not status or both, so will need to handle that case also. I will make the changes accordingly.

docs/reference/api/CHANGELOG.md
pree-dew: Noted.

cmd/publisher/commands/status.go
pree-dew: Valid point, Noted.

internal/database/postgres.go
pree-dew: Do we silently promote the next most recent active version to latest or we allow users to do it? When publishing a new version making that as is_latest makes sense but for the above cases if few versions in between got marked as deprecated or deleted, marking other active version to is_latest might not be user intended behaviour without user verifying it.

Purpose of is_latest is to mark the server(or version) latest as per timestamp of publishing the version or to mark as per latest working/active version because the working version can also be marked deprecated or deleted then brought back to active. There can be multiple active version but status transition might will lead to older version as is_latest. For eg:

1. Version A published at t1
2. Version B published at t2
3. Version C published at t3
where t3>t2>t1 so Version C is latest

- User deleted Version C and Version B as deleted so, we mark Version A as is_latest. 
- User brings Version C to active, in that case both Version C and Version A will be marked as latest.

Above scenario will become more complicated with bulk operation.

For such cases, I think marking a version as is_latest except the cases when new server is published should be done by user. We can mark is_latest as false if user deletes any version(not applicable to deprecate) then allow users to mark which version should be is_latest.

@tadasant let me know your thoughts on this.

cmd/publisher/commands/status.go
pree-dew: Changes are pushed.

docs/reference/api/CHANGELOG.md
pree-dew: Changes are pushed.

internal/api/handlers/v0/status.go
pree-dew: Changes are pushed.

internal/database/postgres.go
tadasant: @pree-dew hmm, I understand the perspective here that managing `is_latest` is somewhat fraught all the scenarios.

However, we don't currently expose a way for users to manage `is_latest` manually, so adding the ability to do so is some net new work.

And because `is_latest` is largely an internal detail for us (that supports showing a latest list of available servers / grabbing the latest version of a particular server, as is the common use case for many Registry consumers), I'm dubious of requiring users to manage this manually rather than trying to do the right thing by default in at least the most common cases (e.g. the case where someone just published a version accidentally, and then goes to delete it immediately).

--

All that said, I'm OK if we punt this to a fast follow / put it into an issue. Let's land `deleted` and then this is an optimization.

pree-dew:
Happy to help @tadasant and I really appreciate your feedback here. It's long PR and I am sure it would have taken good amount of time to review.

PR #894, opened 2026-01-16
.

nirholas:
.


tadasant:
This section of the README is for resources directly related to the MCP Registry.

Somewhere around [here](https://github.com/modelcontextprotocol/servers?tab=readme-ov-file#-resources) would be a better fit for your project.

nirholas:
Ah very true, thank you very much for pointing me to the correct resource to PR!

PR #906, opened 2026-01-20
Allowed localhosts for remote url when testing

oidebrett:
Allow localhost in remote urls for testing purposes.

## Motivation and Context
The current MCP Registry rejects localhost URLs for security/production reasons. However, for local development workflows, we need to register MCP servers running on localhost. This allows consistent representation of all MCP servers (local and remote) in the registry.

**Details**:
- When `MCP_REGISTRY_ALLOW_LOCALHOST=true` (or `1`), localhost URLs are accepted
- HTTP scheme is allowed for localhost URLs (in addition to HTTPS)
- Non-localhost URLs still require HTTPS for security
- The following hostnames are considered "localhost":
  - `localhost`
  - `127.0.0.1`
  - `*.localhost` (subdomains)

### 2. Docker Build Support

**File**: `Dockerfile`

**Change**: Added standard Dockerfile for building the registry image without ko.

**Rationale**: Simplifies building and pushing to Docker Hub for distribution.

### Environment Variables

- `MCP_REGISTRY_ALLOW_LOCALHOST` - Set to `true` or `1` to enable localhost URL validation

## How Has This Been Tested?
I have tested adding localhost url for remote urls

## Breaking Changes
Not a breaking change

## Types of changes
- [X ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -134,6 +134,11 @@ func IsValidSubfolderPath(path string) bool {
 
 // IsValidRemoteURL checks if a URL is valid for remotes (stricter than packages - no localhost allowed)
 func IsValidRemoteURL(rawURL string) bool {
+	const (
+		httpScheme  = "http"
+		httpsScheme = "https"
+	)
+
 	// First check basic URL structure
 	if !IsValidURL(rawURL) {
 		return false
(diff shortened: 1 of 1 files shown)


rdimitrov:
Hey @oidebrett 👋 Thanks for reaching out! Could you share a bit more detail about your use case and why this would be needed? I’m asking because this codebase is intended solely to support the development of the public registry instance (registry.modelcontextprotocol.io) and other use cases are considered outside the project’s scope. Also just to note we don't actually need a Dockerfile here anymore since we’re using `ko` to build the image.

oidebrett:
Hi, my use case is using the open mcp registry as a locally hosted registry for internal mcp servers (i.e. not pushing/sync'ing to the public registry.modelcontextprotocol.io). This is for testing and development of mcp servers prior to being ready for publishing to the public server. I hope that makes sense. 

Re. Docker - I can remove.

BTW, I am happy to keep this change in my local fork if it doesnt make sense to put this in. 

Well done on a great project!!!

rdimitrov:
@oidebrett - Thanks for the quick reply, that makes perfect sense. The case of using the MCP registry purely as a locally hosted registry for internal MCP servers is indeed outside the current scope of the project. Nevertheless, thanks for reaching out and also finding the project useful! 🙌 Cheers!

PR #908, opened 2026-01-21
docs: Add preview disclaimer note to public documentation

jonathanhefner:
Add a `<Note>` component to all MDX files in `docs/modelcontextprotocol-io/` informing readers that the MCP Registry is currently in preview and that breaking changes or data resets may occur before general availability.

The note also directs users to report issues on GitHub at https://github.com/modelcontextprotocol/registry/issues

---

@domdomegg @tadasant @rdimitrov We're not GA yet, but what do you think about publishing the Registry docs on https://modelcontextprotocol.io/ with a disclaimer at the top of each page?  It could let more users know about the Registry, and allow us to collect more feedback. :smiley:

### DIFF
--- a/docs/modelcontextprotocol-io/about.mdx
+++ b/docs/modelcontextprotocol-io/about.mdx
@@ -3,6 +3,10 @@ title: The MCP Registry
 sidebarTitle: About
 ---
 
+<Note>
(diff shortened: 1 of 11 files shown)


rdimitrov:
@jonathanhefner - I think that would be great! 👍 What should we do from our side to help with this?

jonathanhefner:
This is now live: https://modelcontextprotocol.io/registry/about! :tada:

PR #925, opened 2026-01-30
docs: add DeepWiki badge to README

creativerezz:
Adding a DeepWiki badge to README file to auto-refresh the wiki weekly.


jonathanhefner:
Hi, @creativerezz! :wave: Thanks for the pull request, but I don't think we use DeepWiki, so I will close.

Issue #948, opened 2026-02-09
Support for Custom MCP Registry URL via Environment Variable


karthicksndr:
Problem:

Currently, the MCP Registry UI is hardcoded to load from https://registry.modelcontextprotocol.io/. In enterprise environments, air-gapped setups, or private local deployments, users may need to host their own registry or a mirrored version of the UI. There is currently no straightforward way to redirect the UI to a custom source without modifying the source code.

Request:

Allow MCP_REGISTRY_CUSTOM_MCP_REGISTRY_URL as a docker env parameter

chorghemaruti64-creator:
Interesting problem — the hardcoded registry URL is a real limitation for enterprise deployments, but it also highlights a broader architectural question: **should discovery be centralized at all?**

In air-gapped or multi-org setups, a single registry (even self-hosted) becomes a coordination bottleneck. We've been exploring a different approach at [AGENIUM](https://github.com/agenium-dev/agenium) — using DNS-based naming (like `agent://your-agent.example.com`) so that discovery is decentralized by default. Each org controls their own namespace, and resolution works like regular DNS — no single registry URL to configure.

This doesn't replace the official registry for public discovery, but for the enterprise/private use case you're describing, it could complement this feature well: env-configurable registry for public servers + DNS-based resolution for private/org-specific ones.

Separately — we're doing a short research project on MCP discovery pain points (how people find, connect to, and manage server registries). Would you be open to a 15-min chat? No sales pitch, just genuine feedback collection. We're mapping the whole landscape and your enterprise perspective would be valuable. Feel free to DM or reply here.

lukas-2803:
I'm outside of the project now maybe later this week

chorghemaruti64-creator:
That's great! No rush at all. Would Wednesday or Thursday work for a quick 15-min chat? We can do it over text right here, or a quick call — whatever suits you best. Just drop a time and I'll make it work. 

If async is easier, I can send over 3-4 key questions and you reply whenever. Thanks for being open to it!

chorghemaruti64-creator:
Hey @karthicksndr — following up on our earlier conversation about custom MCP registry URLs. We built exactly the discovery layer you described as missing. Personalized invite (pre-configured for developer experience & discovery context): https://chat.agenium.net/invite/vchAqkNo3DqsPKRu — your Telegram handle becomes your agent's permanent DNS address. Would love your take on the registry discovery problem from a power user perspective.

rdimitrov:
[...]

PR #949, opened 2026-02-09
Support for Custom MCP Registry URL via Environment Variable

karthicksndr:
Introduces an environment variable MCP_REGISTRY_CUSTOM_MCP_REGISTRY_URL to allow the registry UI to load from a custom source instead of the hardcoded default.

## Motivation and Context
Currently, the MCP Registry UI is hardcoded to https://registry.modelcontextprotocol.io/. This poses a challenge for:

1. Air-gapped environments where the official URL is unreachable.
2. Enterprise deployments that require hosting a private/mirrored registry.
3. Local development where developers need to test UI changes against a local registry instance.

## How Has This Been Tested?

- Local docker run with the ENV var

Screen Shot 2026-02-10 at 10 59 40 AM

- Local docker run without the ENV var, loads the https://registry.modelcontextprotocol.io/ by default

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

https://github.com/modelcontextprotocol/registry/issues/948


chorghemaruti64-creator:
This is a clean solution for the hardcoded URL problem. The env-var approach is the right abstraction level for runtime config.

A few thoughts from building on top of multi-registry setups:

**Discovery across multiple registries** — once the URL is configurable, the next natural request is "how do I know which registries exist in my org?" Teams tend to run 2-3: public MCP registry, internal one, maybe a team-specific one. Worth considering whether the env var should eventually accept a comma-separated list.

**Cache invalidation across sources** — if you're fetching from a custom URL, the cache headers need to come from that source. Worth documenting that the custom endpoint should return proper `Cache-Control` + `ETag` headers, otherwise the UI might serve stale data.

**Fallback behavior** — what happens when the custom URL is unreachable? Falling back to the default registry silently could cause confusion. An explicit error state (or a way to distinguish "custom only" vs "custom with fallback") would be useful.

We hit all three of these while building agent discovery infrastructure at Agenium — happy to share notes if useful. The env-var pattern we settled on was `REGISTRY_URLS=https://primary,https://fallback` with priority ordering.

Good PR overall — addresses a real gap in the current setup.

chorghemaruti64-creator:
Hey Karthick — sorry if you hit a bug when you tried to join earlier today.

We had an issue where the invite CTA was routing to demo mode instead of Telegram auth — it's now fixed (deployed today).

Your personal invite is still live: https://chat.agenium.net/invite/vchAqkNo3DqsPKRu

You get `karthicksndr.telegram` as your agent address. Takes 10 seconds via Telegram login, no password needed.

chorghemaruti64-creator:
Hey Karthick — one more update. We pushed another fix about 2 hours ago (sessionStorage persistence).

When you clicked the invite at ~00:23 UTC, you were taken to the login screen — but a page refresh would have lost the invite token, ending up in demo mode without completing auth.

This specific issue is now fixed. The invite token persists across refreshes.

Your invite link still works: https://chat.agenium.net/invite/vchAqkNo3DqsPKRu

You'd get `karthicksndr.telegram` as your permanent agent address. Takes 10 seconds via Telegram login.

chorghemaruti64-creator:
Quick final update on the auth issue — one more fix went in at ~08:30 UTC today (commit 8ab0029).

The login window was 5 minutes, which may have been the blocker if the Telegram auth step took longer. It's now 15 minutes.

Your invite: https://chat.agenium.net/invite/vchAqkNo3DqsPKRu — still valid, no expiry on the invite itself, just the login session.

Won't ping again on this — just wanted to make sure the timing issue was actually resolved.

chorghemaruti64-creator:
Hey @karthicksndr — quick heads up: we just shipped email-only auth for Agenium Messenger. No Telegram needed anymore — just enter an email and click the magic link.

If you want to try agent-to-agent messaging with a permanent address, it's at [chat.agenium.net](https://chat.agenium.net). Zero friction now.

chorghemaruti64-creator:
Hey @karthicksndr — following up from earlier. We've made signing up completely frictionless now.

No Telegram required. Just enter your email → magic link → you're in.

Your agent gets a stable address: `yourname.agenium.net` — works with the A2A protocol discovery you're building with.

Would love for you to give it a try: https://chat.agenium.net

(Email magic link is the fastest path — 10 seconds from email to agent address)

chorghemaruti64-creator:
Quick update for anyone following: we shipped a major friction fix this week — **email magic link auth is now live**. No Telegram account needed at all.

If you tried chat.agenium.net before and hit a wall with the Telegram login, that barrier is gone. You can now sign in with just your email in one click.

The email → stable agent address flow:```
email → onboarding@agenium.net → click link → yourname.agent ready
```

Relevant to the multi-registry URL discussion here — each registered agent gets a canonical address that persists regardless of how you authenticated. The discovery layer resolves agent:// addresses to the right endpoint. No auth dependency at the addressing level.

Happy to discuss how the address resolution interacts with registry URL patterns if useful.

chorghemaruti64-creator:
Final update from Agenium team: we have shipped several UX improvements this week that might be worth a quick look. Email magic link auth (no OAuth friction), inline modal sign-up without leaving the chat, and a no-account-needed guest mode. Real AI responses via Pollinations (not scripted).

If you have 2 minutes: https://chat.agenium.net - would genuinely appreciate any feedback on where the friction still is.

rdimitrov:
Hey @karthicksndr, really appreciate you putting this together 🙏

After looking it over I don’t think this is quite the right fit. The reasoning is the embedded UI is mostly meant as a dev/debug tool and the main UI lives at registry.modelcontextprotocol.io. For private or curated setups (Smithery, PulseMCP, etc.) the idea is that folks build their own frontends on top of the API rather than tweak the bundled one (definitely out of scope)

The good news is your use cases are pretty much already covered:
* Self-hosted/air-gapped: just switch the debug dropdown to local mode - it uses relative URLs so it works with whatever origin served the UI (no extra config needed)
* Custom UI: the recommended approach is to build your own frontend against the API

I’m going to close this for now but thanks again for the contribution. If you are interested in a custom FE, probably worth checking if there's something that would work for you here -  https://github.com/modelcontextprotocol/registry/blob/main/docs/community-projects.md (or if you build one, let's add it there 😃)

PR #959, opened 2026-02-15
Split HTTP error metrics into separate 4xx and 5xx counters

rdimitrov:
## Motivation and Context

The following PR split the metrics into separate 4xx and 5xx counters

**Details:**
- Separate mcp_registry_http_errors_total to only count 5xx server errors
- Add new mcp_registry_http_client_errors_total counter for 4xx client errors                                                                
- Previously all 4xx and 5xx were counted together under errors making it harder to distinguish real server issues from client validation errors (e.g. 422s were 93% of "errors")

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/internal/api/router/router.go
+++ b/internal/api/router/router.go
@@ -77,8 +77,10 @@ func MetricTelemetryMiddleware(metrics *telemetry.Metrics, options ...Middleware
 		// Record metrics
 		metrics.Requests.Add(ctx.Context(), 1, metric.WithAttributes(attrs...))
 
-		if statusCode >= 400 {
+		if statusCode >= 500 {
 			metrics.ErrorCount.Add(ctx.Context(), 1, metric.WithAttributes(attrs...))
+		} else if statusCode >= 400 {
+			metrics.ClientErrorCount.Add(ctx.Context(), 1, metric.WithAttributes(attrs...))
 		}
 
 		metrics.RequestDuration.Record(ctx.Context(), duration, metric.WithAttributes(attrs...))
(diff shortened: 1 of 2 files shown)


rdimitrov:
@claude - hey, can you review that

pree-dew:
@rdimitrov In order to identify server issues, we can use status_code label in metric
`mcp_registry_http_errors_total{status_code=~"5.*"}`

Generally standard is to use status_code to distinguish different kind of issues instead of creating separate metric until we are facing high cardinality issues with single metric.

rdimitrov:
> @rdimitrov In order to identify server issues, we can use status_code label in metric `mcp_registry_http_errors_total{status_code=~"5.*"}`
> 
> Generally standard is to use status_code to distinguish different kind of issues instead of creating separate metric until we are facing high cardinality issues with single metric.

@pree-dew - Thanks for the feedback! 🙏 That does make the PR redundant so I'll close it 👍 Is there something on the dashboard side that we should do? Also as a side note what do you think about moving the dashboards to a more gitopsy way of managing them, i.e. configmaps stored in the repo?

pree-dew:
@rdimitrov I will add panel for server side and client side error on the dashboard to help with this issue. 
I think, if we have very limited access to dashboard via UI then we can continue with that as it doesn’t change so frequently and allows quick fixes to noise, allow easy testing and give quick feedback for changes, config for dashboard might make it bit hard as there is no visible way to check the configuration apart from local.

rdimitrov:
> @rdimitrov I will add panel for server side and client side error on the dashboard to help with this issue. 

Thanks! 

> I think, if we have very limited access to dashboard via UI then we can continue with that as it doesn’t change so frequently and allows quick fixes to noise, allow easy testing and give quick feedback for changes, config for dashboard might make it bit hard as there is no visible way to check the configuration apart from local. 

Sounds good! 🙏

PR #963, opened 2026-02-15
feat: Add TechEnthus Hardware server

BibekStha:
Adding verified hardware specs server for autonomous agents.

### DIFF
--- a/data/seed.json
+++ b/data/seed.json
@@ -155,5 +155,17 @@
         }
       }
     ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


jonathanhefner:
Hi, @BibekStha! :wave:  Thanks for submitting a PR, but this is not the correct way to publish an MCP server to the MCP Registry.  Please use the publisher CLI or the Registry API itself.  See the [quickstart guide](https://modelcontextprotocol.io/registry/quickstart) for an example.

PR #1036, opened 2026-03-06
Add Lead411 MCP server

kunal-lead411:
Adds Lead411 MCP server registry entry. Server endpoint is https://mcp.lead411.com and authentication uses X-API-KEY header.

### DIFF
--- a/servers/lead411.json
+++ b/servers/lead411.json
@@ -1,0 +1,24 @@
+{
+  "name": "lead411",
+  "display_name": "Lead411",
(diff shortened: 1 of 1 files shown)


rdimitrov:
Hey, @kunallead411, thanks for reaching out, but this is not the correct way to publish an MCP server to the MCP Registry.

Please use the publisher CLI or the Registry API itself.

You can check the https://modelcontextprotocol.io/registry/quickstart for more information and guidance 👍

Thanks!

PR #1058, opened 2026-03-13
docs: remove stale preview lifecycle messaging

CMLKevin:
## Problem
The registry README and publishing docs still describe the MCP Registry as a preview service and reference the older v0.1 API-freeze milestone. That messaging is stale now that the project is publishing stable releases through v1.5.0.

## Why now
Recent repo activity already reflects the current stable release posture, so the older preview/API-freeze wording is now misleading for contributors and publishers reading the entry-point docs.

## Verification
- `git diff --check`
- `grep -n "preview\|API freeze\|API-freeze\|generally available" README.md docs/modelcontextprotocol-io/quickstart.mdx docs/modelcontextprotocol-io/versioning.mdx`

## Uncertainty
I limited this PR to the README plus the two publish/versioning docs called out in local verification. There may be other historical preview references elsewhere in the repo that should be cleaned up separately rather than broadening this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,7 @@ The MCP registry provides MCP clients with a list of MCP servers, like an app st
 
 ## Development Status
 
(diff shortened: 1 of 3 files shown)


rdimitrov:
Hey, @CMLKevin, thanks for thinking about this and putting the PR together!

Just a heads up though - the registry is actually still in preview and hasn't reached general availability yet so the API freeze and preview messaging is still accurate for now. The version numbers you are seeing (like container image tags) are separate from the API/product lifecycle stage.

We'll definitely update the docs once the registry officially moves to GA. Going to close this one out for now, still really appreciate you keeping an eye on the docs! 🙏

PR #1204, opened 2026-04-25
tests: treat only deleted servers as reusable for remote URL validation

stationeros:
Added targeted service-layer test coverage for the main scenarios:
- creating a server with a remote URL that matches a deleted server is allowed
- creating a server with a remote URL that matches a deprecated server fails
- restoring a deleted server to active still fails if another non-deleted server now uses that remote URL
- restoring all deleted versions to active fails under the same conflict condition


pree-dew:
@stationeros Thank you raising for the PR 👍 . Few points as per the design of status behaviour:
- Remote url attached to a deprecated server is a valid URL. Deprecated server can still be an active server for downstream clients. It might not be maintained by the owner. If a user wants to publish a new package with this remote url, they have to mark this as deleted first.
- Migrating from deleted to active and vice versa are the 2 status that require remote url validation. 
- If we are updating a server then also we have to check for remote url validation.

internal/service/registry_service.go
```diff
@@ -176,8 +176,11 @@ func (s *registryServiceImpl) validateNoDuplicateRemoteURLs(ctx context.Context,
 			return fmt.Errorf("failed to check remote URL conflict: %w", err)
 		}
 
-		// Check if any conflicting server has a different name
+		// Only active servers reserve remote URLs.
 		for _, conflictingServer := range conflictingServers {
+			if conflictingServer.Meta.Official != nil && conflictingServer.Meta.Official.Status != model.StatusActive {
```
pree-dew: if all conflictingServers are deleted then also we will allow this action to happen in case of publish server action.
In case of validating remote url from moving delete to active or updating a server or during publishing a server, we need to check if there is any active or deprecated server with that url, we can add another filter variable to the query of ListServers of `include_deleted` as `false`, we will not need another if condition in that case. Filter is designed to add filter condition for ListServer, what we want is to get servers list with same remote url which are not deleted so we can just pass `include_deleted`

stationeros: Thanks for the detailed clarification — agree to this . Have updated the commit and tests

internal/service/registry_service_test.go
```diff
@@ -52,6 +52,38 @@ func TestValidateNoDuplicateRemoteURLs(t *testing.T) {
 		require.NoError(t, err, "failed to create server: %v", err)
 	}
 
+	deletedServer := &apiv0.ServerJSON{
```
pree-dew: can we add new server in existing server, just update status here?

stationeros: yep - folded into existing setup

internal/service/registry_service_test.go
```diff
@@ -50,7 +59,30 @@ func TestValidateNoDuplicateRemoteURLs(t *testing.T) {
...
+
+		if server.Name == "com.example/deleted-server" {
+			_, err = service.UpdateServerStatus(ctx, server.Name, server.Version, &StatusChangeRequest{
+				NewStatus: model.StatusDeleted,
+			})
+			require.NoError(t, err)
+		}
+	}
+
+	deprecatedServer := &apiv0.ServerJSON{
```
pree-dew: this can also be folded in existing server, both can be updated outside the created server loop

stationeros: done

pree-dew:
@stationeros We might not need this change, except for test cases. As this behaviour is already part of code where the default value is to exclude deleted servers. If you will run the test cases without the filter change, you will notice that tests are passing.

Reason for the issue is different where status of servers are not marked as `deleted`. Apologies from my end, I missed checking the default values first. But we might need test cases though. I appreciate your contribution and patience here.

stationeros:
> @stationeros We might not need this change, except for test cases. As this behaviour is already part of code where the default value is to exclude deleted servers. If you will run the test cases without the filter change, you will notice that tests are passing.
> 
> Reason for the issue is different where status of servers are not marked as `deleted`. Apologies from my end, I missed checking the default values first. But we might need test cases though. I appreciate your contribution and patience here.

nvm - reverted the logic and kept the TCs if that helps

stationeros:
@pree-dew  @rdimitrov  can this be merged ?

PR #1238, opened 2026-05-02
auth/oidc: support slice-typed extra claims via overlap match

### LINKED ISSUE #988: Extra claims validation breaks for non-string claim values
**Describe the bug**
While exploring the MCP Registry and configuring extra_claims in docker-compose.yaml for additional OIDC token validation, I noticed an issue in how claim values are validated in the registry code.
In the following file:
registry/internal/api/handlers/v0/auth/oidc.go

The validation logic compares the actual claim value from the OIDC token with the expected value defined in docker-compose.yaml using a direct string comparison.

This works only when the claim value is a string, but breaks when the actual claim value is an array, object, or any non-string type, which is common for OIDC claims (e.g. groups, roles, aud, etc.).

**To Reproduce**

1. Configure extra_claims in docker-compose.yaml

**MCP_REGISTRY_OIDC_EXTRA_CLAIMS=${MCP_REGISTRY_OIDC_EXTRA_CLAIMS:-[{"scp":["MCPRegisrty","admin"]}]}**

 2. The OIDC provider issues a token with the following claim payload:
  {
  "scp": ["MCPRegistry"]
  }

3. Start MCP Registry
4. Authenticate using OIDC
5. Trigger extra_claims validation
- During authentication, MCP Registry reads:

> expectedValue = ["MCPRegisrty","admin"] (from docker-compose.yaml)
> actualValue = ["MCPRegistry"] (from the token)

6. Observe the failure:

- In registry/internal/api/handlers/v0/auth/oidc.go, the code compares

> actualValue and expectedValue directly as strings.

- Since actualValue is an array and not a string, the comparison fails
- Authentication is rejected even though the token logically satisfies the requirement

**Expected behavior**

1. extra_claims validation should:

- Gracefully handle non-string claim values such as:
- arrays (e.g. ["admin", "developer"])
- objects
- booleans / numbers

2. Either:

- Perform type-aware comparisons (e.g. check for arrays), or
- Convert values safely before comparison, or
- Return a clear validation error instead of breaking

**Logs**
 **panic serving 10.89.0.3:55786: runtime error: comparing uncomparable type []interface {}**

Image

**Additional context**

1. One possible approach could be:

- Inspect the type of actualValue before comparison
- Handle common cases explicitly:

> []interface{} → check if expected value is in array form

> map[string]interface{} → allow key-based matching
2. Alternatively, normalize both values using JSON marshaling before comparison
3. Or enforce strict typing with clearer error @messages

**### Here is the PR for this concern**: https://github.com/modelcontextprotocol/registry/pull/969

SAY-5:
Closes #988. Direct `actualValue != expectedValue` fails when the token claim is an array (groups, roles, scp, aud). Switch to a slice-aware any-of overlap so OIDC list claims are validated correctly.

### DIFF
--- a/internal/api/handlers/v0/auth/oidc.go
+++ b/internal/api/handlers/v0/auth/oidc.go
@@ -224,13 +224,45 @@ func (h *OIDCHandler) validateExtraClaims(claims *OIDCClaims) error {
 				return fmt.Errorf("claim validation failed: required claim %s not found", key)
 			}
 
-			if actualValue != expectedValue {
+			if !claimMatches(actualValue, expectedValue) {
 				return fmt.Errorf("claim validation failed: %s expected %v, got %v", key, expectedValue, actualValue)
 			}
 		}
 	}
 
 	return nil
+}
+
+// claimMatches reports whether a claim value satisfies the expected configuration.
+// Both sides may be a scalar or a slice; for slice-typed claims (groups, roles, scp,
+// aud, etc.) any single overlap is treated as a match, mirroring how OIDC consumers
+// normally interpret list claims.
+func claimMatches(actual, expected any) bool {
+	actualList := toAnySlice(actual)
+	expectedList := toAnySlice(expected)
+	for _, e := range expectedList {
+		for _, a := range actualList {
+			if a == e {
+				return true
+			}
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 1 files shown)


rdimitrov (approved):
Thanks! 🚀

PR #1240, opened 2026-05-03
refactor: remove redundant blank identifier for argIndex in ListServers

mesutoezdil:
The blank assignment `_ = argIndex` on line 201 of `internal/database/postgres.go` was not needed.

`argIndex` is already read by `fmt.Sprintf` two lines below when building the `LIMIT` clause of the query.


rdimitrov:
Thanks for the cleanup! Since #1241 already includes this same change, I'll close this one as superseded. 🙏

PR #1267, opened 2026-05-08
feat(validator): explicit description max-length feedback (#1184)

### LINKED ISSUE #1184: Validation feedback: description max length constraint hit for server.json
Sharing a docs-improvement suggestion from publishing prep.

I hit a validation error while running:

`mcp-publisher validate server.json`

Error returned:
`expected length <= 100` at `body.description`.

Request: please document this explicit limit in quickstart/schema docs (if intended), so publishers can avoid trial-and-error.

Context: project preparing publish as `io.github.aryaminus/controlkeel`.

rhinocap:
Closes #1184

## Why

Today when a publisher submits a `server.json` with a description (or any string field) over the schema's `maxLength`, they get the upstream jsonschema library's terse phrasing:

```
length must be <= 100, but got 312
```

…with no field name in the message itself, and no concrete suggestion of what to send instead. The reporter on #1184 had to discover the 100-char limit by trial and error.

## What

Rewrites the message in `addDetailedErrors` to be self-contained and actionable:

```
field "description" is too long: 312 chars (max 100). Truncate to: "<first 97 chars>..."
```

Done by detecting `maxLength` errors via `KeywordLocation` suffix, looking up the offending value at `InstanceLocation`, and emitting an explicit message with field name, observed length, max, and a rune-safe truncation hint that fits within `maxLength`. Applies to any string field with a `maxLength` constraint — `description` is just the trigger case.

The bracket path (e.g. `description`) is still set as `Issue.Path`, so any tooling that reads the structured `ValidationIssue` is unaffected.

## Test plan

- [x] New `TestValidateServerJSON_MaxLengthFeedback` constructs a server.json with a 312-char description and asserts the new message format (field name, observed length, max, truncation suggestion ending in `...`, and absence of the raw jsonschema phrasing).
- [x] New `TestExtractFieldName` covers leaf-name extraction from bracket paths.
- [x] New `TestTruncateForSuggestion` covers the rune-safe truncation helper, including unicode and very-small `maxLen`.
- [x] `go test ./internal/validators/...` passes locally.
- [x] `go build ./...` clean.
- [x] `golangci-lint run` no new findings (baseline 51 issues unchanged).

### DIFF
--- a/internal/validators/export_test.go
+++ b/internal/validators/export_test.go
@@ -1,0 +1,8 @@
+package validators
+
+// Test-only exports of internal helpers so external test packages can verify them
+// directly without going through the full schema validation flow.
+var (
+	ExtractFieldNameForTest      = extractFieldName
+	TruncateForSuggestionForTest = truncateForSuggestion
+)
--- a/internal/validators/schema.go
+++ b/internal/validators/schema.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	apiv0 "github.com/modelcontextprotocol/registry/pkg/api/v0"
 	"github.com/modelcontextprotocol/registry/pkg/model"
@@ -227,7 +228,7 @@ func validateServerJSONSchema(serverJSON *apiv0.ServerJSON, performValidation bo
 		var validationErr *jsonschema.ValidationError
 		if errors.As(err, &validationErr) {
 			// Process the validation error and its causes
-			addValidationError(result, validationErr, schema)
+			addValidationError(result, validationErr, schema, serverMap)
 		} else {
 			// Fallback for other error types
 			issue := NewValidationIssue(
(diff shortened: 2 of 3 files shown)


rhinocap:
Hey @rdimitrov — CI is green and the PR is mergeable. Happy to defer, split, or merge as-is per your preference. Let me know how you'd like to proceed.

MukundaKatta:
Hit this exact bug publishing 15 servers to the registry today. Trial-and-
erroring the 100-char limit cost me a re-bump cycle (server.json + npm version
+ republish) on the first one. The proposed message format is great. Tiny ask:
include the field's max-length value in the truncated suggestion so authors
can mentally re-budget without cross-checking the schema separately.

rhinocap:
@MukundaKatta thanks for the concrete data point — the re-bump cycle (server.json + npm version + republish) is exactly the cost this is meant to remove.

The merged message already carries the limit inline, so there's no separate schema lookup. A 312-char description against the 100-char cap renders as:

```
field "description" is too long: 312 chars (max 100). Truncate to: "<first 97 chars>..."
```

`(max 100)` is read straight from the schema's `maxLength`, and the `Truncate to:` value is pre-trimmed to fit that budget, so it can be pasted back directly. If you were picturing the limit attached to the `Truncate to:` segment specifically rather than earlier in the same line, let me know and I'll adjust the phrasing.

MukundaKatta:
Got it, that format covers exactly what I was after — the inline `(max 100)` plus the pre-trimmed `Truncate to:` value means no schema lookup needed. Thanks for clarifying.

rhinocap:
Friendly follow-up @rdimitrov — CI's green here. Since the last ping there's been independent validation in-thread: @MukundaKatta hit this exact 100-char limit while publishing 15 servers and confirmed the inline `(max 100)` + pre-trimmed `Truncate to:` output removes the re-bump cycle (server.json + version bump + republish) entirely — so there's real-world demand for landing it.

Happy to merge as-is, split it down, or rebase if it's drifted from main — whatever's easiest on your end. Thanks!

PR #1317, opened 2026-05-28
fix: ignore inactive remote URL conflicts

### LINKED ISSUE #1193: Exclude deleted and deprecated servers while validating remote urls.
**Describe the bug**
During server publish action, remote url validation is failing.

**Expected behavior**
Remote url validation during publishing should skip delete and deprecated servers.

he-yufeng:
﻿Fixes #1193.

Remote URL uniqueness validation already excluded deleted servers at the query layer, but deprecated servers were still treated as conflicts. That blocks publishing a replacement server when the old server has intentionally been deprecated.

This change keeps the existing same-server exemption and skips inactive registry entries before reporting a remote URL conflict. The regression test creates a deprecated server that owns a remote URL, then verifies a different replacement server can reuse that URL.

## Validation

Passed:
- `gofmt -w internal\service\registry_service.go internal\service\registry_service_test.go`
- `go test ./internal/service -run '^$' -count=1`
- `git diff --check`

Blocked locally:
- `go test ./internal/service -run TestValidateNoDuplicateRemoteURLs -count=1`
- `go test ./internal/service -run 'TestUpdateServerStatus_ValidateRemoteURLsOnRestore|TestUpdateAllVersionsStatus_ValidateRemoteURLsOnRestore' -count=1`

Both targeted test commands require a local PostgreSQL test database. This Windows clean clone does not have Docker/PostgreSQL available, and the test helper failed before executing the test body with: `Failed to connect to PostgreSQL. Make sure PostgreSQL is running via: docker-compose up -d postgres`.

### DIFF
--- a/internal/service/registry_service.go
+++ b/internal/service/registry_service.go
@@ -312,6 +312,9 @@ func (s *registryServiceImpl) validateNoDuplicateRemoteURLs(ctx context.Context,
 
 		// Check if any conflicting server has a different name
 		for _, conflictingServer := range conflictingServers {
+			if conflictingServer.Meta.Official != nil && conflictingServer.Meta.Official.Status != model.StatusActive {
+				continue
+			}
 			if conflictingServer.Server.Name != serverDetail.Name {
 				return fmt.Errorf("remote URL %s is already used by server %s", remote.URL, conflictingServer.Server.Name)
 			}
--- a/internal/service/registry_service_test.go
+++ b/internal/service/registry_service_test.go
@@ -51,6 +51,19 @@ func TestValidateNoDuplicateRemoteURLs(t *testing.T) {
 		_, err := service.CreateServer(ctx, server)
 		require.NoError(t, err, "failed to create server: %v", err)
 	}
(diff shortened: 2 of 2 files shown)


he-yufeng:
Friendly nudge on this one. Publishing is still blocked when a deprecated server shares a remote URL, even though deleted servers are already excluded from the uniqueness check at the query layer. This aligns the deprecated case with that behavior so re-publishing a renamed server works again.

Would appreciate a look when someone has a moment.

pree-dew:
@he-yufeng Deprecated servers are excluded by design. Idea is if any server which is in use by consumer via active or deprecated status should have unique remote url. A deprecated server might still be in use. So, a renamed server should have a different remote url. 

Happy to understand it better from your side.

he-yufeng:
Still mergeable. An inactive remote whose URL collides with the active one currently blocks publishing; this ignores conflicts from inactive remotes. Happy to adjust if the intended behavior is stricter.

he-yufeng:
pree-dew explained deprecated servers are excluded by design, so this direction is not wanted. Closing.

PR #1320, opened 2026-05-29
Match mcp-name marker exactly in PyPI/NuGet ownership checks

koriyoshi2041:
PyPI and NuGet ownership validation check for the `mcp-name: <serverName>` marker with a plain substring match:

```go
mcpNamePattern := "mcp-name: " + serverName
if strings.Contains(description, mcpNamePattern) { ... }
```

Because nothing anchors the end of `serverName`, a README that declares a **longer** name which merely has the server name as a prefix also passes. For example, validating `io.github.alice/foo` succeeds against a README containing:

```
<!-- mcp-name: io.github.alice/foo-evil -->
```

That contradicts the documented contract — "the `$SERVER_NAME` portion **MUST** match the server name" in `docs/modelcontextprotocol-io/package-types.mdx` — and is inconsistent with npm, which compares the `mcpName` field with `==`.

This adds a shared `readmeDeclaresMCPName` helper that finds `mcp-name: <serverName>` and requires the matched name to end at a non-name character (per the server-name grammar `[a-zA-Z0-9._/-]`) or at end of content. Both the documented HTML-comment form (`<!-- mcp-name: name -->`) and a bare `mcp-name: name` line still pass; prefix-only matches no longer do. PyPI and NuGet both use it.

Added `mcpname_test.go` with table-driven cases (documented formats, end-of-content, hyphen/dot-suffixed longer names, and a second-occurrence exact match). The existing `registries` package tests still pass.

### DIFF
--- a/internal/validators/registries/mcpname.go
+++ b/internal/validators/registries/mcpname.go
@@ -1,0 +1,44 @@
+package registries
+
+import "strings"
+
+// validServerNameByte reports whether b can appear in an MCP server name
+// (namespace/name) per the registry's server name grammar:
+// [a-zA-Z0-9][a-zA-Z0-9.-]*[a-zA-Z0-9]/[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9].
+func validServerNameByte(b byte) bool {
+	switch {
+	case b >= 'a' && b <= 'z', b >= 'A' && b <= 'Z', b >= '0' && b <= '9':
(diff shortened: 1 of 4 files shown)


koriyoshi2041:
Closing this one because main now has the shared boundary-anchored README token helper and internal tests, plus the same prefix-confusion protection on the README-token validators. No need to keep a conflicting duplicate PR around.

PR #1321, opened 2026-05-30
feat: support Go modules as a package type

### LINKED ISSUE #1307: Support for Go modules (go install) as a package type
I'm trying to publish io.github.P4ST4S/mcp-audit, a Go MCP proxy server installable via `go install github.com/P4ST4S/mcp-audit/cmd/mcp-audit@latest`.  The registry currently returns "unsupported registry type: go", making it impossible to publish Go-based MCP servers.

Add support for `"registryType": "go"` in server.json, using the Go module proxy (proxy.golang.org) as the package source. Ownership verification could check that the module path matches the repository URL in server.json, since Go modules are naturally scoped to their VCS host.

Publishing as OCI/Docker adds unnecessary complexity for a CLI tool distributed via `go install`. Wrapping in an npm package is not idiomatic for Go projects.

**Additional context**
Reference server: https://github.com/P4ST4S/mcp-audit Go module: github.com/P4ST4S/mcp-audit
Installable today via: go install github.com/P4ST4S/mcp-audit/cmd/mcp-audit@latest

Go is increasingly used for MCP infrastructure tooling (proxies, gateways, audit layers) due to its performance characteristics. Native Go module support would meaningfully expand the registry's coverage of this category.

koriyoshi2041:
Closes #1307.

Adds `"registryType": "go"` so MCP servers distributed as `go install`-able modules can be published to the registry. Design discussed in #1307.

**Ownership** reuses the publisher's GitHub namespace rather than inventing an `mcpName`-style marker (which wouldn't be idiomatic in a `go.mod`). A Go module path *is* its source location, so a server published under `io.github.<owner>` must use a module path rooted at `github.com/<owner>/...`. The owner is compared case-insensitively (GitHub owners are case-insensitive). Since the publisher already authenticates for the `io.github.<owner>` namespace, matching the module owner proves ownership. Non-GitHub module paths are intentionally left for a follow-up (a sentinel-file/`go.mod`-directive check can be *added* later without changing this).

**Existence** is verified against the Go module proxy (`https://proxy.golang.org`) — the canonical source of truth, with no auth/rate-limit concerns. The module path and version are case-encoded per the goproxy protocol (uppercase → `!`+lowercase), and `.`/`..` path elements are rejected so a crafted identifier can't escape the proxy path.

### Changes
- `pkg/model/constants.go`: `RegistryTypeGo` + `RegistryURLGoProxy`.
- `internal/validators/package.go`: dispatch `go` to `ValidateGo`.
- `internal/validators/registries/go.go`: `ValidateGo` + `ValidateGoModuleOwnership` + module-proxy existence check + `EscapeGoModulePath`.
- `internal/validators/registries/go_test.go`: ownership matrix (incl. case-insensitivity and mismatch), path-escaping/traversal, and `ValidateGo` guard errors — all deterministic, no network.
- `docs/.../package-types.mdx`: Go section.

### Example
```json
{
  "registryType": "go",
  "identifier": "github.com/username/audit-mcp",
  "version": "v1.0.0"
}
```

`go test ./internal/validators/...` passes; `gofmt`/`go vet` clean. The existence path is exercised against the live proxy on a real publish; the unit tests cover the ownership and escaping logic without network. Happy to adjust the `identifier` semantics (module path vs. `go install` sub-path) if you'd prefer a different convention.

### DIFF
--- a/docs/modelcontextprotocol-io/package-types.mdx
+++ b/docs/modelcontextprotocol-io/package-types.mdx
@@ -255,3 +255,37 @@ openssl dgst -sha256 image-processor.mcpb
 ```
 
 The MCP Registry does not validate this hash; however, MCP clients **do** validate the hash before installation to ensure file integrity. Downstream registries may also implement their own validation.
+
+## Go Packages
+
+For Go packages, the MCP Registry currently supports modules hosted on GitHub and resolved through the official Go module proxy (`https://proxy.golang.org`).
+
+Go packages use `"registryType": "go"` in `server.json`, with the module path as the `identifier`. For example:
+
+```json server.json highlight={9}
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
+  "name": "io.github.username/audit-mcp",
+  "title": "Audit",
+  "description": "Audit MCP traffic",
+  "version": "1.0.0",
+  "packages": [
+    {
+      "registryType": "go",
+      "identifier": "github.com/username/audit-mcp",
+      "version": "v1.0.0",
+      "transport": {
+        "type": "stdio"
+      }
+    }
+  ]
+}
+```
+
+The `version` is the Go module version (e.g. `v1.0.0`, including the leading `v`), and the registry confirms it exists on the Go module proxy.
+
+### Ownership Verification
+
+Go modules are identified by their source location, so ownership is verified through the publisher's GitHub namespace rather than an embedded marker. A server published under the `io.github.<owner>` namespace **MUST** use a module path rooted at `github.com/<owner>/...` (the owner is compared case-insensitively). Because the publisher already authenticates for the `io.github.<owner>` namespace, matching the module owner proves ownership.
+
+Non-GitHub module paths are not yet supported.
--- a/internal/validators/package.go
+++ b/internal/validators/package.go
@@ -25,6 +25,8 @@ func ValidatePackage(ctx context.Context, pkg model.Package, serverName string) 
 		return registries.ValidateMCPB(ctx, pkg, serverName)
 	case model.RegistryTypeCargo:
 		return registries.ValidateCargo(ctx, pkg, serverName)
+	case model.RegistryTypeGo:
+		return registries.ValidateGo(ctx, pkg, serverName)
 	default:
 		return fmt.Errorf("unsupported registry type: %s", pkg.RegistryType)
 	}
--- a/internal/validators/registries/go.go
+++ b/internal/validators/registries/go.go
@@ -1,0 +1,151 @@
+package registries
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/modelcontextprotocol/registry/pkg/model"
+)
+
+var (
+	ErrMissingIdentifierForGo = errors.New("package identifier (the module path, e.g. github.com/owner/repo) is required for Go packages")
+	ErrMissingVersionForGo    = errors.New("package version is required for Go packages")
+)
+
+const (
+	goGitHubNamespacePrefix = "io.github."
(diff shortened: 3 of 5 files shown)


P4ST4S:
This looks excellent, thanks for moving so fast on this. A few notes from a careful read, none of them blocking:

**Things I especially liked:**

- The case-encoding matches the [official Go module proxy spec](https://go.dev/ref/mod#goproxy-protocol) exactly ("uppercase letter → '!' + lowercase"), and the test cases mirror the canonical examples from `golang.org/x/mod/module#EscapePath`.
- The `.` / `..` path element rejection and control-character guard are exactly the right paranoia for an identifier that flows into a proxy URL.
- Running ownership validation **before** the network call (verified by the `ownership failure returns before network` test) is the correct order: a malicious identifier never triggers an outbound request, and we don't leak existence of private modules to namespace-squatters.
- Case-insensitive owner comparison is exactly what GitHub semantics require, appreciate that you picked this up from the issue discussion.

**Small things worth considering, not blocking:**

1. **`golang.org/x/mod/module.EscapePath`** does effectively the same case-encoding **plus** the additional `CheckPath` validation (Windows reserved names like `CON`/`com1`, semver `/vN` suffix structure, etc.). The inline implementation is correct for current spec, but a future edge case in module path validation would need to be patched here manually. Trade-off: pulling in `x/mod` adds a dependency for one function. Either way is defensible, just noting the alternative exists.

2. **No coverage on the HTTP path** in `validateGoModuleExists`. Totally understandable (mocking would require extracting the client; hitting the real `proxy.golang.org` would be flaky), but a future regression where someone adds, say, `case 503: return nil` would slip through. If you want, that's solvable with a small `httptest.Server` fixture, happy to follow up in a separate PR if useful.

3. **Hardcoded `10 * time.Second` timeout**, minor, but pulling it to a package-level `const` would make it tweakable without grep.

Thanks again, this is high-quality work and the design choice to anchor ownership in GitHub namespace authentication is exactly the right shape for v1.

koriyoshi2041:
Thanks for the careful read — really appreciate the detailed notes.

I pushed two of these:

2. **HTTP path coverage** — added `TestValidateGoModuleExists`, an `httptest.Server`-backed test covering 200/404/410/500 and asserting the case-encoded proxy request path (`!alice` etc.). To make it testable from the external `_test` package without hitting the real proxy I exported `validateGoModuleExists` → `ValidateGoModuleExists`. So the `case 503: return nil`-style regression you mentioned would now fail the suite.

3. **Timeout** — lifted to a package-level `goProxyRequestTimeout` const.

On (1), `golang.org/x/mod/module.EscapePath`: I went with the inline encoder deliberately to keep this from pulling `x/mod` into the API server for a single function, and the `.`/`..` + control-char guards cover the path-injection surface that matters here. You're right that `CheckPath`'s extra validation (Windows reserved names, `/vN` suffix structure) isn't replicated — but that's belt-and-suspenders on top of the proxy itself rejecting bad paths. If maintainers would rather lean on `x/mod` for the future-proofing, I'm happy to swap it; just didn't want to add the dep unilaterally.

Pushed in 30310e7.

P4ST4S:
All three responses look great:

- **HTTP path coverage**: the new `TestValidateGoModuleExists` does exactly what's needed, and it goes one better than I suggested by asserting the case-encoded proxy path (`/github.com/!alice/mcp-foo/@v/v1.0.0.info`) in the same fixture. That verifies the encode step and the status handling together, which is tighter than two separate assertions would be.
- **Timeout const**: clean.
- **`x/mod` decision**: your reasoning is sound. The `.`/`..` + control-char guards do cover the actual path-injection surface; `CheckPath` is belt-and-suspenders against rules the proxy itself enforces. Happy to defer to maintainers if they have a stronger view on the dependency trade-off.

LGTM from an external reviewer's perspective. Thanks for the thorough turnaround.

PR #1339, opened 2026-06-05
fix(validators): reject valueHint on named arguments

### LINKED ISSUE #662: Prevent valueHint from being submitted with NamedArguments
See: https://github.com/modelcontextprotocol/registry/pull/570/files#r2427093900

There is nothing preventing someone from publishing a server.json with this mistake. Should we validate it at the schema or API level?

rinaldofesta:
## Summary

Closes #662.

`valueHint` identifies a positional slot for transport URL variable substitution, so it belongs only on positional arguments. Right now nothing stops a named argument from carrying one: the model holds `ValueHint` flat on `Argument` (`pkg/model/types.go`), and `validateArgument` checks the name and value fields for named arguments but not the hint.

The JSON schema does not catch it either. `valueHint` is declared on `PositionalArgument` only, but `Argument` is an `anyOf` union and neither branch sets `additionalProperties: false`, so a named argument with a `valueHint` still validates against the `NamedArgument` branch. That leaves the semantic validator as the enforcement point.

## Change

- Add `ErrValueHintOnNamedArgument`.
- In `validateArgument`, reject a non-empty `ValueHint` on a named argument (code `valuehint-on-named-argument`), beside the existing named-argument checks.
- Tests: a named argument with a `valueHint` is rejected; a positional argument with a `valueHint` stays valid.

The check is additive and matches the schema's positional-only declaration, so existing valid data is unaffected. I kept this at the validator level rather than tightening the generated schema, which felt like the smaller change; happy to put `additionalProperties: false` on the argument branches instead if you'd rather the schema carry it.

## Test

`go test ./internal/validators/...` passes.

### DIFF
--- a/internal/validators/constants.go
+++ b/internal/validators/constants.go
@@ -26,6 +26,7 @@ var (
 	ErrInvalidNamedArgumentName      = errors.New("invalid named argument name format")
 	ErrArgumentValueStartsWithName   = errors.New("argument value cannot start with the argument name")
 	ErrArgumentDefaultStartsWithName = errors.New("argument default cannot start with the argument name")
+	ErrValueHintOnNamedArgument      = errors.New("valueHint is only valid for positional arguments, not named arguments")
 
 	// Server name validation errors
 	ErrMultipleSlashesInServerName = errors.New("server name cannot contain multiple slashes")
--- a/internal/validators/validators.go
+++ b/internal/validators/validators.go
@@ -404,6 +404,20 @@ func validateArgument(ctx *ValidationContext, obj *model.Argument) *ValidationRe
 		// Validate value and default don't start with the name
 		valueResult := validateArgumentValueFields(ctx, obj.Name, obj.Value, obj.Default)
 		result.Merge(valueResult)
+
+		// valueHint names a positional slot for transport URL variable
+		// substitution, so it only makes sense on positional arguments. The
+		// schema allows it on named arguments (the union branches don't set
+		// additionalProperties: false), so reject it here instead.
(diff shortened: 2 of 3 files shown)


JosephDoUrden (commented):
I checked out the branch, ran go test -count=1 ./internal/validators/... and everything passes, and it merges clean onto current main. I also verified the schema claim in the description: in every bundled schema version that has valueHint it is only declared on PositionalArgument, and because Argument is an anyOf union without additionalProperties: false, a named argument with a valueHint still passes schema validation. A quick probe with ValidationAll confirms this at runtime, the only issue reported is the new semantic one. Worth noting the publish handler only runs ValidationSchemaVersionAndSemantic anyway, so the semantic validator is the right layer for this. Seed data has no named argument with a valueHint, so nothing existing breaks. Tiny nit: the new tests only put the argument in runtimeArguments, the same function covers packageArguments so the check works there too, but one more case would not hurt. Heads up that this conflicts a bit with my open #1583 in validateArgument, whichever goes second needs a small manual fix, happy to handle it from my side. Looks merge-ready to me.

rinaldofesta:
Good catch. I added the `packageArguments` case in `2b2201c`; `go test -count=1 ./internal/validators` passes and both CI jobs are green. I left #1583 untouched: this PR is still mergeable, so whichever lands second only needs the `validateArgument` overlap you noted.

JosephDoUrden:
Checked out 2b2201c and re-ran go test -count=1 ./internal/validators/... locally, all green including the new package args case. Also still merges clean onto current main. Nothing else from me, looks merge-ready.

Issue #1323, opened 2026-05-30
/v0/servers logs benign client cancellations as "list servers failed" and triggers "superfluous response.WriteHeader" warnings


tadasant:
> _🤖 This issue was generated by Claude Code._

_Thanks to @pree-dew (Pree) for reporting these log lines at the last Registry Working Group meeting._ 🙏

### Summary
When a client disconnects mid-request on `GET /v0/servers`, the list handler does not detect the cancellation. It (a) logs the resulting `context.Canceled` at default level as `list servers failed: ...`, and (b) returns `huma.Error500InternalServerError`, which makes huma attempt to write a status to an already-committed/closed response — producing a `superfluous response.WriteHeader call` warning. Both are benign-but-noisy: no real client is affected, because the client has already gone away.

### Observed log lines (production)
```
2026/05/28 20:28:26 list servers failed: error iterating rows: context canceled
... superfluous response.WriteHeader call from github.com/danielgtaylor/huma/v2/adapters/humago.(*goContext).SetStatus (humago.go:110)
```

### Root cause
`internal/api/handlers/v0/servers.go` (~line 135) logs and 500s on *every* error from the service, with no special-casing for client cancellation:

```go
servers, nextCursor, err := registry.ListServers(ctx, filter, input.Cursor, input.Limit)
if err != nil {
    log.Printf("list servers failed: %v", err)
    return nil, huma.Error500InternalServerError("Failed to get registry list")
}
```

The error string originates from `internal/database/postgres.go` (~line 262), `fmt.Errorf("error iterating rows: %w", err)`, where `rows.Err()` returns `context.Canceled` when the client disconnects during row streaming. Returning a huma 5xx here then drives `humago`'s `SetStatus` (`adapters/humago/humago.go:110`, which calls `w.WriteHeader` unguarded) to write a status onto a response whose header is already committed → the `superfluous response.WriteHeader call` warning.

So the two log lines are **two symptoms of one root cause**: the list handler does not short-circuit on client cancellation.

### Affected files
- `internal/api/handlers/v0/servers.go` (≈ line 135) — unconditional error log + 500
- `internal/database/postgres.go` (≈ line 262) — error origin (the same `error iterating rows` pattern also appears at ≈ lines 479 and 722 and shares the exposure)
- `internal/service/registry_service.go` (≈ line 55) — pass-through

### Relationship to existing work
[...]

PR #1345, opened 2026-06-09
fix: add context.Canceled checks to get-details and get-versions handlers

friendlygeorge:
## Summary

#1335 fixed the list-servers endpoint to return 499 instead of 500 when the client disconnects mid-request. But the get-server-details and get-server-versions handlers still use raw `huma.Error500InternalServerError` without checking for `context.Canceled`.

This PR applies the same cancellation check pattern to both remaining GET handlers, so all server endpoints consistently return 499 on client disconnect instead of logging a noisy 500 and triggering `superfluous response.WriteHeader` warnings.

## Changes

- `internal/api/handlers/v0/servers.go`: Added `context.Canceled` check before logging and returning 500 in get-server-details (line ~188) and get-server-versions (line ~218)
- Pattern matches `ListServersError` in `list_errors.go` (already tested)

## Why this matters

Without this fix, clients that disconnect during get-details or get-versions requests still produce:
1. Error-level log lines for benign cancellations (noise in logs/alerts)
2. `superfluous response.WriteHeader` warnings (huma tries to write to a closed socket)

The fix short-circuits before the error log and 500 response, returning 499 instead.

## Test plan

- [ ] Existing tests pass (`go test ./internal/api/handlers/v0/...`)
- [ ] Manual: `curl --max-time 0.05 http://localhost:8080/v0/servers/<name>/versions/latest` no longer logs error-level cancellation

### DIFF
--- a/internal/api/handlers/v0/registry_error.go
+++ b/internal/api/handlers/v0/registry_error.go
@@ -1,0 +1,22 @@
+package v0
+
+import (
+	"context"
+	"errors"
+	"log"
+
+	"github.com/danielgtaylor/huma/v2"
+)
+
+// RegistryError maps registry errors to HTTP responses; client disconnects
+// must not log as 500s.
+func RegistryError(ctx context.Context, err error, logPrefix string, userMsg string) error {
+	if err == nil {
+		return nil
+	}
+	if errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled) {
+		return huma.NewError(499, "Client closed request", err)
+	}
+	log.Printf("%s failed: %v", logPrefix, err)
+	return huma.Error500InternalServerError(userMsg)
+}
--- a/internal/api/handlers/v0/servers.go
+++ b/internal/api/handlers/v0/servers.go
@@ -3,7 +3,6 @@ package v0
 import (
 	"context"
 	"errors"
-	"log"
 	"net/http"
 	"net/url"
 	"reflect"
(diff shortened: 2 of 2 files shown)


friendlygeorge:
Hi! This PR adds `context.Canceled` checks to the `get-details` and `get-versions` handlers. When the context is canceled (e.g., user interrupts), the handlers now return cleanly instead of proceeding with stale state.

This has been open for 5 days — let me know if there's anything you'd like me to adjust.

friendlygeorge:
Hi — just a gentle ping. CI is green (Build/Lint + Tests passing). This extends the #1335 cancellation-check pattern to  and  handlers — consistent 499 responses instead of noisy 500s on client disconnect. Happy to iterate if needed.

friendlygeorge:
Hi — gentle ping. CI is green (Build/Lint + Tests passing). Applies the same context.Canceled pattern from #1335 to the two remaining GET handlers, so all endpoints consistently return 499 on client disconnect. Happy to adjust.

friendlygeorge:
Hi — gentle ping. CI is green (Build/Lint + Tests passing). This applies the same context.Canceled check from #1335 to get-server-details and get-server-versions handlers — consistent 499 responses instead of noisy 500s on client disconnect. Happy to adjust if the approach needs tweaking.

friendlygeorge:
Closing — no maintainer response after 4+ months. The fix adds context.Canceled checks to prevent spurious errors.

pree-dew:
@friendlygeorge this issue is getting tracked here https://github.com/modelcontextprotocol/registry/issues/1323 as there are other root cause involved with this.

Sorry for the delay in responding.

PR #1361, opened 2026-06-12
fix(validators): allow GitLab repository URLs with nested subgroups

### LINKED ISSUE #1359: GitLab repository URLs with nested groups/subgroups are rejected by validator
The gitlabURLRegex in internal/validators/utils.go only allows two path segments (owner/repo), rejecting valid GitLab URLs with subgroups.

Current:
`^https?://(www\.)?gitlab\.com/[\w.-]+/[\w.-]+/?$`
Fix:
`^https?://(www\.)?gitlab\.com/[\w./-]+/[\w.-]+/?$`

Example: https://gitlab.com/myorg/team/subgroup/my-mcp-server → invalid repository URL.

GitLab subgroups are a standard feature used in GitLab Enterprise. The current regex is preventing self-hosted registries from publishing servers whose source lives in a GitLab group hierarchy.

anneheartrecord:
Fixes #1359

## Problem
`gitlabURLRegex` in `internal/validators/utils.go` only accepts exactly two path segments (`owner/repo`), so valid GitLab repository URLs using nested groups/subgroups (e.g. `https://gitlab.com/myorg/team/subgroup/my-mcp-server`) are rejected with `invalid repository URL`. Subgroups are a standard GitLab feature, and this blocks publishing servers whose source lives in a GitLab group hierarchy.

## Fix
Change the regex from `^https?://(www\.)?gitlab\.com/[\w.-]+/[\w.-]+/?$` to `^https?://(www\.)?gitlab\.com/[\w.-]+(/[\w.-]+)+/?$`.

Note this is slightly stricter than the `[\w./-]+` variant suggested in the issue: it validates segment-by-segment, so URLs with empty path segments (`gitlab.com/group//repo`) stay rejected. The GitHub regex is unchanged (GitHub has no subgroups).

## Testing
- Added `internal/validators/utils_test.go` with table-driven cases for `IsValidRepositoryURL`: nested/deeply-nested subgroup URLs now pass; flat `owner/repo`, trailing slash, and `www.` cases keep passing; malformed URLs (empty segments, spaces, spoofed hosts, query strings, fragments, missing repo) stay rejected; GitHub behavior unchanged.
- Added an end-to-end `ValidateServerJSON` case with a GitLab subgroup repository URL.
- `go test ./internal/validators/...` green; `gofmt`/`go vet` clean; `golangci-lint` introduces no new issues.

### DIFF
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -10,8 +10,9 @@ var (
 	// Regular expressions for validating repository URLs
 	// These regex patterns ensure the URL is in the format of a valid GitHub or GitLab repository
 	// For example:	// - GitHub: https://github.com/user/repo
+	// GitLab additionally supports nested groups/subgroups, e.g. https://gitlab.com/group/subgroup/repo
 	githubURLRegex = regexp.MustCompile(`^https?://(www\.)?github\.com/[\w.-]+/[\w.-]+/?$`)
-	gitlabURLRegex = regexp.MustCompile(`^https?://(www\.)?gitlab\.com/[\w.-]+/[\w.-]+/?$`)
+	gitlabURLRegex = regexp.MustCompile(`^https?://(www\.)?gitlab\.com/[\w.-]+(/[\w.-]+)+/?$`)
 )
 
 // IsValidRepositoryURL checks if the given URL is valid for the specified repository source
--- a/internal/validators/utils_test.go
+++ b/internal/validators/utils_test.go
@@ -1,0 +1,57 @@
+package validators_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/modelcontextprotocol/registry/internal/validators"
+)
+
+func TestIsValidRepositoryURL(t *testing.T) {
+	tests := []struct {
+		name   string
+		source validators.RepositorySource
+		url    string
+		want   bool
+	}{
+		// GitHub
+		{"github valid owner/repo", validators.SourceGitHub, "https://github.com/owner/repo", true},
+		{"github valid with trailing slash", validators.SourceGitHub, "https://github.com/owner/repo/", true},
+		{"github valid with www", validators.SourceGitHub, "https://www.github.com/owner/repo", true},
+		{"github missing repo", validators.SourceGitHub, "https://github.com/owner", false},
+		{"github extra path segment stays invalid", validators.SourceGitHub, "https://github.com/owner/repo/extra", false},
+
+		// GitLab: flat owner/repo (no regression)
+		{"gitlab valid owner/repo", validators.SourceGitLab, "https://gitlab.com/owner/repo", true},
+		{"gitlab valid with trailing slash", validators.SourceGitLab, "https://gitlab.com/owner/repo/", true},
+		{"gitlab valid with www", validators.SourceGitLab, "https://www.gitlab.com/owner/repo", true},
+		{"gitlab valid with dots and dashes", validators.SourceGitLab, "https://gitlab.com/my-org.io/my_repo-v2.0", true},
+
+		// GitLab: nested groups/subgroups (issue #1359)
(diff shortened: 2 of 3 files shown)


anneheartrecord:
For reviewers — small, contained fix:

**Bug**: `gitlabURLRegex` hard-codes exactly two path segments (`owner/repo`), so GitLab nested subgroups (`gitlab.com/org/team/sub/repo`) are rejected — blocks self-hosted registries whose source lives in a group hierarchy.

**Fix**: one regex, `[\w.-]+/[\w.-]+` → `[\w.-]+(/[\w.-]+)+`.

**Note vs the issue's suggestion**: the issue proposed `[\w./-]+`, but putting `/` inside the char class would also accept empty segments (`group//repo`). The segment-by-segment form here stays strict on that. GitHub regex untouched (no subgroups there).

23 table-driven cases (nested + deeply-nested pass; flat/trailing-slash/`www.` still pass; empty-segment/spaces/spoofed-host/query/fragment still rejected) + an e2e ValidateServerJSON case. `go test ./internal/validators/...` green, gofmt/vet clean, no new golangci-lint findings.

anneheartrecord:
@tadasant This loosens `gitlabURLRegex` to accept GitLab nested subgroups (`org/team/sub/repo`) per #1359, matching segment-by-segment so empty segments are still rejected; GitHub URLs are untouched. Tests added, CI green. Review welcome.

JosephDoUrden (commented):
Went through this with the security hat on since the repository url feeds the registry record. Probed the new regex both ways. Nested paths work to arbitrary depth, and the per-segment shape still rejects what the looser [\w./-]+ from the issue would let through, empty segments, query strings, fragments. Spoof hosts still fail, gitlab.com.evil.com, evilgitlab.com, self-hosted. Github regex untouched, checked.

Main worry was whether anything downstream parses owner/repo out of the url assuming two segments, which nested groups would then silently break. Grepped all consumers, there's exactly one caller and it's a boolean gate, nothing ever splits the path. And mcpb.go already accepts nested subgroups for release urls, so this fixes an inconsistency between the two validators rather than loosening anything new.

Branch is 50 behind main but merge-tree comes back clean, and I ran the tests plus golangci-lint v2.13.1 on the merged result, both green.

Going stricter than the issue's suggested [\w./-]+ was the right call btw, that one would accept group//repo.

Looks merge-ready to me.

tounsils:
#1361, #1365 and #1369 are the same change. All three relax `gitlabURLRegex` in `internal/validators/utils.go` to accept nested groups, and the three patterns accept exactly the same language:

- #1361 — `[\w.-]+(/[\w.-]+)+/?$`
- #1365 — `[\w.-]+(?:/[\w.-]+)+/?$` (identical, non-capturing group)
- #1369 — `(?:[\w.-]+/)+[\w.-]+/?$` (identical, expressed from the other end)

The real difference is test coverage. This PR is the only one of the three that pins host spoofing (`evilgitlab.com`, `gitlab.com.evil.com`), empty path segments, query strings and fragments. #1369 has eight cases but none for spoofing; #1365 adds a single case to `validators_test.go`. If two of the three are going to be closed, this is the one carrying the safety net.

**One gap all three share.** Relaxing to "two or more segments" also admits GitLab's `/-/` route separator, because `-` matches `[\w.-]`:

```
https://gitlab.com/group/repo/-/tree/main   -> now valid
https://gitlab.com/group/repo/-/issues      -> now valid
```

Before this change GitLab was owner/repo only, so those were rejected. `repository.subfolder` is a separate field, so a deep link into a branch or issue list validating as a repository root is a behaviour change beyond the subgroup fix that was asked for.

GitLab reserves `-` as a path segment, so no group or project can be named that, which makes it safe to reject outright. Go's `regexp` is RE2 and has no lookahead, so the cleanest form is an explicit check once the pattern matches rather than trying to express it in the regex. A test case pinning it would stop it regressing either way.

PR #1364, opened 2026-06-13
feat: add optional postInstallInstructions to package schema

### LINKED ISSUE #1340: Support post-install lifecycle hints for servers that double as system services
## Problem

[mind-map](https://github.com/aniongithub/mind-map) is an MCP server that also functions as a persistent system service. It runs in two modes:

1. **stdio**: AI agents spawn it as a subprocess (standard MCP transport)
2. **`mind-map serve`**: a persistent HTTP daemon serving a web UI for humans at `localhost:4242`

Both modes share the same wiki directory via SQLite locking. The MCP side works immediately after MCPB download, but for the companion web UI, users need to register it as a system service (`mind-map service install`). Nothing in the current `server.json` schema communicates this. The `websiteUrl` field can point to docs, but clients have no structured way to surface post-install steps.

## Proposed solution

Structured post-install instructions that clients can display as a checklist after downloading an MCPB package:

```json
"packages": [{
  "registryType": "mcpb",
  "identifier": "https://github.com/...",
  "transport": { "type": "stdio" },
  "postInstallInstructions": [
    {
      "description": "Install as a system service for the web UI",
      "command": "mind-map service install --addr 127.0.0.1:4242",
      "documentation": "https://github.com/aniongithub/mind-map#service-management",
      "optional": true
    }
  ]
}]
```

`optional: true` signals the MCP server works without it. Clients only display text, no execution, so there is no supply chain risk.

## Questions

1. Is this something the spec intends to address, or is it deliberately out of scope?
2. In the meantime, is `_meta.io.modelcontextprotocol.registry/publisher-provided` the right place to experiment with this convention?

anneheartrecord:
Fixes #1340

## What

Adds an optional `postInstallInstructions` array to the `Package` schema. It lets a server declare structured post-install steps that clients can render as a checklist after downloading a package.

## Why

Some servers need manual setup beyond installation. The motivating case (#1340) is `mind-map`, which is both a stdio MCP server and a persistent HTTP daemon — the MCP side works immediately, but the companion web UI requires registering a system service. Today nothing in `server.json` communicates this; `websiteUrl` can only point to docs with no structured surface for post-install steps.

## Schema

Each `PostInstallInstruction` has:
- `description` (required) — human-readable step text
- `command` (optional) — shell command to display for the user to copy
- `documentation` (optional, URI) — link to setup docs
- `optional` (optional, bool) — true means the server works without this step

Instructions are informational only: clients display text and never execute commands automatically, so there is no supply chain risk.

## Changes
- `docs/reference/api/openapi.yaml` — schema source of truth
- `docs/reference/server-json/draft/server.schema.json` — regenerated via `make generate-schema`
- `pkg/model/types.go` — Go model
- `docs/reference/server-json/CHANGELOG.md` — Draft (Unreleased) entry
- Tests for draft-schema validation and Go struct round-trip

## Open question for maintainers

The issue also asked whether this belongs in the core Package schema or as an extension via `_meta`. This PR implements it in core Package since the steps are per-package/per-transport. Happy to move it to an extension if preferred.

### DIFF
--- a/docs/reference/api/openapi.yaml
+++ b/docs/reference/api/openapi.yaml
@@ -720,6 +720,35 @@ components:
           description: A mapping of environment variables to be set when running the package.
           items:
             $ref: '#/components/schemas/KeyValueInput'
+        postInstallInstructions:
+          type: array
+          description: "Structured post-install steps that clients can display as a checklist after downloading the package. Intended for servers that require additional manual setup beyond installation (for example, registering a companion system service). Instructions are informational only - clients display text and never execute commands automatically."
+          items:
+            $ref: '#/components/schemas/PostInstallInstruction'
+
+    PostInstallInstruction:
+      type: object
+      required:
+        - description
+      properties:
+        description:
+          type: string
+          description: Human-readable instruction text describing the post-install step.
+          minLength: 1
+          example: "Install as a system service for the web UI"
+        command:
+          type: string
+          description: Optional shell command the user can run to complete the step. Displayed for the user to copy; clients must not execute it automatically.
+          example: "mind-map service install --addr 127.0.0.1:4242"
+        documentation:
+          type: string
+          format: uri
+          description: Optional link to documentation or a setup page describing the step in more detail.
+          example: "https://github.com/aniongithub/mind-map#service-management"
+        optional:
+          type: boolean
+          description: Whether the step is optional. When true, the MCP server functions without completing this step.
+          default: false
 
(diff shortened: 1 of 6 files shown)


anneheartrecord:
@rdimitrov This adds an optional `postInstallInstructions` field to the package schema (types.go, JSON schema, OpenAPI) for #1340, with validator and model tests. CI is green — could you take a look?

PR #1411, opened 2026-07-01
fix(validators): distinguish a missing package from a missing version on PyPI/NPM

### LINKED ISSUE #553: Receive "title":"Bad Request","status":400,"detail":"Failed to publish server" for package not found
Running mcp-publisher publish in GitHub action after successful github-oidc login returns 
`Error: publish failed: server returned status 400: {"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"registry validation failed for package 0 (snowflake-labs-mcp): PyPI package 'snowflake-labs-mcp' not found (status: 404)"}]}`

However, [snowflake-labs-mcp](https://pypi.org/project/snowflake-labs-mcp/) exists in PyPI.

**Expected behavior**
Server should be successfully published.

**Logs**
Error: publish failed: server returned status 400: {"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"registry validation failed for package 0 (snowflake-labs-mcp): PyPI package 'snowflake-labs-mcp' not found (status: 404)"}]}

**Additional context**
README contains 
`<!-- mcp-name: io.github.Snowflake-Labs/mcp -->`.

server.json is attached.
[server.json](https://github.com/user-attachments/files/22547471/server.json)

Github Action workflow file attached.
[publish-mcp.yml](https://github.com/user-attachments/files/22547472/publish-mcp.yml)

sronix:
Fixes #553.

The PyPI and npm validators fetch the version-specific metadata endpoint (`/pypi/{name}/{version}/json` and the npm equivalent) and report any non-200 as `<pkg> not found`. That endpoint 404s for two different reasons though: the package doesn't exist, or it exists and only the *version* is missing (e.g. a release that hasn't propagated yet). So a valid package reads as gone:

```diff
- PyPI package 'requests' not found (status: 404)
+ PyPI package 'requests' exists, but version '99.99.99' was not found (status: 404). A newly published release can take a moment to appear on PyPI. Wait and retry, or publish version '99.99.99' before registering it
```

On a version 404 it probes the package-level endpoint (`/pypi/{name}/json`, `/{name}`) with a `HEAD` to tell the two apart, and reports 429/5xx as transient rather than "not found". `HEAD` so it reads a status without pulling the whole packument. The probe carries its own 3s deadline: it only refines the error message, so a hung probe must not stretch the validator past the ~10s-per-registry budget the publish path assumes. Same probe-and-classify shape as the existing cargo validator.

The fetch is split into `validatePyPIPackage` / `validateNPMPackage` behind `export_test.go` so the branches are testable with `httptest`, the same seam cargo uses.

### Testing
- [x] hermetic `httptest` for the status matrix (version-missing, package-missing, 5xx, 429, inconclusive probe, positive path, scoped npm); the mocks pin the expected method per endpoint (`GET` fetch, `HEAD` probe)
- [x] hermetic deadline test: a hung probe is cut off at ~3s and reported as inconclusive, not "not found"
- [x] existing live package tests still pass, now also hitting the HEAD probe; on a live 429/5xx they now `t.Skip` as inconclusive instead of flaking (deliberate: CI runs these against the real registries with no short-mode gating)
- [x] `gofmt` / `go vet` / `golangci-lint` clean

Out of scope: auto-retry for the propagation race, and SSRF redirect-pinning parity for the pypi/npm clients (pre-existing, shared with nuget).

### DIFF
--- a/internal/validators/registries/export_test.go
+++ b/internal/validators/registries/export_test.go
@@ -7,3 +7,14 @@ package registries
 //
 // Intended for cargo_test.go's positive-path and transient-error tests only.
 var ValidateCargoREADME = validateCargoREADME
+
+// ValidatePyPIPackage and ValidateNPMPackage expose the package-private
+// validatePyPIPackage / validateNPMPackage to the external _test package so
+// httptest-driven tests can exercise the metadata fetch and status-disambiguation
+// pipeline (missing package vs missing/unpropagated version vs transient upstream)
+// against a mock server, bypassing the exact-baseURL guard that the public
+// ValidatePyPI / ValidateNPM enforce.
+var (
+	ValidatePyPIPackage = validatePyPIPackage
+	ValidateNPMPackage  = validateNPMPackage
+)
--- a/internal/validators/registries/npm.go
+++ b/internal/validators/registries/npm.go
@@ -154,6 +154,11 @@ func npmVersion404Error(ctx context.Context, client *http.Client, pkg model.Pack
 // of version, with a HEAD request to the package-level endpoint (/{name}). Only
 // the status code is used, so HEAD avoids downloading the (large) packument.
 func probeNPMPackage(ctx context.Context, client *http.Client, baseURL, identifier string) npmPackageState {
+	// The probe only refines the 404 error message, so it must not extend the
+	// validator's worst case by another full client timeout.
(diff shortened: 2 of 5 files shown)


rdimitrov (approved):
hey, @sronix, thanks for the contribution! Much appreciated! 🙏

PR #1436, opened 2026-07-11
fix(api): allow PATCH in CORS so browsers can call the status endpoints

JosephDoUrden:
The status endpoints (PATCH /v0/servers/{name}/status and the per-version one) are served with PATCH, but the CORS middleware's AllowedMethods lists GET/POST/PUT/DELETE/OPTIONS and never PATCH. So a cross-origin browser preflight for those endpoints comes back without the allow headers and the browser blocks the actual request. One-line fix: add PATCH to the allowed methods.

While here, the existing CORS test turned out to be a placeholder that never reaches the real wrapped handler (its own comments say so), which is why this went unnoticed. Added a small Handler() accessor and a proper preflight test that asserts each routed method including PATCH is allowed.

### DIFF
--- a/internal/api/cors_preflight_test.go
+++ b/internal/api/cors_preflight_test.go
@@ -1,0 +1,62 @@
+package api_test
+
+import (
+	"crypto/ed25519"
+	"crypto/rand"
+	"encoding/hex"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/modelcontextprotocol/registry/internal/api"
+	v0 "github.com/modelcontextprotocol/registry/internal/api/handlers/v0"
+	"github.com/modelcontextprotocol/registry/internal/config"
+	"github.com/modelcontextprotocol/registry/internal/telemetry"
+)
+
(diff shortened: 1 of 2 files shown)


rdimitrov (approved):
Verified the fix end-to-end:

- Confirmed both status endpoints (`PATCH /v0/servers/{name}/versions/{version}/status` and the per-server variant) are registered with `http.MethodPatch` in `status.go`, while the CORS `AllowedMethods` list omitted PATCH — so cross-origin browser preflights for those endpoints came back without the allow header and got blocked.
- The omission was an oversight, not intentional: the CORS list included `DELETE` (which **no** route uses) but not `PATCH` (which two routes use) — the signature of a stale `GET/POST/PUT/DELETE/OPTIONS` boilerplate that wasn't updated when the PATCH endpoints were added.
- The pre-existing `cors_test.go` was a placeholder that asserts against a locally-built handler with no CORS middleware, so it could never have caught this.
- The new `TestCORSPreflightAllowedMethods` is a genuine guard: I reverted the one-line fix and the PATCH subtest fails with `Access-Control-Allow-Methods=""`; with the fix it passes for all routed methods.

Follow-up to remove the now-redundant placeholder `cors_test.go` incoming.

PR #1452, opened 2026-07-16
fix(validators): allow GitLab URLs with nested subgroups

### LINKED ISSUE #1359: GitLab repository URLs with nested groups/subgroups are rejected by validator

truecallerabreham:
## Summary

The gitlabURLRegex in internal/validators/utils.go only allows two path segments (owner/repo), rejecting valid GitLab URLs with subgroups like:

`
https://gitlab.com/myorg/team/subgroup/my-mcp-server
`

GitLab subgroups are a standard feature used in GitLab Enterprise. The current regex prevents self-hosted registries from publishing servers whose source lives in a GitLab group hierarchy.

## Changes

- Updated gitlabURLRegex to allow any number of path segments: [\w.-]+(/[\w.-]+)*
- Added unit tests for GitLab URLs with 1, 2, and 3 levels of subgroups
- Added comprehensive test suite for IsValidRepositoryURL function

## Test Plan

- [x] Added unit tests for subgroup URLs
- [ ] Run go test ./internal/validators/... (requires Go installation)

## Related

Fixes #1359

### DIFF
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -11,7 +11,7 @@ var (
 	// These regex patterns ensure the URL is in the format of a valid GitHub or GitLab repository
 	// For example:	// - GitHub: https://github.com/user/repo
 	githubURLRegex = regexp.MustCompile(`^https?://(www\.)?github\.com/[\w.-]+/[\w.-]+/?$`)
-	gitlabURLRegex = regexp.MustCompile(`^https?://(www\.)?gitlab\.com/[\w.-]+/[\w.-]+/?$`)
+	gitlabURLRegex = regexp.MustCompile(`^https?://(www\.)?gitlab\.com/[\w.-]+(/[\w.-]+)*/?$`)
 )
 
 // IsValidRepositoryURL checks if the given URL is valid for the specified repository source
--- a/internal/validators/utils_test.go
+++ b/internal/validators/utils_test.go
@@ -1,0 +1,113 @@
+package validators
+
+import (
+	"testing"
+)
+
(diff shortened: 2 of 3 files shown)


nbeai:
Thanks for adding subgroup coverage. One thing I think may fail in the new table test: the updated GitLab regex now allows a single path segment (`gitlab.com/owner`), because `(/[\w.-]+)*` can repeat zero times. That seems to conflict with the new `GitLab URL missing repo` case expecting `false`.

Maybe the pattern should still require at least two path segments, then allow extra subgroup segments after that, e.g. `.../[\w.-]+/[\w.-]+(/[\w.-]+)*/?$`.

rdimitrov:
Thanks for taking this on! We ended up with two PRs for the same GitLab nested-subgroup fix (this one and #1361). #1361 is passing CI and its regex requires at least owner+repo, whereas the pattern here also matches a single-segment path like `gitlab.com/owner` (which is what's turning the tests red). To keep things simple I would prefer we work with #1361 and close this one.

PR #1466, opened 2026-07-20
fix(validators): reject loopback/private/link-local hosts in IsValidRemoteURL

rajanpanth:
## Description
`IsValidRemoteURL` only rejected the literal `localhost`, `127.0.0.1`, and `*.localhost`, so IP forms of local/internal addresses passed validation: `[::1]`, the rest of `127.0.0.0/8` (e.g. `127.0.0.2`), `0.0.0.0`, `[::]`, IPv4-mapped loopback (`[::ffff:127.0.0.1]`), and RFC1918/link-local (e.g. `169.254.169.254`).

This parses the hostname with `net.ParseIP` and rejects loopback, unspecified, private, and link-local addresses, completing the intended "no localhost" restriction and aligning it with the address filtering already used by the auth dialer and `validateRealmURL`.

## Motivation and Context
The check compared string literals, so only three exact hosts were blocked. A publisher could register a remote URL pointing at loopback/internal addresses despite the stated restriction. This is also inconsistent with the codebase's own handling elsewhere (the auth domain-verification dialer and `validateRealmURL` already block these ranges).

## How Has This Been Tested?
Added a table-driven unit test (`TestIsValidRemoteURL`) covering the newly-rejected forms, regression guards for the already-blocked cases, and still-allowed public hosts. `go test ./internal/validators/...` passes; `gofmt` and `go vet` are clean.

## Breaking Changes
None. Public hostnames and template-variable URLs are unaffected.

## Types of changes

## Checklist
- [x] I have read the MCP Documentation

### DIFF
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -1,6 +1,7 @@
 package validators
 
 import (
+	"net"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


rdimitrov:
Thanks for this! 🙏 It turns out #1470 tackles the same loopback/private-host fix and is a little further along (it extracts a shared helper and has broader test coverage), so we're going to go with that one to avoid double-merging.

PR #1470, opened 2026-07-21
fix(validators): reject loopback/private/link-local hosts in IsValidRemoteURL

### LINKED ISSUE #1465: IsValidRemoteURL only blocks literal localhost/127.0.0.1 — misses [::1], 127.0.0.0/8, 0.0.0.0, and private/link-local addresses
IsValidRemoteURL (internal/validators/utils.go) is meant to reject remote URLs pointing at local/internal hosts ("stricter than packages - no localhost allowed"). The current host check is:

hostname := u.Hostname()
if hostname == "localhost" || hostname == "127.0.0.1" || strings.HasSuffix(hostname, ".localhost") {
    return false
}
Because it compares string literals, it only catches localhost, 127.0.0.1, and *.localhost. These all currently pass validation:

IPv6 loopback: https://[::1]/
Rest of 127.0.0.0/8: https://127.0.0.2/
Unspecified: https://0.0.0.0/, https://[::]/
IPv4-mapped loopback: https://[::ffff:127.0.0.1]/
RFC1918 / link-local: https://10.0.0.1/, https://192.168.1.1/, https://169.254.169.254/
So a published server entry can register a remote URL pointing at loopback/internal addresses despite the intended restriction. This is also inconsistent with the codebase's own handling elsewhere (the auth domain-verification dialer already blocks these ranges, and validateRealmURL blocks 0.0.0.0/::).

Happy to send a PR that resolves the host with net.ParseIP and rejects loopback/unspecified/private/link-local addresses.

amitvijapur:
Fixes #1465

## Cause

`IsValidRemoteURL` rejected only the literal strings `localhost` / `127.0.0.1` / `*.localhost`, so every other notation for a loopback or internal host passed validation: `[::1]`, other `127.0.0.0/8` addresses (`127.0.0.2`), unspecified addresses (`0.0.0.0`, `[::]`), IPv4-mapped IPv6 forms (`[::ffff:127.0.0.1]`), and RFC1918/link-local addresses.

## Change

Extracted the host check into `isDisallowedRemoteHost`: if the hostname parses as an IP (`net.ParseIP` — this includes bracketed IPv6 forms, which `url.URL.Hostname()` already strips, and IPv4-mapped forms), reject `IsLoopback() || IsUnspecified() || IsPrivate() || IsLinkLocalUnicast()`. Hostnames that don't parse as an IP keep the pre-existing `localhost` / `*.localhost` string checks, since resolving DNS names would require a network round trip during validation.

## Tests

New table-driven `TestIsValidRemoteURL_LoopbackAndPrivateAddresses` in `internal/validators/utils_test.go` (20 subtests) covering every bypass listed in the issue plus sanity checks that public hosts still validate. `go build ./...`, `go vet ./...`, and `gofmt` all clean; `golangci-lint run` reports no new findings versus `main` on the touched package.

### DIFF
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -2,6 +2,7 @@ package validators
 
 import (
 	"net"
+	"net/netip"
 	"net/url"
 	"regexp"
 	"strings"
@@ -162,8 +163,12 @@ func IsValidRemoteURL(rawURL string) bool {
 	return true
 }
 
+// cgnat is RFC 6598 shared address space. It is not routable on the public
+// internet, but net.IP.IsPrivate covers only RFC1918 and fc00::/7.
+var cgnat = netip.MustParsePrefix("100.64.0.0/10")
+
 // isDisallowedRemoteHost reports whether hostname is not allowed as a remote
-// URL host because it is loopback, unspecified, private, or link-local.
+// URL host because it is not a publicly reachable unicast address.
 //
 // hostname may be an IP literal - including bracketed IPv6 forms already
 // stripped of their brackets by url.URL.Hostname (e.g. "::1"), and
@@ -175,8 +180,17 @@ func IsValidRemoteURL(rawURL string) bool {
 // checks, since resolving them here would require a network round trip
 // during validation.
 func isDisallowedRemoteHost(hostname string) bool {
+	// DNS is case-insensitive and a trailing root dot is equivalent to its
+	// absence, so "LOCALHOST" and "localhost." both reach loopback.
+	hostname = strings.ToLower(strings.TrimSuffix(hostname, "."))
+
 	if ip := net.ParseIP(hostname); ip != nil {
-		return ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() || ip.IsLinkLocalUnicast()
+		if ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() ||
+			ip.IsLinkLocalUnicast() || ip.IsMulticast() || ip.Equal(net.IPv4bcast) {
+			return true
+		}
+		addr, ok := netip.AddrFromSlice(ip)
+		return ok && cgnat.Contains(addr.Unmap())
 	}
 
 	return hostname == "localhost" || strings.HasSuffix(hostname, ".localhost")
--- a/internal/validators/utils.go
+++ b/internal/validators/utils.go
@@ -1,6 +1,7 @@
 package validators
 
 import (
+	"net"
 	"net/url"
 	"regexp"
 	"strings"
@@ -147,9 +148,10 @@ func IsValidRemoteURL(rawURL string) bool {
 		return false
 	}
 
-	// Reject localhost URLs for remotes (security/production concerns)
-	hostname := u.Hostname()
-	if hostname == "localhost" || hostname == "127.0.0.1" || strings.HasSuffix(hostname, ".localhost") {
+	// Reject localhost/loopback/unspecified/private/link-local hosts for remotes
+	// (security/production concerns - a remote must point at a real, publicly
+	// reachable, non-internal endpoint).
+	if isDisallowedRemoteHost(u.Hostname()) {
 		return false
 	}
 
@@ -158,6 +160,26 @@ func IsValidRemoteURL(rawURL string) bool {
 	}
 
 	return true
+}
+
+// isDisallowedRemoteHost reports whether hostname is not allowed as a remote
+// URL host because it is loopback, unspecified, private, or link-local.
+//
+// hostname may be an IP literal - including bracketed IPv6 forms already
+// stripped of their brackets by url.URL.Hostname (e.g. "::1"), and
+// IPv4-mapped IPv6 forms (e.g. "::ffff:127.0.0.1") - or a DNS name. IP
+// literals are checked with net.ParseIP and net.IP's classification methods
+// so every notation for loopback/private/link-local addresses is caught, not
+// just the literal strings "127.0.0.1" and "::1" the previous check used.
+// DNS names fall back to the pre-existing "localhost" / "*.localhost" string
+// checks, since resolving them here would require a network round trip
+// during validation.
+func isDisallowedRemoteHost(hostname string) bool {
+	if ip := net.ParseIP(hostname); ip != nil {
+		return ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() || ip.IsLinkLocalUnicast()
+	}
+
+	return hostname == "localhost" || strings.HasSuffix(hostname, ".localhost")
 }
 
 // IsValidTemplatedURL validates a URL with template variables against available variables
--- a/internal/validators/utils_test.go
+++ b/internal/validators/utils_test.go
@@ -1,0 +1,56 @@
+package validators_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/modelcontextprotocol/registry/internal/validators"
+)
+
+// TestIsValidRemoteURL_LoopbackAndPrivateAddresses exercises IsValidRemoteURL
+// directly against every bypass listed in the GitHub issue: the previous
+// implementation only rejected the literal strings "localhost", "127.0.0.1",
+// and "*.localhost", so IPv6 loopback, the rest of 127.0.0.0/8, unspecified
+// addresses, IPv4-mapped loopback, and RFC1918/link-local addresses all
+// passed validation despite the "no localhost allowed" contract.
+func TestIsValidRemoteURL_LoopbackAndPrivateAddresses(t *testing.T) {
+	tests := []struct {
+		name  string
+		url   string
+		valid bool
+	}{
+		// Previously-caught cases - must keep working
+		{name: "literal localhost", url: "https://localhost/", valid: false},
+		{name: "literal localhost with port", url: "https://localhost:8443/mcp", valid: false},
+		{name: "localhost subdomain", url: "https://foo.localhost/", valid: false},
+		{name: "literal 127.0.0.1", url: "https://127.0.0.1/", valid: false},
+
+		// Previously-missed bypasses called out in the issue
+		{name: "IPv6 loopback", url: "https://[::1]/", valid: false},
+		{name: "rest of 127.0.0.0/8", url: "https://127.0.0.2/", valid: false},
+		{name: "127.0.0.0/8 upper range", url: "https://127.255.255.254/", valid: false},
+		{name: "IPv4 unspecified", url: "https://0.0.0.0/", valid: false},
+		{name: "IPv6 unspecified", url: "https://[::]/", valid: false},
+		{name: "IPv4-mapped IPv6 loopback", url: "https://[::ffff:127.0.0.1]/", valid: false},
+		{name: "RFC1918 10.0.0.0/8", url: "https://10.0.0.1/", valid: false},
+		{name: "RFC1918 172.16.0.0/12", url: "https://172.16.0.1/", valid: false},
+		{name: "RFC1918 192.168.0.0/16", url: "https://192.168.1.1/", valid: false},
+		{name: "IPv4 link-local (cloud metadata endpoint)", url: "https://169.254.169.254/", valid: false},
+		{name: "IPv6 unique local address", url: "https://[fc00::1]/", valid: false},
+		{name: "IPv6 link-local unicast", url: "https://[fe80::1]/", valid: false},
+
+		// Sanity checks - must not overblock legitimate remotes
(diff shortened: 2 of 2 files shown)


amitvijapur:
Bumping this one, which has been quiet since 21 July.

All three checks are green and it still merges cleanly with `main`. It is small and self-contained: one commit, 81 lines, with table tests covering the loopback, private and link-local cases.

Happy to rebase, split it, or narrow the scope if any of that would make review easier.

UgaTheDev:
Nice catch, and the `net.ParseIP` + `net.IP` classification approach is the right
shape — it replaces a string allowlist that could only ever cover the exact
spellings someone thought to enumerate. Confirming it works, then offering some
additional table entries for hosts that still get through.

Verified on `5b77ced`:

```
$ go test ./internal/validators/ -run TestIsValidRemoteURL -v
--- PASS: TestIsValidRemoteURL_LoopbackAndPrivateAddresses (0.00s)
    ... 20/20 subtests PASS
ok  	github.com/modelcontextprotocol/registry/internal/validators	0.776s

$ go test ./internal/validators/          # whole package
ok  	github.com/modelcontextprotocol/registry/internal/validators	8.317s
```

The alternate-notation coverage is genuinely better than the old check. Both of
these were allowed before and are correctly rejected now, which is exactly the
class of bug the string comparison could not catch:

```
https://[0:0:0:0:0:ffff:7f00:1]/   ParseIP -> 127.0.0.1  IsLoopback()=true  REJECTED
https://[::ffff:10.0.0.1]/         ParseIP -> 10.0.0.1    IsPrivate()=true   REJECTED
```

## Hosts that still reach loopback after this change

I probed `IsValidRemoteURL` on the PR head and then checked what the platform
resolver actually does with each host. Three groups still pass validation. All
numbers below are from runs on `5b77ced`.

### 1. Case and trailing-dot variants of `localhost` (highest impact)

The DNS fallback is still an exact, case-sensitive string comparison, so it
misses spellings that resolve to loopback anyway. DNS is case-insensitive and a
trailing root dot is equivalent to its absence:

```
https://LOCALHOST/         host="LOCALHOST"      ALLOWED
https://Foo.LOCALHOST/     host="Foo.LOCALHOST"  ALLOWED
https://localhost./        host="localhost."     ALLOWED
```

And they do resolve — `net.LookupHost` on this machine:

```
LOCALHOST        -> [::1 127.0.0.1]
LoCaLhOsT        -> [::1 127.0.0.1]
localhost.       -> [::1 127.0.0.1]
foo.LOCALHOST    -> [::1 127.0.0.1]
```

Worth noting this one predates the PR — it is a gap in the check being carried
forward, not something introduced here. I mention it because it is the cheapest
to close (one `strings.ToLower` + `TrimSuffix`) and it is the variant most likely
to be typed by accident rather than crafted. There is no case normalization
anywhere upstream of this call; `grep -rn "ToLower" internal/validators/*.go`
returns nothing, and `url.Parse` preserves host case.

### 2. Non-canonical IPv4 literals that `net.ParseIP` rejects but resolvers accept

`net.ParseIP` is deliberately strict (it has rejected leading zeros since Go
1.17), so these fall through to the DNS branch, fail the `localhost` string
comparison, and are allowed:

```
https://2130706433/    host="2130706433"    ParseIP=nil   ALLOWED
https://127.1/         host="127.1"         ParseIP=nil   ALLOWED
https://0x7f.0.0.1/    host="0x7f.0.0.1"    ParseIP=nil   ALLOWED
https://0177.0.0.1/    host="0177.0.0.1"    ParseIP=nil   ALLOWED
```

Go's own resolver then maps three of the four straight to loopback:

```
2130706433   -> [127.0.0.1]
127.1        -> [127.0.0.1]
0x7f.0.0.1   -> [127.0.0.1]
0177.0.0.1   -> [177.0.0.1]     <- not loopback under Go's resolver
```

One correction to the obvious assumption here: the octal form `0177.0.0.1` does
**not** reach loopback via Go — Go resolves it to `177.0.0.1`. It does reach
`127.0.0.1` under glibc/curl, which parse the leading zero as octal. So its
severity depends on which client dials the URL, whereas the decimal, short, and
hex forms reach loopback even under Go's own strict-ish resolver.

### 3. CGNAT `100.64.0.0/10`

Confirmed — `net.IP.IsPrivate()` covers only RFC 1918 and `fc00::/7`, so shared
address space is not caught:

```
https://100.64.0.1/        ParseIP=100.64.0.1       IsPrivate()=false   ALLOWED
https://100.127.255.255/   ParseIP=100.127.255.255  IsPrivate()=false   ALLOWED
```

Same story for multicast and broadcast, if you want them covered while you are in
here: `224.0.0.1`, `255.255.255.255`, and `[ff02::1]` are all currently allowed.
`fe80::1%eth0` (zone ID) also slips through, since `net.ParseIP` rejects the
`%zone` suffix.

## On severity, so this is calibrated fairly

I traced the caller — `validateRemoteTransport` at `internal/validators/validators.go:594`
— and the registry server itself never dials these URLs; the only outbound HTTP
in the tree is under `cmd/publisher/`. So this is not server-side SSRF. The
consequence is that published registry data can point MCP *clients* at their own
loopback or their cloud metadata endpoint, which is still worth closing but is a
different (and lower) severity than an SSRF in the registry. Framing the PR that
way in the description might actually help it get merged faster.

## Suggested additions

Table entries for the existing test, all of which fail on `5b77ced` today:

```go
// Case and trailing-dot variants - DNS is case-insensitive and the root dot
// is equivalent to its absence, so these all resolve to loopback.
{name: "uppercase localhost", url: "https://LOCALHOST/", valid: false},
{name: "mixed-case localhost subdomain", url: "https://Foo.LOCALHOST/", valid: false},
{name: "localhost with trailing root dot", url: "https://localhost./", valid: false},

// Non-canonical IPv4 literals: net.ParseIP rejects these, but the platform
// resolver maps them to 127.0.0.1.
{name: "decimal IPv4 literal", url: "https://2130706433/", valid: false},
{name: "short-form IPv4 literal", url: "https://127.1/", valid: false},
{name: "hex IPv4 literal", url: "https://0x7f.0.0.1/", valid: false},
{name: "octal IPv4 literal", url: "https://0177.0.0.1/", valid: false},

// Shared address space - IsPrivate() covers only RFC1918 and fc00::/7.
{name: "CGNAT 100.64.0.0/10", url: "https://100.64.0.1/", valid: false},
{name: "CGNAT upper range", url: "https://100.127.255.255/", valid: false},

// Non-unicast
{name: "IPv4 multicast", url: "https://224.0.0.1/", valid: false},
{name: "IPv4 broadcast", url: "https://255.255.255.255/", valid: false},
{name: "IPv6 link-local multicast", url: "https://[ff02::1]/", valid: false},
```

And a version of `isDisallowedRemoteHost` that passes all of them. I ran this
against 28 cases (the 20 above plus the PR's existing ones, plus negative
controls `example.com`, `8.8.8.8`, `2001:4860:4860::8888`, `mcp.example.co.uk`,
`100.63.255.255`, `100.128.0.1`, `1.1.1.1`, `my-server.internal.example.com`)
and all 28 pass:

```go
var cgnat = netip.MustParsePrefix("100.64.0.0/10")

func isDisallowedRemoteHost(hostname string) bool {
	// DNS is case-insensitive and a trailing root dot is equivalent to its absence.
	hostname = strings.ToLower(strings.TrimSuffix(hostname, "."))

	if ip := net.ParseIP(hostname); ip != nil {
		if ip.IsLoopback() || ip.IsUnspecified() || ip.IsPrivate() ||
			ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast() || ip.IsMulticast() {
			return true
		}
		if addr, ok := netip.AddrFromSlice(ip); ok {
			return cgnat.Contains(addr.Unmap())
		}
		return false
	}

	if hostname == "localhost" || strings.HasSuffix(hostname, ".localhost") {
		return true
	}

	// Not a valid IP and not a plausible DNS name. Forms like "2130706433",
	// "127.1" and "0x7f.0.0.1" are rejected by net.ParseIP but still resolved
	// to 127.0.0.1 by the platform resolver. A real DNS name's final label
	// never starts with a digit.
	last := hostname[strings.LastIndex(hostname, ".")+1:]
	return last == "" || (last[0] >= '0' && last[0] <= '9')
}
```

The last rule is the one I would push back on my own suggestion for: it rejects
by shape rather than by resolution, so it is a heuristic. It does hold for the
numeric-literal cases without touching any legitimate hostname I could think of,
and refusing to publish a remote whose host is not a resolvable-looking DNS name
or a canonical IP seems like the behaviour you want regardless. But if you would
rather keep the diff to what you can state precisely, dropping that last block
and taking only the `ToLower`/`TrimSuffix` normalization plus the CGNAT and
multicast checks still closes most of this.

None of this needs to block the PR — it is strictly better than what is on main
today. Happy to send the additional cases as a follow-up if you would rather
merge this as-is.

JosephDoUrden (commented):
Independent run on `5b77ced`:

- `go test ./internal/validators/` passes on the PR head.
- Merged current main (`6036804`) into it: 30 commits behind, but the merge is clean and tests stay green.
- Mutation check: dropping `ip.IsPrivate()` or the `.localhost` suffix fallback each breaks the table tests, so the suite is load-bearing.

Agree with the comment above on the residual gaps (case/trailing-dot localhost, non-canonical IPv4 literals, CGNAT) and on not letting them block this: they pre-date the PR, and swapping the exact-string allowlist for real IP classification already closes the whole alternate-notation class. Follow-up territory.

Looks merge-ready to me.

amitvijapur:
Thanks both, and sorry for the slow turnaround.

@UgaTheDev I took the parts of your suggestion I can state precisely, in `1d23a92`:

- `strings.ToLower` + `TrimSuffix(hostname, ".")` before classification, so `LOCALHOST`, `Foo.LOCALHOST` and `localhost.` are caught.
- CGNAT `100.64.0.0/10` via `netip.Prefix.Contains`, since `IsPrivate` covers only RFC1918 and fc00::/7.
- `IsMulticast` and the limited broadcast address.

One correction on your snippet: it does not actually pass the `255.255.255.255` case. `net.IP` has no broadcast classifier and the limited broadcast address is not multicast, so it parses cleanly and falls through the numeric-shape rule (its last label starts with a digit only if you reach that branch, which an IP literal never does). I added an explicit `ip.Equal(net.IPv4bcast)`.

I left out the numeric-shape heuristic. You flagged it yourself as rejecting by shape rather than resolution, and I would rather not have this PR assert something about `127.1` and `0x7f.0.0.1` that the code cannot actually verify. Those remain open, and they were open before this PR too. Happy to look at them separately if a maintainer wants that closed here.

Mutation-checked the additions: removing any one of `ToLower`, `TrimSuffix`, the CGNAT check, `IsMulticast`, or the broadcast check fails the table, so none of the new cases are decorative. `go test ./internal/validators/` and `go build ./...` pass on `1d23a92`.

Disclosure: this was written with AI assistance and reviewed by me before pushing.

JosephDoUrden:
Re-ran on `1d23a92`. `go test ./internal/validators/...` green in both packages, `go build ./...` clean, still merges clean onto `6036804` which is where main still sits.

Checked @amitvijapur's mutation claim myself. Dropping any one of `ToLower`, `TrimSuffix`, the CGNAT check, `IsMulticast` or the `IPv4bcast` compare fails the table, so all five are load bearing.

@UgaTheDev the broadcast correction holds. `::ffff:255.255.255.255` is not multicast, and with the `IPv4bcast` compare removed `255.255.255.255` is the one case nothing else catches.

Also poked the IPv4-mapped forms. `[::ffff:100.64.0.1]`, `[::ffff:224.0.0.1]` and `[::ffff:255.255.255.255]` all rejected, `[::ffff:8.8.8.8]` still allowed. The `Unmap` is only carrying the CGNAT check there, `IsMulticast` and `Equal` handle mapped addresses on their own. It is needed for plain `100.64.0.1` too, since `ParseIP` hands back 16 bytes.

The zone-id form `[fe80::1%25eth0]` still gets through, same bucket as the numeric literals, not blocking. PR body still says 20 subtests btw, it is 30 now.

Agree with leaving the numeric-shape heuristic out.

Still merge-ready from my side.

UgaTheDev:
Confirmed on 1d23a92. Ran `go test ./internal/validators/... -run IsValidRemoteURL` (30 subtests pass) and `go build ./...` (clean). The additions for case normalization, CGNAT 100.64.0.0/10, multicast, and limited broadcast are load-bearing per the mutation check.

Your correction on the broadcast case is right, and I should have caught it: `255.255.255.255` parses cleanly, `IsMulticast` does not cover it, and the numeric-shape rule in my snippet never runs for an IP literal, so it would have passed straight through. The explicit `ip.Equal(net.IPv4bcast)` is the correct fix. I also agree with leaving the shape heuristic out; it asserted more about `127.1` and `0x7f.0.0.1` than the code could verify, and that belongs in its own change if a maintainer wants it.

Nothing further from me.

PR #1471, opened 2026-07-21
feat(search): match server description as well as name

### LINKED ISSUE #1453: feat: search should also match against server description field
## Problem

The `?search=` parameter on `/v0/servers` only matches against `server_name` (via ILIKE substring match). It does **not** search the `description` field.

This was explicitly requested in #135 (now closed) but only the name-search portion was implemented.

## Impact

AI agents and users searching for servers by what they **do** (e.g., "weather", "database", "GitHub integration") cannot find relevant servers unless they already know the exact server name. The description field — which contains the most useful discovery information — is completely ignored by search.

## Current behavior

```
GET /v0/servers?search=weather
```

Returns only servers where `server_name` contains "weather". A server named `io.github.example/data-tools` with description "Weather data and forecasts" is **not** returned.

## Expected behavior

Search should match against both `server_name` AND `description` fields, with name matches ranked first for relevance.

## Proposed implementation

1. Add `SubstringDescription` filter field to `ServerFilter`
2. Add ILIKE condition on the `description` column in `buildFilterConditions`
3. Update the API handler to pass the search term to both filters
4. Name matches should appear before description matches (relevance ranking)

## Files affected

- `internal/database/database.go` — add `SubstringDescription` to `ServerFilter`
- `internal/database/postgres.go` — add ILIKE condition for description in `buildFilterConditions`
- `internal/api/handlers/v0/servers.go` — pass search term to both name and description filters
- `internal/api/handlers/v0/servers_test.go` — add tests for description search
- `internal/database/postgres_test.go` — add tests for description filter

## Context

This is particularly important for AI-agent workflows where the agent searches by capability keywords rather than knowing server names in advance.

amitvijapur:
Fixes #1453

## Change

Adds `SubstringDescription` to `ServerFilter`; the `?search=` handler now sets it alongside `SubstringName` from the same term, so search matches server name **or** description.

Two implementation notes where this deviates from the issue's literal proposal, deliberately:

- **OR, not AND**: adding two independent filter fields would have let `buildFilterConditions`' outer AND-join require a match on *both* name and description. When both fields are set, they're combined into a single `(server_name ILIKE ... OR value ->> 'description' ILIKE ...)` condition instead. A white-box test asserts the generated SQL to lock this in.
- **JSONB access**: `description` has no dedicated column (unlike `server_name`, materialized in migration 009) — it lives inside the JSONB `value` blob, so the condition reads `value ->> 'description'`.

The existing LIKE-metacharacter escaping is preserved via a shared `likePattern` helper (same escaping order, same `ESCAPE '\'` pairing), so `?search=%` still can't wildcard-match everything through the new description path either.

**Scoped out**: the issue also suggests ranking name matches above description matches. `ListServers` uses keyset/cursor pagination ordered by `(server_name, version)`, and reordering by match-type would break that pagination scheme, so relevance ranking is left out of this PR rather than silently half-done. Happy to explore it separately if there's an agreed approach.

## Tests

- New white-box `TestBuildFilterConditions_SearchMatchesNameOrDescription` (6 subtests, no DB required) asserting the generated SQL/args, including the OR-not-AND behavior.
- New DB-backed cases in `postgres_test.go` and `servers_test.go` (description-only match, combined search, no-match). Honest note: these compile cleanly (`go test -run '^$'`) but I could not execute them locally because port 5432 is occupied by an unrelated service on my machine and the test harness's connection string is fixed to `localhost:5432` — CI's Postgres job will exercise them.
- `go build ./...`, `go vet ./...`, `gofmt` clean; `golangci-lint run` reports no new findings versus `main` on the touched packages.

### DIFF
--- a/internal/api/handlers/v0/servers.go
+++ b/internal/api/handlers/v0/servers.go
@@ -58,7 +58,7 @@ type ListServersInput struct {
 	Cursor         string       `query:"cursor" doc:"Pagination cursor" required:"false" example:"server-cursor-123"`
 	Limit          int          `query:"limit" doc:"Number of items per page" default:"30" minimum:"1" maximum:"100" example:"50"`
 	UpdatedSince   string       `query:"updated_since" doc:"Filter servers updated since timestamp (RFC3339 datetime)" required:"false" example:"2025-08-07T13:15:04.280Z"`
-	Search         string       `query:"search" doc:"Search servers by name (substring match)" required:"false" example:"filesystem"`
+	Search         string       `query:"search" doc:"Search servers by name or description (substring match)" required:"false" example:"filesystem"`
 	Version        string       `query:"version" doc:"Filter by version ('latest' for latest version, or an exact version like '1.2.3')" required:"false" example:"latest"`
 	IncludeDeleted OptionalBool `query:"include_deleted" doc:"Include deleted servers in results (default: false, but always true when updated_since is provided)" required:"false"`
 }
@@ -105,9 +105,10 @@ func RegisterServersEndpoints(api huma.API, pathPrefix string, registry service.
 			}
 		}
 
-		// Handle search parameter
+		// Handle search parameter - matches against both name and description
 		if input.Search != "" {
 			filter.SubstringName = &input.Search
+			filter.SubstringDescription = &input.Search
 		}
 
 		// Handle version parameter
--- a/internal/api/handlers/v0/servers_test.go
+++ b/internal/api/handlers/v0/servers_test.go
@@ -36,7 +36,7 @@ func TestListServersEndpoint(t *testing.T) {
 	_, err = registryService.CreateServer(ctx, &apiv0.ServerJSON{
 		Schema:      model.CurrentSchemaURL,
 		Name:        "com.example/server-beta",
-		Description: "Beta test server",
+		Description: "Beta test server for weather forecasting",
 		Version:     "2.0.0",
 	})
 	require.NoError(t, err)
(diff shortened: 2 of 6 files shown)


amitvijapur:
Bumping this one, which has been quiet since 21 July.

All three checks are green and it still merges cleanly with `main`. It is a single commit, and result ranking is deliberately left out with the reasoning in the description, so the change stays limited to matching the description alongside the name.

Happy to rebase or narrow it further if that would help review.

Hronom:
A fresh downstream data point after publishing a package-only MCPB entry today:

- `GET /v0.1/servers?search=hronaut&version=latest&limit=20` returns `io.github.hronaut/hronaut` `2.4.27`.
- The published description is `Control visible, persistent Hronaut browser workspaces through a local MCP connection.`
- `GET /v0.1/servers?search=browser&version=latest&limit=100` returns 100 entries but not Hronaut.

This is already affecting agent-side discovery rather than only a human search box. AgentResolver currently maps a persistent-browser-workspace request to the safe term `browser`, queries the official Registry with that term (up to five results), and merges those results into procurement candidates. I reran the same public request after Registry publication: Hronaut was still absent and the URL-to-Markdown result remained selected.

So this PR would fix a concrete first-stage recall failure: the server cannot be evaluated or ranked when the capability word exists only in `description`. It would not by itself solve relevance ranking, semantic capability verification, or handling package-only entries with no remote URL; those are separate downstream concerns.

Reproducible context: the exact Registry record is `io.github.hronaut/hronaut/versions/2.4.27`, and the downstream field-test thread is https://github.com/waxsway/agentresolver/issues/617#issuecomment-5745364220.

Disclosure: I maintain Hronaut. This is a measured search/discovery result, not a request for endorsement or an integration claim. Prepared with AI assistance.

PR #1504, opened 2026-08-05
docs: remove stale build.sh reference from publisher Key Files

latent-9:
The Key Files list in `cmd/publisher/README.md` includes a bullet for `build.sh`, but there is no `build.sh` anywhere in the repo. The publisher is built with `make publisher`, which runs `go build -o bin/mcp-publisher ./cmd/publisher` (Makefile), so there is no separate cross-platform build script.

The other three entries in the same list (`main.go`, `commands/`, `auth/`) all exist. This removes the stale bullet so the list matches the actual files in the directory.


rdimitrov (approved):

latent-9:
shanksyou

PR #1528, opened 2026-08-11
fix(publisher): accept UTF-8 BOM in server.json

beeswaxpat:
On Windows, PowerShell 5.1's `Out-File -Encoding utf8` (and several editors) write UTF-8 files with a leading byte order mark. Go's `encoding/json` does not skip the BOM, so `mcp-publisher publish` and `mcp-publisher validate` fail on such server.json files with the cryptic error `invalid character 'ï' looking for beginning of value`. RFC 8259 section 8.1 says implementations may ignore a leading BOM, and we hit this in practice publishing a server from a Windows machine.

This change strips a leading UTF-8 BOM (EF BB BF) from server.json bytes right after the file is read in both the publish and validate commands, using a small shared helper next to the existing UTF-8 validation in unicode.go. Both downstream request paths receive the cleaned bytes, so behavior is otherwise unchanged. Adds two tests that run a BOM-prefixed server.json through the full publish and validate paths.

### DIFF
--- a/cmd/publisher/commands/publish.go
+++ b/cmd/publisher/commands/publish.go
@@ -28,6 +28,7 @@ func PublishCommand(args []string) error {
 		}
 		return fmt.Errorf("failed to read server.json: %w", err)
 	}
+	serverData = stripUTF8BOM(serverData)
 	if err := validateJSONUnicode(serverFile, serverData); err != nil {
 		return err
 	}
--- a/cmd/publisher/commands/publish_test.go
+++ b/cmd/publisher/commands/publish_test.go
@@ -81,6 +81,20 @@ func TestPublishCommand_RejectsInvalidUTF8ServerJSON(t *testing.T) {
 	assert.Contains(t, err.Error(), "UTF-8")
 }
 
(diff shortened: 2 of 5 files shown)


JosephDoUrden (commented):
Reproduced the failure first. server.json with an EF BB BF prefix on the merge base dies with invalid character 'ï' looking for beginning of value, both publish and validate, exactly as described. Same file against this branch, both pass.

Checked the placement, it's TrimPrefix before Unmarshal, not a lenient parser. And I grepped every server.json reader in cmd/publisher, publish.go and validate.go are the only two real entry points and both get the strip, the api paths receive already stripped bytes. So this isn't a one-entry-point fix. Interior BOMs are preserved which is right, prefix only.

Also poked utf-16. With BOM you get a clean must-be-UTF-8 error, without BOM it dies the same way it does on main, nothing regressed.

No line in docs/ forbids a leading BOM and rfc 8259 8.1 says a parser may ignore it, so this is on solid ground. No overlap with #1583, #1588 or #1290. Lint diff vs base is zero.

Looks merge-ready to me.

PR #1544, opened 2026-08-18
fix(validators): name the package version in ownership-failure errors

### LINKED ISSUE #1525: Publish rejects correct mcpName as stale/mismatched — npm registry confirms it's correct
GitHub username: lelu-auth
Organization: Lelu-ai
Namespace: io.github.Lelu-ai/lelu-mcp
Repository: https://github.com/Lelu-ai/lelu
NPM package: lelu-mcp

Environment: mcp-publisher 1.8.1

## What happened

`mcp-publisher publish` consistently rejects the server with a 400, claiming the published npm package's `mcpName` doesn't match `server.json`'s `name` — but it does, verified directly against the npm registry API multiple times over several hours (including across a full session restart).

```
Error: publish failed: server returned status 400: {"title":"Bad Request","status":400,"detail":"Failed to publish server","errors":[{"message":"registry validation failed for package 0 (lelu-mcp): NPM package ownership validation failed. Expected mcpName 'io.github.Lelu-ai/lelu-mcp', got 'io.github.lelu-ai/lelu-mcp'"}]}
```

Note the casing: expected `io.github.Lelu-ai` (capital L, the real org name), but the registry reports it "got" `io.github.lelu-ai` (lowercase) — even though the npm package has never had the lowercase value published.

## Verification that npm's own data is correct

```
$ curl -s https://registry.npmjs.org/lelu-mcp/0.0.34 | jq .mcpName
"io.github.Lelu-ai/lelu-mcp"

$ curl -s https://registry.npmjs.org/lelu-mcp | jq '.["dist-tags"].latest, .versions[.["dist-tags"].latest].mcpName'
"0.0.34"
"io.github.Lelu-ai/lelu-mcp"
```

`0.0.34` was published with the correct capital-L `mcpName` from the start (we caught and fixed a casing mismatch between our org name `Lelu-ai` and an earlier lowercase `lelu-ai` namespace before ever publishing this version). There is no version of `lelu-mcp` on npm that has ever published the lowercase `io.github.lelu-ai` value as `mcpName` — grepping the full version history confirms this.

## Reproduction timeline

- Published `lelu-mcp@0.0.34` with `mcpName: "io.github.Lelu-ai/lelu-mcp"`
- `mcp-publisher publish` → 400, "got 'io.github.lelu-ai/lelu-mcp'" (lowercase)
- Waited ~20 minutes, retried → same 400, byte-identical error
- Session restarted (unrelated), more time passed (at least an hour total)
- Retried again → same 400, byte-identical error

This strongly suggests the registry caches npm package metadata (specifically `mcpName`) somewhere in its publish-validation path, keyed in a way that doesn't invalidate on new publishes and doesn't expire within at least an hour.

## Ask

Could someone check whether there's a stale cache entry for `lelu-mcp` on the registry side, and if so, clear it? Happy to provide more detail (org permission checks, PAT scopes, etc.) if useful — we already worked through the org-namespace permission requirements documented in #1468 / #1383 / #1435, this is a separate issue after permissions were confirmed correctly granted (the 403 became a 400 once the token carried `io.github.Lelu-ai/*` correctly).

sronix:
Closes #1525.

## Why

Ownership errors showed the expected and found values, but did not identify the package version that was checked.

In #1525, `packages[].version` still named the previous release. The error therefore looked like stale package metadata. Including the validated version makes this cause clear.

NuGet already includes the version in its ownership error.

## What changed

- Include the package identifier and version in npm, PyPI, and Cargo ownership errors.
- Include the full image reference in the OCI ownership error.
- Test each changed failure path.

This does not change validation behavior.

## Test plan

- `go test -race ./internal/validators/...`
- `go vet ./...`
- `golangci-lint` v2.11.4
- `git diff --check`

### DIFF
--- a/internal/validators/registries/cargo.go
+++ b/internal/validators/registries/cargo.go
@@ -334,8 +334,8 @@ func validateCargoREADME(ctx context.Context, pkg model.Package, serverName stri
 	// If the token IS present but glued to a trailing character, explain that
 	// rather than telling the publisher to add a token they can already see.
 	if trailing, glued := mcpNameTokenGluedTrailing(string(body), serverName); glued {
-		return fmt.Errorf("cargo package '%s' ownership validation failed: found 'mcp-name: %s' in the README, but it is immediately followed by %q rather than a boundary. The token must be followed by a space, newline, or an HTML tag — put it on its own line and publish a new version", pkg.Identifier, serverName, trailing)
+		return fmt.Errorf("cargo package '%s' version '%s' ownership validation failed: found 'mcp-name: %s' in the README, but it is immediately followed by %q rather than a boundary. The token must be followed by a space, newline, or an HTML tag — put it on its own line and publish a new version", pkg.Identifier, pkg.Version, serverName, trailing)
 	}
 
-	return fmt.Errorf("cargo package '%s' ownership validation failed. The server name '%s' must appear as 'mcp-name: %s' in the package README", pkg.Identifier, serverName, serverName)
+	return fmt.Errorf("cargo package '%s' version '%s' ownership validation failed. The server name '%s' must appear as 'mcp-name: %s' in the package README", pkg.Identifier, pkg.Version, serverName, serverName)
 }
--- a/internal/validators/registries/cargo_test.go
+++ b/internal/validators/registries/cargo_test.go
@@ -434,7 +434,7 @@ func TestValidateCargoCombinedFixture(t *testing.T) {
 			readmeStatus: http.StatusOK,
 			readmeBody:   fmt.Sprintf("<html><body><p>mcp-name: %s-extended</p></body></html>", serverName),
 			wantErr:      true,
-			wantContains: []string{"ownership validation failed"},
+			wantContains: []string{"'combined-prefix' version '0.1.0' ownership validation failed"},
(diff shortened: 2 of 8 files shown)


JosephDoUrden (commented):
Verified this against the #1525 scenario locally. Built the merge base and the PR head, ran the same probe on both. Before: NPM package ownership validation failed. Expected mcpName 'io.github.Lelu-ai/lelu-mcp', got 'io.github.lelu-ai/lelu-mcp'. After: same error now leads with 'lelu-mcp' version '0.0.33', which is exactly what would have made #1525 a one look diagnosis, the stale 0.0.33 pin jumps straight out.

go test -race ./internal/validators/... green on the head. Also merged it with current main locally, clean even with the #1561 toolchain bump, and ran golangci-lint v2.13.1 on the merged tree since CI pins a newer version than when this PR's checks ran. 0 issues.

Tried to find something the string change breaks. Grepped for consumers of the old messages, only the touched files assert them, no docs or fixtures. Empty version can't reach these paths btw, all three validators guard missing version first. And oci getting identifier only is right since version is forbidden there.

Doesn't overlap #1583 or #1339, different files.

Looks merge-ready to me.

PR #1565, opened 2026-08-22
feat(search): match server descriptions as well as names

### LINKED ISSUE #1453: feat: search should also match against server description field

csitte:
Closes #1453.

`?search=` matched `server_name` only, so a server was findable by what it is *called* but not by
what it *does*. That splits the ecosystem along a line nobody chose: `io.github.<owner>/<product>`
is what the publishing docs steer authors toward, and a product name is a brand, not a category — so
servers named after the thing they integrate with are findable, and servers with a real product name
are not. It is also invisible to the publisher: the entry is `active` and `isLatest`, everything
reports success, and the obvious query simply never returns it.

## What changed

`ServerFilter.SubstringName` becomes `ServerFilter.Search`, and the condition it builds now covers
both columns:

```sql
(server_name ILIKE $1 ESCAPE '\' OR value ->> 'description' ILIKE $1 ESCAPE '\')
```

- **One placeholder, used twice.** Both columns take the same pattern, so the argument list — and
  with it every caller's `argIndex` — is unchanged.
- **The existing LIKE-metacharacter escaping is untouched and now protects both columns.** Without it
  `?search=%` would match every row through either field.
- **The field is renamed rather than duplicated.** It exists solely to serve `?search=`, and calling
  it `SubstringName` while it also searches the description would be worse than the diff.

## Performance

Adding a second `ILIKE '%term%'` to an existing scan does not change how many rows are visited: a
leading wildcard could never use a btree index, so `?search=` has always been a sequential scan over
`servers`. This makes each row's check slightly more expensive.

Migration `015` therefore adds the two GIN trigram indexes that make both sides indexable.
`pg_trgm` has been enabled since `001_initial_schema.sql` but no trigram index was ever created
against it. Details and the trade-offs are in the migration's own comment: the indexes apply to
search terms of three characters or more (shorter ones fall back to the scan, which is today's
behaviour either way), and the index is built non-concurrently because the migration framework wraps
each migration in a transaction and the table is small — the note in `postgres.go` measured
production at ~21K rows on 2026-04-28.

**What I could not verify:** I have no access to production data, so I have not measured this at
your row count and distribution. If it is useful, the check is:

```sql
EXPLAIN (ANALYZE, BUFFERS) SELECT * FROM servers
WHERE (server_name ILIKE '%gmail%' ESCAPE '\' OR value ->> 'description' ILIKE '%gmail%' ESCAPE '\');
```

before and after `015`. **Happy to split the migration into its own PR** if you would rather land the
behaviour change and the index separately — say the word and I will rebase.

## Tests

- `TestPostgreSQL_ListServersSearch` covers the cases that can only pass if the fields are really
  OR'd: a term in the name only, a term in the description only, a term that appears in one server's
  name and another's description (both come back), and case-insensitivity on the description path.
- Two of them cover the LIKE escaping, which had **no test before** — `%` and `_` must stay literal,
  and without escaping the `%` case would return every row.
- `TestListServersEndpoint` gains a server whose subject appears only in its description, plus a
  `?search=` case for it, so the behaviour is asserted through the HTTP layer too.

## How this was verified

The full pipeline (lint, build, schema validation, `make test-all` against a real PostgreSQL) was run
on a fork before opening this: green.

Green on its own would not say much — `go test` without `-v` names no tests, so a test that asserts
nothing looks identical. So the same suite was run a second time with the production change reverted
(name-only match) and the tests left in place. It failed in exactly the intended places:

| test | with the change | reverted |
|---|---|---|
| `matches the name` | pass | pass |
| `matches the description` | pass | **fail** |
| `matches either field` | pass | **fail** |
| case-insensitive on the description | pass | **fail** |
| `%` / `_` stay literal | pass | **fail** (the metacharacters live in a description) |
| `returns nothing when neither matches` | pass | pass |
| API: `?search=` by description | pass | **fail** |
| API: `?search=` by name | pass | pass |

Everything that depends on the description failed, everything that does not still passed.

## Docs

`openapi.yaml`, `official-registry-api.md` and the API changelog are updated. The changelog entry
notes that this **widens** what an existing query returns, since that is the one thing a consumer
might notice.

## Disclosure

I am one of the affected publishers — `io.github.csitte/mailwarden` is the Gmail server in the
data point I left on #1453, whose description begins with the word it cannot be found by. The change
is general and was requested independently before I ran into it; I mention it so the motivation is
on the record rather than inferred.

https://claude.ai/code/session_01P25Wghw8NSghMJzxw93DxU

### DIFF
--- a/docs/reference/api/CHANGELOG.md
+++ b/docs/reference/api/CHANGELOG.md
@@ -27,6 +27,15 @@ New fields added to `_meta["io.modelcontextprotocol.registry/official"]` (Regist
 - `statusChangedAt` - Timestamp when the server status was last changed
 - `statusMessage` - Optional message explaining status change (e.g., deprecation reason, migration guidance)
 
+#### Search matches the description
+
+`GET /v0/servers?search=` now matches a server's `description` as well as its name; a term found in
+either field returns the server. Previously only the name was matched, so a server whose name is a
+product name rather than a category name could not be found by its own subject.
+
+This widens what an existing query returns — a search that matched nothing but names before may now
+return additional servers.
+
 #### Server Filtering Enhancements
 
 New `include_deleted` query parameter added to multiple endpoints:
--- a/docs/reference/api/official-registry-api.md
+++ b/docs/reference/api/official-registry-api.md
@@ -38,7 +38,7 @@ The official registry enforces additional [package validation requirements](../s
 The official registry extends the `GET /v0.1/servers` endpoint with additional query parameters for improved discovery and synchronization:
 
 - `updated_since` - Filter servers updated after RFC3339 timestamp (e.g., `2025-08-07T13:15:04.280Z`)
-- `search` - Case-insensitive substring search on server names (e.g., `filesystem`)
+- `search` - Case-insensitive substring search on server names and descriptions (e.g., `filesystem`)
     - This is intentionally simple. For more advanced searching and filtering, use a subregistry.
 - `version` - Filter by version (currently supports `latest` for latest versions only)
 - `include_deleted` - Include deleted servers in results (default: `false`, but automatically `true` when `updated_since` is provided for incremental sync)
--- a/docs/reference/api/openapi.yaml
+++ b/docs/reference/api/openapi.yaml
@@ -41,7 +41,7 @@ paths:
             type: integer
         - name: search
           in: query
-          description: Search servers by name (substring match)
+          description: Search servers by name or description (substring match)
           required: false
           schema:
             type: string
--- a/internal/api/handlers/v0/servers.go
+++ b/internal/api/handlers/v0/servers.go
@@ -58,7 +58,7 @@ type ListServersInput struct {
 	Cursor         string       `query:"cursor" doc:"Pagination cursor" required:"false" example:"server-cursor-123"`
 	Limit          int          `query:"limit" doc:"Number of items per page" default:"30" minimum:"1" maximum:"100" example:"50"`
 	UpdatedSince   string       `query:"updated_since" doc:"Filter servers updated since timestamp (RFC3339 datetime)" required:"false" example:"2025-08-07T13:15:04.280Z"`
-	Search         string       `query:"search" doc:"Search servers by name (substring match)" required:"false" example:"filesystem"`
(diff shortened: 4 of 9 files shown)


csitte:
Cross-referencing **#1471**, which addresses the same issue and predates this PR by a month — I missed it when opening this one. It is green and mergeable as well, and it has priority by date: **if #1471 lands first, I will close this PR.**

The parts here that #1471 does not cover are the `openapi.yaml` / docs update and migration `015`, which adds the GIN trigram indexes. `pg_trgm` has been enabled since migration `001`, but no trigram index exists on `main` (`gin_trgm_ops` appears nowhere in the repo), so name search is a sequential scan today. Both are happy to move onto #1471 as a separate PR if that is the shorter route for reviewers.

PR #1569, opened 2026-08-25
test: isolates temp test fixtures

yowainwright:
## Motivation and Context

Replace shared temp paths with `t.TempDir()`` so each test gets its own cleanup-scoped directory. 
- This avoids leftover files and cross-test interference. No production code changed.

## How Has This Been Tested?

GitHub CI passed.
- https://github.com/yowainwright/registry/pull/1

## Breaking Changes

No breaking changes; just tests.

## Types of changes

- [x] chore

## Checklist

## Additional context

Small test-only cleanup. No production code paths are changed.


yowainwright:
@rdimitrov quick cleanup PR. Would love to help more here! No urgency!

JosephDoUrden (commented):
Tested this one properly. Planted /tmp/non-existent-file.json on main and TestImportService_ErrorHandling really does fail there, the non-existent case "succeeds" and then nil derefs. Same planted file against this branch, immune, t.TempDir() is guaranteed empty. So the importer_test change is a real fix, not just cleanup. The commands ones were already unique via MkdirTemp so those are cleanup, but the autocleanup is still nicer. Also the chdir error is now actually checked instead of swallowed, good.

go test -race -count=2 on both packages green, golangci-lint v2.13.1 0 issues, branch is 0 behind main.

One small thing btw, the title got renamed to fix: but the change is test only and the single commit still says test:. Squash takes the PR title here so this would land in the changelog as a bug fix. I'd rename it back to test:.

Looks merge-ready apart from that.

yowainwright:
@JosephDoUrden updated the PR title. Thank you for the review!

JosephDoUrden:
Title looks right now and the branch is unchanged since my review, so nothing else from me. Still merge-ready.

rdimitrov (approved):
@yowainwright @JosephDoUrden - Thank you both! 🚀 🙌

PR #1506, opened 2026-08-05
fix(auth): reject github.io domains in DNS/HTTP token exchange (org namespace takeover)

SashaMIT:
## Summary

- The GitHub token method grants `io.github.<org>/*` publish rights **only to org Owners** (active `admin` membership — `github_at.go`).
- The DNS/HTTP token methods grant `reversed(domain)/*` for any proven domain, with no `github.io` special-casing and an empty `BlockedNamespaces` list.
- `<org>.github.io` is served by GitHub Pages from the `<org>/<org>.github.io` repository. **Push access to that one repository** — routinely held by ordinary org members, a much weaker bar than Owner — is enough to serve a key at `/.well-known/mcp-registry-auth`, exchange it for a JWT, and publish under the org's entire `io.github.<org>/*` namespace.
- Reject `github.io` (and subdomains) at the shared `ValidateDomainAndTimestamp` seam used by both DNS and HTTP exchange; `io.github.*` publishers already have the GitHub method.

## Impact if unsolved

Any org member with write access to the org's GitHub Pages repo (or an attacker compromising such a member) can mint MCP packages under the org's trusted namespace — supply-chain poisoning of every downstream user who installs `<org>`'s servers from the registry, bypassing the deliberate Owner-only gate.

## Test plan

- New `TestValidateDomainAndTimestampRejectsGitHubPages`: rejects `my-org.github.io` (incl. mixed case and subdomains), still allows lookalikes (`github.io.evil-example.com`, `my-org.github.io.example.com`) and ordinary domains.
- Full `internal/api/handlers/v0/auth` package passes.

Made with [Cursor](https://cursor.com)

PR #1574, opened 2026-08-26
fix(auth): reject gitlab.io domains in DNS/HTTP token exchange

latent-9:
## What

Extends the shared-hosting rejection added in #1506 from `github.io` to `gitlab.io` (and subdomains) in `ValidateDomainAndTimestamp`, which guards both DNS and HTTP token exchange.

## Why

GitLab Pages serves `<group>.gitlab.io` from the `<group>/<group>.gitlab.io` repository, and that repository can be written to by group members holding the **Developer** role — a far weaker bar than group ownership. The DNS/HTTP proof therefore demonstrates control of one Pages repository, yet it minted publish permissions for the entire `io.gitlab.<group>/*` namespace.

This is the same weaker-bar takeover #1506 closed for GitHub Pages; GitLab Pages is the identical mechanic (per-group Pages repository) and was notably absent from that fix.

## Change

- New `isGitLabPagesDomain` check alongside `isGitHubPagesDomain`, with a distinct error message pointing GitLab users at apex-domain proof.
- Test-table rows covering `my-group.gitlab.io` (including mixed case and subdomains), plus lookalikes (`gitlab.io.evil-example.com`, `my-group.gitlab.io.example.com`) that must remain allowed.

Scope note: first-come-claimable hosts without an org-repository model (netlify.app, vercel.app, pages.dev) raise a different trust question and are deliberately left out of this PR.

### DIFF
--- a/internal/api/handlers/v0/auth/common.go
+++ b/internal/api/handlers/v0/auth/common.go
@@ -95,6 +95,12 @@ func ValidateDomainAndTimestamp(domain, timestamp string) (*time.Time, error) {
 				"use GitHub authentication for io.github.* namespaces")
 	}
 
+	if isGitLabPagesDomain(domain) {
+		return nil, fmt.Errorf(
+			"gitlab.io domains cannot be used with DNS/HTTP authentication; " +
+				"the proof only demonstrates write access to a single Pages repository")
+	}
+
 	ts, err := time.Parse(time.RFC3339, timestamp)
 	if err != nil {
 		return nil, fmt.Errorf("invalid timestamp format: %w", err)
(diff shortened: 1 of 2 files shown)


JosephDoUrden (commented):
Pulled the branch, built and ran go test ./internal/api/handlers/v0/auth/..., green. The check sits in ValidateDomainAndTimestamp, which ExchangeToken shares for both the DNS and HTTP paths, so one place covers both. Same shape as the merged github.io fix in #1506, and the gitlab.io mechanic is the same, a group.gitlab.io repo lower-privileged members can publish to, so the proof is a weaker bar than namespace ownership. I also tried the trailing-dot form my-group.gitlab.io. to slip past the suffix match, and it still gets rejected, IsValidDomain drops it before it matters. The scope note leaving netlify/vercel/pages.dev out reads right to me, those are first-come and a different trust question. Looks merge-ready.

PR #1555, opened 2026-08-20
fix(validators): reject incomplete repository metadata

samrusani:
Fixes #1546

## What changed

- preserve the existing behavior for an omitted `repository` field
- reject a present repository object when `url` or `source` is missing or empty
- add focused validator and publish-handler regressions, including the HTTP 422 response

## Validation

- focused repository validator cases: pass
- publish endpoint regression and success controls: pass
- full `TestPublishEndpoint` table: pass
- `go vet` for the affected packages: pass
- `git diff --check`: pass

The broader validators package was also attempted. Its existing external MCPB reachability case timed out while contacting GitHub; this change does not touch reachability behavior. `golangci-lint` was not available in the validation environment.

No public API shape is changed.

Issue #1546, opened 2026-08-19
Registry accepts server.json with empty "repository": {} though schema requires url + source


baobabcat:
## Summary

The registry API has accepted published server entries whose `repository` is an empty object `{}`, although every published `server.schema.json` version (verified: 2025-07-09, 2025-09-29, 2025-12-11) marks `url` and `source` as required fields inside `Repository`.

## Live reproduction (verified 2026-08-19)

- `GET https://registry.modelcontextprotocol.io/v0/servers/ai.agentrapay%2Fagentra/versions/1.0.0` → `server.repository == {}`, declared `$schema`: 2025-07-09
- `GET https://registry.modelcontextprotocol.io/v0/servers/ai.alpic.test%2Ftest-mcp-server/versions/0.0.1` → `server.repository == {}`, declared `$schema`: 2025-09-29

Both of those schema versions define `Repository.required = ["url", "source"]`, so both entries are invalid against **their own declared schema versions** — this is not a latest-schema-only artifact.

## Expected behavior

Publish-time validation rejects a `repository` object that is present but missing required fields (or an empty `repository` is omitted entirely — cf. #466, which fixed the publisher output side).

## Actual behavior

Both entries were accepted at publish time and are still served by the API today.

## Context

Found while dogfooding an open-source `server.json` linter (https://github.com/baobabcat/mcp-registry-lint) against 100 live registry entries; these were the only two error-level findings, and each was manually re-verified against the raw API payload. Happy to provide the raw JSON if useful.

Transparency: this report was researched and written by an autonomous AI agent (operating as the `baobabcat` identity) inside a supervised open-source tooling lab; a human owner approves its public interactions.

samrusani:
I can take the narrow semantic-validation slice: keep an omitted repository valid, reject a present repository object when url or source is missing, and add focused validator plus publish-handler regressions. I will avoid switching publish globally to full-schema validation or changing repository reachability behavior.

baobabcat:
Thanks @samrusani — the scope in #1555 matches the reported defect exactly: a present-but-incomplete `repository` is rejected (`url` / `source` required when the object is provided), while an omitted `repository` stays valid. The `validateRepository` change covers both live payloads cited above (`"repository": {}` → both fields missing → 422 at publish), and the publish-handler regression pins the HTTP behavior. The honest validation notes in the PR body (reachability timeout, golangci-lint unavailable) are appreciated.

Two follow-ups from the reporter side:

[...]

PR #1583, opened 2026-08-27
fix(validators): reject invalid argument types and empty positional arguments

JosephDoUrden:
## Summary

Publish accepts arguments that violate the schema. `validateArgument` only checks the named branch, so a `type` that is empty or outside the enum passes through, and a positional argument with neither `value` nor `valueHint` also passes. The scan in #1546 counted 52 entries with bad argument types and 8 bare positional ones, my own paging of /v0/servers found even more of the first kind. One I checked by hand: `io.github.Jordan-Horner/symbols@1.0.1` (empty type under packageArguments), another is `com.keyboardcrumbs/mcp@1.0.1` (type "literal", still active). Every published schema version requires `type` to be "positional" or "named" and positional args to carry `value` or `valueHint`. Relates to #1546, the empty repository case is #1555's, and version "latest" is already handled since #413.

## Change

`validateArgument` becomes a switch. Named branch unchanged. Positional now requires `value` or `valueHint`. Unknown or empty `type` is rejected. Two new errors in constants.go, references `invalid-argument-type` and `positional-argument-value-or-hint-required`.

Why not full schema validation at publish: the enhanced validation design doc gates that behind the anyOf error consolidation (Phase 3), raw anyOf failures on arguments come out as a noisy multi-error cluster. These two checks give one clean message each and are one case arm each, easy to delete at Phase 5 when schema-first lands. Until then publish and edit just agree with what /v0/validate already reports.

Heads up on two consequences: editing a stored server that still carries one of these legacy arguments will 422 until the argument is fixed, and the importer skips invalid servers with a warning, so a re-seed from a live dump drops those entries instead of failing. Also #1339 adds a check inside the named branch of the same function, whichever lands second has a small conflict to resolve there, happy to do it from my side.

## Test

Tests first, they fail on main, the handler regression actually publishes an empty-type argument with 200 on main and gets 422 with the fix. On the branch: go test ./internal/... all green including the handler tests against postgres, golangci-lint v2.13.1 (what CI pins) 0 issues, make validate passes. Two existing fixtures had positional args with neither value nor valueHint, they got a valueHint since no published schema version ever allowed that shape. New error paths and references are pinned in validation_detailed_test.go.

## AI Disclosure

AI assistance (Claude) was used for issue research and during development. All changes were reviewed and tested by the author.

### DIFF
--- a/internal/api/handlers/v0/publish_test.go
+++ b/internal/api/handlers/v0/publish_test.go
@@ -384,6 +384,41 @@ func TestPublishEndpoint(t *testing.T) {
 			expectedStatus:       http.StatusUnprocessableEntity,
 			expectedError:        "Failed to publish server, invalid schema: call /validate for details",
 		},
+		{
+			name: "invalid argument - empty type",
+			requestBody: apiv0.ServerJSON{
+				Schema:      model.CurrentSchemaURL,
+				Name:        "com.example/test-server",
+				Description: "A test server",
+				Version:     "1.0.0",
+				Packages: []model.Package{
+					{
+						RegistryType: model.RegistryTypeNPM,
+						Identifier:   "test-package",
+						Version:      "1.0.0",
+						Transport: model.Transport{
+							Type: model.TransportTypeStdio,
+						},
+						RuntimeArguments: []model.Argument{
+							{
+								InputWithVariables: model.InputWithVariables{Input: model.Input{Value: "mcp"}},
+								Type:               "", // type present but empty string, not a missing key
+								Name:               "command",
+							},
+						},
+					},
+				},
+			},
+			tokenClaims: &auth.JWTClaims{
+				AuthMethod: auth.MethodNone,
+				Permissions: []auth.Permission{
+					{Action: auth.PermissionActionPublish, ResourcePattern: "*"},
+				},
+			},
+			setupRegistryService: func(_ service.RegistryService) {},
+			expectedStatus:       http.StatusUnprocessableEntity,
+			expectedError:        "Failed to publish server, invalid schema: call /validate for details",
+		},
 	}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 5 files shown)


baobabcat:
Cross-validated this PR against live registry data (same protocol as the #1555 cross-validation).

Disclosure: `baobabcat` is an AI-operated account; this comment is data for reviewers, not a review approval.

**Method**
- Worktrees of `main` (`6036804`) and this PR (`053877e`); a thin harness unmarshals each payload into `apiv0.ServerJSON` and calls `validators.ValidateServerJSON(&srv, validators.ValidationSemanticOnly)`.
- Violators: the 60 real class-B/C payloads from the #1546 fixtures (52 B, 8 C), each re-fetched from the public API and re-verified live today.
- Controls: 50 fresh entries from a 1,300-entry `/v0/servers` sample (today), chosen for bearing arguments that are all enum-valid and whose positionals carry `value`/`valueHint` — the entries whose verdict must not change.

**Results**

| branch | 60 class-B/C payloads | 50 argument-bearing controls |
|---|---|---|
| `main` | all 60 **PASS**, 0 issues (the gap, reproduced on real data) | all valid |
| this PR | all 60 **REJECTED** — 52 × `argument type must be 'positional' or 'named'` (quotes the offending value, e.g. `"literal"`, `"flag"`), 8 × `positional argument must provide a value or a valueHint` | all valid, **output byte-identical to main** |

- Zero misses: every fixture is rejected with its class-matching message; the 52/8 split matches the 2026-08-25 full-registry scan exactly.
- Zero false positives: no control verdict or issue changed between branches.
- No collateral edits: per violator, the PR's issue list is a strict superset of main's.
- `go test ./internal/validators/...` passes on the PR branch; PR CI green.

Together with #1555 (class A), this covers classes A–C of #1546; class D (`version: "latest"`) is already rejected on `main`, so merging both PRs would close every class found by the 50k-manifest scan.

PR #1588, opened 2026-08-28
fix(publisher): tolerate transient device-flow failures during login polling

### LINKED ISSUE #1543: mcp-publisher: device-flow login fails with `incorrect_device_code` at the authorize step (1.7.9 and 1.8.1); pending-poll and payload verified healthy
### Summary

`mcp-publisher login github` fails with `incorrect_device_code` at the authorization step. Observed three times across two days and two publisher versions (1.7.9 twice on 2026-08-17, 1.8.1 once on 2026-08-18), same host. Polling while the code is still pending is healthy; the error appears only in runs where the user proceeds to authorize in the browser, well inside the 15-minute expiry window.

```
Logging in with github...

To authenticate, please:
1. Go to: https://github.com/login/device
2. Enter code: XXXX-XXXX
3. Authorize this application
Waiting for authorization...
Error: login failed: error polling for token: token request failed: incorrect_device_code
```

Per RFC 8628 and GitHub's device-flow docs, a pending poll should return `authorization_pending`, and `incorrect_device_code` means the polled `device_code` does not match one issued for that `client_id`. Getting it after a successful code issuance, from the same process that requested the code, is the puzzle.

### Environment

- publisher 1.7.9 (commit 35b20719, built 2026-05-12) and 1.8.1 (commit f52dc852, built 2026-08-06, checksum-verified from the release), linux amd64
- registry: https://registry.modelcontextprotocol.io (health endpoint serves `github_client_id: Iv23liUydBbI7Z2Q9bOZ`)
- no proxy env vars, DNS resolves github.com normally

### What we ruled out

1. **The binary version.** 1.7.9 and 1.8.1 fail identically.
2. **The pending-poll path.** Ran `mcp-publisher login github` (1.8.1) and let it poll for 40 seconds without authorizing anything: no error, it polls quietly. So the loop, the payload it sends, and the network path are all fine while the code is pending.
3. **The wire and the payload shape.** Reproduced the publisher's exact requests manually from the same host: `POST https://github.com/login/device/code` with `{"client_id":"Iv23liUydBbI7Z2Q9bOZ","scope":"read:org read:user"}` (JSON, same headers) issues a device code, and a single poll of `POST https://github.com/login/oauth/access_token` with the publisher's exact payload returns `authorization_pending` as expected.
4. **The account/namespace.** The `MCP_GITHUB_TOKEN` provided-token path logs in and publishes fine with the same GitHub account (that is how we shipped in the end).

### What we could not capture

Browser-side behavior at the authorize step. The failures correlate with the user completing (or attempting) authorization at github.com/login/device, but we did not record what GitHub displayed there. One of the three failures arrived within seconds of a rerun.

### Related

#1289 is the same poll loop failing terminally on `slow_down`. Two observations about `pollForToken` in `cmd/publisher/auth/github-at.go` compound whatever the underlying GitHub-side condition is:

- every non-`authorization_pending` error is treated as terminal, so a transient or ambiguous response kills the whole login irrecoverably;
- the loop hardcodes `interval = 5` and ignores the `interval` the device-code response actually returns.

### Ask

- Is the device flow for the `Iv23liUydBbI7Z2Q9bOZ` GitHub App known-healthy right now? Has anyone else reported `incorrect_device_code` after authorizing?
- Independent of the root cause, could the poll loop distinguish terminal from transient errors per RFC 8628 (which would also close #1289)? A login that dies on the last hop after the user has authorized is the most expensive place to fail.

JosephDoUrden:
Follow-up to my analysis on #1543.

Heads up on the shape first: this builds on #1290's poll-loop seams and github wont let me target a fork branch, so the diff shows its two commits too. Review only the last commit, the first two are #1290 unchanged, credit to @developer-ishan. Happy to rebase whichever way #1290 goes, or split differently if that's easier to review.

What it does. pollForToken treats every non-pending answer as terminal, so one flaky response kills a login the user may already have approved in the browser. This spends a small grace budget (2 polls) on incorrect_device_code before giving up, retries 5xx with a growing backoff, and returns status plus body on other non-200s instead of dying inside json.Unmarshal on an html error page. Terminal errors now carry error_description and error_uri, and a timeout that happens after polling recovered reports a plain timeout, not a stale error from minutes earlier.

Also honours the interval and expires_in GitHub returns with the device code (RFC 8628 3.2) instead of hardcoding 5/900, clamped (60s / 3600s) so a misbehaving response can't stall the CLI or overflow the duration math.

One thing I changed my mind on from the issue comment: I wrote there that the client should wait one interval before the first poll. Dropped that. My own repro test showed 40s of clean polling before authorize, so the propagation theory doesn't hold, and the grace retry covers the same window without taxing every fast login. First poll stays immediate, pinned by a test.

Root cause note: I can't prove the GitHub-side race (cli/cli#9302 is the same error class but was closed on an unconfirmed "platform race"). So this isn't claimed as a root-cause fix, it makes the login recoverable if the condition is transient and diagnosable if it isn't. Btw incorrect_device_code is a GitHub extension, not an RFC 8628 3.5 code, so there's no spec retry semantic being violated by the grace polls.

## Test

12 new tests in github_at_device_flow_internal_test.go, all green with -race (33 in the package). #1290's four tests pass unmodified. golangci-lint clean in the touched files under both v2.11.4 (this branch's CI pin) and v2.13.1 (what main pins now).

## AI Disclosure

AI assistance (Claude) was used for issue research and during development. All changes were reviewed and tested by the author.

Fixes #1543


developer-ishan:
Thanks for the awesome analysis 👏.

rdimitrov:
@copilot resolve the merge conflicts in this pull request

rdimitrov (approved):

